Issue #145, opened 2015-01-07
Expose config for problems host


kytrinyx:
At the moment we only expose the config for the web api (to submit problems, etc), but not the problems api (where we get the actual READMEs and test suites). This is usually fine, since most people can simply use the website locally and point to the remote problem API.

This becomes a problem when people are running a completely separate installation of exercism with custom problems.

The configuration should optionally handle an additional argument to configure the problems API.

/cc @morphatic

kytrinyx:
Closed in #147

PR #146, opened 2015-01-07
add --api option to exercism configure command

morphatic:
In response to #145 

Summary of changes:

* add `--api` and `-a` parameters to `exercism configure`
* updated `config_test.go` to test success of changes
* replaced `github.com/exercism/cli` in all `import` statements with `..` to refer to local code

I found in working on this that my changes to code didn't get reflected in builds because the `import` statements always pulled the other cli library code from github instead of the local package, hence the 3rd change above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ go:
   - 1.1.2
   - tip
 
-install: true
+install:
+  go get -d -v ./... && go build -v ./...
 
 before_install:
   go get github.com/codegangsta/cli && go get github.com/stretchr/testify/assert
--- a/api/api.go
+++ b/api/api.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/exercism/cli/config"
+	"../config"
 )
 
 var (
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -6,7 +6,7 @@ import (
 	"os"
 
 	"github.com/codegangsta/cli"
-	"github.com/exercism/cli/config"
+	"../config"
 )
 
 // Configure stores settings in a JSON file.
@@ -22,7 +22,8 @@ func Configure(ctx *cli.Context) {
 	key := ctx.String("key")
 	host := ctx.String("host")
 	dir := ctx.String("dir")
-	c.Update(key, host, dir)
+	api := ctx.String("api")
+	c.Update(key, host, dir, api)
 
 	if err := os.MkdirAll(c.Dir, os.ModePerm); err != nil {
 		log.Fatalf("Error creating exercism directory %s\n", err)
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -7,7 +7,7 @@ import (
 	"runtime"
 
 	"github.com/codegangsta/cli"
-	"github.com/exercism/cli/config"
+	"../config"
 )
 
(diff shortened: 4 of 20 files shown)


Tonkpils:
The local code you're working on, is it under the same hierarchy that `go get` places it under?

morphatic:
@Tonkpils No, I didn't.  This is my first time working in `go` and I didn't really understand the `go get` command.  I've done a little reading and I think I understand it now, but I think I'm still a bit confused about how to submit patches to this project.  Here's what I did (which I think follows the normal workflow for contributing to github projects):

* forked the project ([my fork](https://github.com/morphatic/cli))
* checked out my fork locally
* modified local code to add new functionality and tests
* committed back to my repo
* issued a pull request

What the [Development](https://github.com/exercism/cli#development) section of this project says is that I should just use `go get github.com/exercism/cli` and then modify the code in `$GOPATH/src/github.com/exercism/cli/` and then issue a pull request.  Correct me if I'm wrong, but this won't work because my local version is not in a public repo and so there's no URL that I can use in the `git request-pull` command.  (BTW, when I do it this way I can get all of the tests to pass without changing the paths in the `import` statements, but then I can't push changes because I don't have permission.)

Here's the crux of the problem: `import` statements that refer to `github.com/exercism/cli/`.  It creates a problem for would-be contributors because either:

* you fork `exercism/cli` and your tests won't pass locally because they all refer to the upstream repo
* you `go get` `exercism/cli`, make your changes and then are stuck because there's no way to make a pull request.

To illustrate more concretely, in order to add a new command line option to the `exercism configure` command, I had to change the `Update` function in `config/config.go` to accept a 4th parameter (the URL for the xapi service), and then I changed the `Configure` function in `cmd/configure.go` to use the new 4th parameter.

However, as it currently stands, the `cmd/configure.go` file imports `github.com/exercism/cli/config` which means that any changes I make to `config/config.go` are NOT seen and the tests fail saying that there are too many parameters.

I see a couple of potential ways to resolve the issue:

* change the paths in the `import` statements to something more generic
* update the `.travis.yml` instructions to accommodate the relative reference, i.e. `../`, syntax that I used

All of this being said, this is my first `go` project so I wouldn't be surprised at all if I didn't understand something correctly.

Thanks!

Tonkpils:
I see your point. I read around a bit and from most of what I've read about relative path imports is that they are only kept around for legacy reasons and should probably not be used.
I can't tell you exactly what the reasoning behind it is but I have not seen any Go projects use it. I will keep doing some more reading on this. 

The workflow I used was:

1. Fork this repo
1. `go get` this repo
1. add the forked repo git url as a new remote in the original project (the one used in go get)
1. make modifications on the original but push to the remote that was added (the one pointing to your fork)

That still allowed me to maintain the import paths and make changes to different packages in the project while still being able to push to my fork and creating a pull request from Github. 

Another workflow I saw while reading about this was to follow your workflow but once the PR was approved then change the import paths back to their original ones. 

If anyone else can shine some light into why relative import paths are not commonly used I'd definitely like some clarification as well.

morphatic:
Gotcha.  I'm going to try your process and issue a new pull request.

kytrinyx:
This is a common enough issue that I blogged about it a while back: https://blog.splice.com/contributing-open-source-git-repositories-go/

PR #153, opened 2015-01-18
Refactor config

kytrinyx:
We only ever do three things from the outside:

1. Load an existing config, falling back to defaults if it doesn't exist.
2. Update values on a config.
3. Write the config to disc.
    
Everything else is details.
    
In order to be able to set a fake homeDir on this, the New function delegates to a separate load function. This also allows us to avoid potential race conditions in the tests by not setting ENV variables.


config/config.go
```diff
@@ -134,30 +141,39 @@ func (c *Config) Read(file string) error {
...
-
-	e := json.NewEncoder(f)
-	return e.Encode(c)
+	return strings.Replace(path, "~", h, 1), nil
 }
 
-func (c *Config) configure() (*Config, error) {
-	c.sanitize()
-
+func (c *Config) setDefaults() error {
```
lcowell: this name does a much better job of communicating what it does. :thumbsup:

kytrinyx: Yeah, everything seemed a bit tangled before. Or perhaps: both tangled and fragmented. This time around I've kept larger chunks. I think fewer abstractions will be easier to manage.

Tonkpils:
This looks good to me, it definitely improves readability and its easier to understand whats going on. :+1:

kytrinyx:
OK, cool. Merging.

PR #154, opened 2015-01-18
Add List command to display a list of available exercises

### LINKED ISSUE #89: Display list of all the exercises
```bash
$ exercism list go
```

Should show a list of all the exercises, and explain how to get specific ones.

lcowell:
WIP - I just realize that I need to handle a 500 response in the case that someone tries to list for a language like 'rub' or 'haskll'.

Closes #89


cmd/list.go
```diff
@@ -0,0 +1,44 @@
...
+	"io"
+	"log"
+	"os"
+
+	"github.com/codegangsta/cli"
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
+)
+
+const msgExplainFetch = "In order to fetch a specific assignment, call the fetch command with a specific assignment.\n\nexercism fetch ruby/matrix"
```
kytrinyx: The fetch command actually takes two arguments: `exercism fetch ruby matrix`

cmd/list.go
```diff
@@ -0,0 +1,44 @@
...
+
+	problems, err := api.List(language, c.XAPI)
+	if err != nil {
+		log.Fatal(err)
+	}
+	GenerateOutput(problems, os.Stdout)
+}
+
+// Output writes out the list of assignments
+func GenerateOutput(problems []string, w io.Writer) {
```
kytrinyx: I don't really like exposing a function like `GenerateOutput` to the entire `cmd` package. What do you think of adding a new package that can hold on to track-related things? Earlier I've isolated some things into the `Homework` type in the `user` package. Would that approach work?

cmd/list.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, p := range problems {
+		fmt.Fprintf(w, "%s\n", p)
```
kytrinyx: Should we strip off the `ruby/` prefix?

kytrinyx:
It's cool to see progress again on this project! :sparkles: :heart: :sparkles:

kytrinyx:
I think the failing test was due to a stupid mistake I made on master.

cmd/list.go
lcowell: Good catch. It actually worked for me to say `exercism fetch ruby/matrix`, but we should be telling people the right thing to do, not the thing that happens to work in this release.

cmd/list.go
lcowell: Definitely, that would be easier to copy and paste.

cmd/list.go
lcowell: I think I was having a ruby moment here. Of course `GenerateOutput` is going to be available to everything in the `cmd` package. The motivation behind adding the function was to allow me to easily test the generated output. However, the test is quite brittle and adds little value.

My leaning would be towards:
* changing the function name to be something unexported and more appropriate like: `outputProblemList` -or-
* folding the function in to `cmd.List` and only abstracting as required

I'd favour eliminating the function for now. What do you think about doing this for now and exploring the value of a `track` package once we merge this in ?

lcowell:
One last thing I was wondering about is: should we be generating a better response when someone requests a language that doesn't exist ? eg `http://x.exercism.io/tracks/rby` will just give me a 500. That's ok, but I can't really know if I'm getting a 500 because of the language or because there's something going on with the server.

cmd/list.go
kytrinyx: Yeah, I think inlining it is good. We can refactor later if necessary.

kytrinyx:
> http://x.exercism.io/tracks/rby

That should return a 404, not a 500.

https://github.com/exercism/x-api/issues/37

api/api.go
```diff
@@ -128,6 +128,31 @@ func Submit(url string, iter *Iteration) (*Submission, error) {
 	return ps.Submission, nil
 }
 
+// List available problems for a language
+func List(language, host string) ([]string, error) {
```
Tonkpils: I like the way this function takes separate arguments as opposed to the composed url. Even providing host seems a bit off but until we have a Client struct for the api package this is fine. I think we should follow this same pattern for the other functions in this file. Any thoughts on this?

lcowell: Thanks! I think what I like about this is that the caller of `api.List` doesn't need to know about how to compose a URL. AFAIK each of the api commands takes a url as an argument which means those URLs are probably being generated in each of the `cmd` funcs. I don't think that `cmd` package should know about URLs.

What do you think about making functions for generating URLs and putting them in the api package ? It kind of seems like the right place for them to me.

```go
  func ListURL(language, host string) string {}
  func UnsubmitURL(host string) string {}
  func FetchURL(parts ...string) string {}
```

kytrinyx: Yeah, agreed. Let's let the api worry about urls (functions or hard-coded), and the function arguments can be whatever information is needed.

kytrinyx: That said, that is probably an unrelated refactoring.

kytrinyx:
The API now returns a 404 for non-existent languages.

Do you want to update the PR with this before we merge? Once you're ready would you mind squashing the commits?

lcowell:
Handling 404s is in place. Commits squashed. Wahoo!

kytrinyx:
Sweeeet!

PR #157, opened 2015-01-21
Clean up API package.

Tonkpils:
Use an API Client from the cmd package
Have functions form up their own urls and give them parameters they need

I wanted to get this proof of concept here to get some feedback. Meanwhile, I'll write some tests around this new structure.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -7,16 +7,11 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
-
-	"github.com/exercism/cli/config"
 )
 
 var (
-	// UserAgent lets the API know where the call is being made from.
-	// It's set from main() so that we have access to the version.
-	UserAgent string
-
-	UnknownLanguageError = errors.New("the language is unknown")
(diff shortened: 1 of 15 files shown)


api/api.go
```diff
@@ -7,16 +7,11 @@ import (
...
 )
 
 var (
-	// UserAgent lets the API know where the call is being made from.
-	// It's set from main() so that we have access to the version.
-	UserAgent string
-
-	UnknownLanguageError = errors.New("the language is unknown")
+	// ErrUnknownLanguage represents an error returned when the language requested does not exist
+	ErrUnknownLanguage = errors.New("the language is unknown")
```
kytrinyx: Much more idiomatic!

kytrinyx:
This looks much, much cleaner. I would say go for it.

Tonkpils:
I'll keep adding tests as I go but the main idea is there.

lcowell:
Awesome! Don't forget to give yourself some credit in the changelog.

kytrinyx:
I've added an entry to the changelog for this. Thanks!

PR #164, opened 2015-03-09
Add golang version dependency.

dkinzer:
I already had an older version of Go installed on my computer and when I tried building the CLI I got the following error:

```
cmd/debug.go:50: unknown http.Client field 'Timeout' in struct literal 
```


kytrinyx:
Good call, thanks!

Issue #35, opened 2013-12-15
Provide support for submitting multiple files


kytrinyx:
```json
{
    "files": [
    {
      "filename": "path/to/file.ext",
      "code": "# code goes here...",
    },
  ]
}
```

See related issue: exercism/exercism.io#1165

kytrinyx:
The backend supports accepting multiple files (see https://github.com/exercism/exercism.io/issues/1165), so this one can be bumped to the top.

lcowell:
Would the usage now be that I would specify multiple files from the command line eg:
`exercism submit AbstractBobPresenter.java Bob.java`

The `Iteration` struct, is what is used with `client.Submit`. Currently it has a File attribute, so we would need to change it to be a slice of Files and update the code that works with iterations to handle multiple files. Does that sound about right so far as what's involved in completing this?

kytrinyx:
> Would the usage now be that I would specify multiple files from the command line eg:

Yeah, that's what I was thinking.

Right now the API expects a JSON map of filename to contents. That would mean using a `map[string]string` serialized to `json:"solution"`, but I could change it to expect the serialized `[]File` if that seems cleaner.

It's a small change in one place, so really no big deal either way.

ClashTheBunny:
Just to make sure, the the path of the file would be nice to preserve in
the metedata.  Some languages give significance to deep directories.

kytrinyx:
Yeah, the path of the file is currently being preserved, and we need to continue to do so.

kytrinyx:
Closed in https://github.com/exercism/cli/pull/166

PR #166, opened 2015-03-11
Accept multiple files in submissions

kytrinyx:
This addresses #35


api/iteration_test.go
```diff
@@ -1,8 +1,57 @@
...
+		"lib/three.py": "# three\n",
+	}
+	for filename, code := range expected {
+		if iter.Solution[filename] != code {
+			t.Errorf("Expected %s to contain %s, had %s", filename, code, iter.Solution[filename])
+		}
+	}
+}
+
+/*
```
Tonkpils: What is the reason for this comment?

api/iteration.go
```diff
@@ -2,52 +2,74 @@ package api
...
 import (
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
 	"strings"
 )
 
 const (
-	msgUnidentifiable = "unable to identify track and problem"
+	errUnidentifiable = "unable to identify track and problem"
```
Tonkpils: Since we know these are errors and for the most part we are returning them with `fmt.Errorf(error)` why not just create them with `errors.New(...)` and return the error itself without `fmt.Errorf`?

api/iteration_test.go
kytrinyx: Oh, ouch, that was a left-over something-or-other. Let me figure out what happened and get rid of it.

api/iteration.go
kytrinyx: Agreed.

kytrinyx:
I've updated the PR to reflect @Tonkpils's comments.

Tonkpils:
This LGTM, how would this work on the CLI though `exercism submit file1 file2 ...`? or just a path to a directory?

kytrinyx:
Yeah, right now it's `file1`, `file2`

Tonkpils:
If there's no other comments on this we can merge after its been updated with master. :heart:

kytrinyx:
Rebased.

kytrinyx:
@hintjens I'm sorry this took forever. If you get the urge to implement some C exercises, we now have multiple file support in the CLI and app.

Issue #137, opened 2014-10-15
Add a "current" command?


kytrinyx:
https://github.com/exercism/exercism.io/issues/1972

When working in multiple languages, or when you come back to exercism after a spell away from it, it can be tricky to figure out what you were working on.

It's possible to sort directories by date (`ls -ct`), but it would be handy to be told directly what's still in the works.

Carpetsmoker:
If you use `fetch, it actually shows you which one to use:

    Not Submitted: 21 problems
                   Bob (clojure) /home/martin/exercism/clojure/bob
              Bob (coffeescript) /home/martin/exercism/coffeescript/bob
                       Bob (cpp) /home/martin/exercism/cpp/bob
	[...]

I think it would help a lot to just change the output slightly:

    Not Submitted: 21 problems
        clojure (Bob)            /home/martin/exercism/clojure/bob
        coffeescript (Bob)       /home/martin/exercism/coffeescript/bob
        cpp (Bob)                /home/martin/exercism/cpp/bob
	[...]

This makes it a lot easier to scan for the language you want, that's the first thing I usuaully want to know, I don't care so much about the text name.

Perhaps it could also be extended to show the test number, for example maybe like:

    Not Submitted: 21 problems
        clojure (Exercise 2: Bob)            /home/martin/exercism/clojure/bob
        coffeescript (Exercise 5: Bob)       /home/martin/exercism/coffeescript/bob
        cpp (Exercise 1: Bob)                /home/martin/exercism/cpp/bob
	[...]

kytrinyx:
The exercises aren't actually numbered (you can do them out of order by asking for them directly, and the order can change, and these changes aren't tracked).

I think we could sort them by the date of when the directory was created, which can work as a sort of proxy.

I think it makes a lot of sense to change the display to put the language first.

fluxusfrequency:
That will work in some cases, but not if someone has their exercises on github and clones them down to a second machine.

kytrinyx:
Yeah, that's true. I won't bother with the sorting thing.

Having thought about it, I still think that "current" would be useful, but I think that improving the ability to scan for language will help, so I'll fix that.

NobbZ:
I just skipped some exercises in various languages and now came to a point where I skipped the last exercise in a track. The cli seems to think that I have finished that track, and doesn't show it any more after I do `exercism f`. Perhaps display the first skipped exercise of the track again as the current unsubmitted one?

[...]

PR #169, opened 2015-03-15
Tweak output from fetch to make it more scannable

kytrinyx:
It's sometimes difficult to find the problem(s) that you're
currently working on. The previous 'report' output format
ordered by the problem name, grouping "Bobs" in all the languages
together.

It seems like people tend to scan for the language, not the problem.

Makes sense in hindsight.

See:
- https://github.com/exercism/cli/issues/137
- https://github.com/exercism/exercism.io/issues/1972

I just realised how utterly awful I think the homework/report/item stuff is. I'm not sure what I was thinking when I wrote that. When I have finished traveling for a bit I'll take another look at rewriting or refactoring that.

### DIFF
--- a/user/homework.go
+++ b/user/homework.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/exercism/cli/api"
 	"github.com/exercism/cli/config"
@@ -46,7 +47,7 @@ func NewHomework(problems []*api.Problem, c *config.Config) *Homework {
 		hw.Items = append(hw.Items, item)
 	}
 
-	hw.template = fmt.Sprintf("%%%ds %%s\n", hw.maxTitleWidth())
+	hw.template = "%s%s %s\n"
 	return &hw
 }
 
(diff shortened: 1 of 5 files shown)


user/homework.go
```diff
@@ -46,7 +48,7 @@ func NewHomework(problems []*api.Problem, c *config.Config) *Homework {
 		hw.Items = append(hw.Items, item)
 	}
 
-	hw.template = fmt.Sprintf("%%%ds %%s\n", hw.maxTitleWidth())
+	hw.template = fmt.Sprintf("%%s%%s %%s\n")
```
lcowell: As there are no substitutions, you could now reduce this to:

```go
hw.template = "%s%s %s\n"
```

The previous versions generation of the template padding using `%20s` or whatever value besides 20 the actual max length feels more correct. That being said it definitely needed to be rewritten to be more understandable.

user/homework.go
```diff
@@ -32,6 +33,7 @@ const (
 type Homework struct {
 	Items    []*Item
 	template string
+	width    int
```
lcowell: `maxTitleWidth` is being called 2 times and both times from within `Homework.Report`. I think we should calculate the max width at the top of the `Report` function, store it in a local variable, and use that value in `Item.Report` and `hw.heading`. The reason for `width` being a property of `Homework` in not obvious and adds a piece of state that must now be maintained.

user/homework.go
kytrinyx: I agree that it feels more correct, but I couldn't find a way to put the spacing directly into the templating when I wanted to left-align the text.

user/homework.go
kytrinyx: Agreed!

kytrinyx: Also, I don't want to be printing output anywhere except in the cmd package. This fmt.Print stuff is so dirty here.

kytrinyx:
I've updated the code per your comments. I'll see about refactoring this to get rid of the weird reporting stuff in a separate PR.

Issue #172, opened 2015-03-26
EXERCISM_CONFIG_FILE env should handled in app configuration


lcowell:
We have logic in the config package to check the `EXERCISM_CONFIG_FILE` environment variable and the provided path to determine when the config file lives. This is actually something that we can get for free from codegangsta/cli. This would clean up the config package and expose the environment variable in the app help output.

Any thoughts on this before I implement it ?

Tonkpils:
I'd have to see the implementation to see what you mean. Or maybe point me in the direction of the pice of the CLI package the functionality that you're looking to implement?

Aside from that, I'm all for simplicity and readablilty and if this effort improved that, then I'm all for it!

Might this also be a good time to tackle #161 for the XDG_CONFIG_HOME?

kytrinyx:
Yeah, if we can lean on existing functionality in codegangsta/cli, that would be great.

lcowell:
@Tonkpils I'm referring to adding this type of option to our app config: https://github.com/codegangsta/cli#values-from-the-environment

```go
cli.StringFlag{
	Name:  "config, c",
	Usage: "path to config file",
	EnvVar: "EXERCISM_CONFIG_FILE",
},
```

I've built a couple of cli apps using it and it would work like this:
`command line option > environment variable > defaults`


I think the XDG_CONFIG_HOME would require a different approach. My understanding is that env variable points to a directory, so we'd still need to resolve it to a file path. These are closely related, so I think it only makes sense to move ahead with #172 if it doesn't complicate moving ahead with #161

I think this is doable because `cli` gives us the chance to process input flags before calling a command.

Tonkpils:
Looks good to me :+1:

kytrinyx:
Yepp, me too. Go for it.

PR #173, opened 2015-03-30
ENV based config path handed in main app

lcowell:
codegangsta/cli can handle cascading configuration options which means
we don't need to handle the EXERCISM_CONFIG_FILE environment variable
in the config package.

See Issue #172 for more information.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -10,7 +10,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


Tonkpils:
:+1: Looks good to me

kytrinyx:
Nice!

PR #177, opened 2015-04-02
Provide a more robust error message if JSON is invalid.

### LINKED ISSUE #170: Provide a useful error message if the config JSON is invalid
just installed manually into my path on a Win64 machine running Windows 7.

````
$ exercism list
2015/03/22 19:51:00 invalid character 'e' in string escape code

$ exercism -v list
2015/03/22 19:51:51 invalid character 'e' in string escape code

$ exercism --version
exercism version 2.0.0

$ exercism configure
2015/03/22 20:04:44 invalid character 'e' in string escape code
````

### PR DESCRIPTION
**Pull request by Tonkpils**
Fixes #170 

With `json.SyntaxError` we also have the `Offset` at which the error is, so we can highlight where the syntax error is. I can work on that if that's something we want but this should be enough to get going.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -140,8 +141,15 @@ func (c *Config) read() error {
 	}
 	defer f.Close()
 
-	d := json.NewDecoder(f)
-	return d.Decode(&c)
+	if err := json.NewDecoder(f).Decode(&c); err != nil {
+		var extra string
+		if _, ok := err.(*json.SyntaxError); ok {
+			extra = ":\nThe file contains invalid JSON syntax"
+		}
+		return fmt.Errorf("error parsing JSON in the config file %s%s: %s", f.Name(), extra, err)
+	}
+
+	return nil
 }
 
 // IsAuthenticated returns true if the config contains an API key.
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Tonkpils** [comment]
Current output looks like this:

```
error parsing JSON in the config file ~/config_invalid.json:
The file contains invalid JSON syntax: invalid character 'e' in string escape code
```

Here's an example output of how it would look if we were to highlight the syntax error

```
error parsing JSON in the config file ~/config_invalid.json:
The file contains invalid JSON syntax at ':"D:\\Users\e' <~
invalid character 'e' in string escape code
```

Thoughts? Not sure if it adds any more value.

**Comment by mtakac** [comment]
Would it be possible to add full path to the config and number of line causing the error?

**Comment by jwood803** [comment]
Each bit more of info you can supply to folks always helps, I say. :]

**Comment by Tonkpils** [comment]
> Would it be possible to add full path to the config and number of line causing the error?

@mtakac The full path will be displayed, I changed it for the example but it would look like 

```
error parsing JSON in the config file /var/folders/_2/0wpr12717gg2g8btfc3fvr4w0000gn/T/485490851/config_invalid.json
```

Adding the line number should be trivial but the issue is that the config file is usually just 1 line unless the user modifies it to be pretty. 

> Each bit more of info you can supply to folks always helps, I say. :]

Definitely! Though I wanted to get feedback in case too much info can be overwhelming for the user but I guess it couldn't hurt in this scenario since they're already running into an issue.

**Comment by mtakac** [comment]
:+1: looks great and clear.

**Comment by Tonkpils** [comment]
The latest commit makes the error output:

```
2015/04/02 19:42:21 error parsing JSON in the config file /Users/tonkpils/.exercism.json:
The file contains invalid JSON syntax at 'm.io", "foo"}' <~
invalid character '}' after object key
```

kytrinyx:
We could also write the config file pretty--it's not costly for such a small file, and if it would make it easier to give good error messages, that's a plus.

Tonkpils:
@kytrinyx sure, do you want me to get that in with this PR or in a different one?

kytrinyx:
`pretty` can happen in a separate PR, I was just thinking out loud :)

kytrinyx:
Thanks @Tonkpils!

PR #178, opened 2015-04-02
Allow $XDG_CONFIG_HOME to configure config location

### LINKED ISSUE #161: respect $XDG_CONFIG_HOME
@JulianAndrews makes a good point about supporting XDG_CONFIG_HOME on systems where it's set.

See https://github.com/exercism/exercism.io/issues/2212

### LINKED ISSUE #176: Provide a useful error message if the EXERCISM_CONFIG_FILE points to a directory
Someone reported the following:

> I get a message saying 'The handle is invalid' when doing anything

It turns out they had put a folder as the EXERCISM_CONFIG_FILE instead of the actual file

lcowell:
If $XDG_CONFIG_HOME (or any mechanism for setting the config path) points to a directory, append the default filename to the provided path. Resolves #161 and possibly resolves #176

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -167,7 +167,14 @@ func (c *Config) resolvePath(argPath string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.Replace(path, "~", h, 1), nil
+	path = strings.Replace(path, "~", h, 1)
+
+	fi, _ := os.Stat(path)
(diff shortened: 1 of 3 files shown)


Tonkpils:
This looks good, I think it will fix #176 if it's a directory. :+1:

kytrinyx:
Yeah, this looks promising. Thanks :)

PR #182, opened 2015-04-04
Deal with different file encodings when submitting solutions

ambroff:
Fix for https://github.com/exercism/exercism.io/issues/2303

PR #184, opened 2015-04-05
[RECREATED] Deal with different file encodings when submitting solutions

ambroff:
Recreating https://github.com/exercism/cli/pull/182 because it is stuck closed after I recreated the branch.

Fixes exercism/exercism.io#2303


.travis.yml
```diff
@@ -7,7 +7,7 @@ go:
 install: true
 
 before_install:
-  go get github.com/codegangsta/cli && go get github.com/stretchr/testify/assert
+  bin/deps
```
Tonkpils: Would `go get -t ./...` suffice here? It would fetch all dependencies for each package.

kytrinyx: Yeah, probably. I'll pull this in and then tweak that in a branch to make sure it works.

kytrinyx:
Thanks!

PR #186, opened 2015-04-24
Allow specifying demo directory

### LINKED ISSUE #180: exercise download directory defaults to home dir
This should be ask where you want them, then update config file. This should happen if there is no config file.

Tonkpils:
Fixes #180 for the demo command.


kytrinyx:
This looks great. For a moment I was confused about the `c.SetDir(c.Dir)`, but I realized that this lets us sanity-check and have good fallbacks. Makes sense!

PR #189, opened 2015-04-26
Allow submitting test file with --test flag

### LINKED ISSUE #188: Warn instead of blocking when submitting /test/
See https://github.com/exercism/xgo/pull/168 for a discussion

pminten:
This is a bit of a different approach than we discussed earlier. It still blocks accidental test submissions but gives instructions for how to submit tests if that's what you want.

Closes #188


kytrinyx:
Nice, I think this works well.

PR #190, opened 2015-05-10
Compare home path expansion using OS specific separator

Tonkpils:
Fixes https://github.com/exercism/exercism.io/issues/2393

I was able to test it on a windows machine and it worked fine. I'll have to think of a way to write a unit test for it though.


kytrinyx:
Aaah, of course. I keep running into this, and I keep forgetting again!

PR #195, opened 2015-05-20
Print release info on debug

Tonkpils:
This prints out the latest release version as part of the debug and it informs the user if their client is currently outdated and where they can find the latest release. 

This is in the interim while I work on the `upgrade` functionality. 

The wording could use some help though. 

This is mostly for review and thoughts. If we want the upgrade functionality instead of this I'll have it ready fairly soon.

### DIFF
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -38,7 +38,7 @@ func Debug(ctx *cli.Context) {
 		log.Println(err)
 	} else {
 		if rel.Version() != ctx.App.Version {
-			defer fmt.Println("\nYour CLI is outdated. A new release can be found here:", rel.Location)
+			defer fmt.Printf("\nA newer version of the CLI (%s) can be downloaded here: %s\n", rel.TagName, rel.Location)
 		}
 		fmt.Printf("Exercism CLI Latest Release: %s\n", rel.Version())
 	}
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -1,23 +1,48 @@
 package cmd
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/codegangsta/cli"
 	"github.com/exercism/cli/config"
 )
 
+type release struct {
+	Location string `json:"html_url"`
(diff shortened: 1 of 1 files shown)


kytrinyx:
This is great. I'd be happy to put this in the next point release (which I fully intend to cut in the next few days).

The upgrade functionality would be great, but in the meanwhile this would be a really nice help.

cmd/debug.go
```diff
@@ -1,23 +1,48 @@
...
+
 	fmt.Printf("\n**** Debug Information ****\n")
 	fmt.Printf("Exercism CLI Version: %s\n", ctx.App.Version)
+
+	rel, err := checkLatestRelease(client)
+	if err != nil {
+		log.Println(err)
+	} else {
+		if rel.Version() != ctx.App.Version {
+			defer fmt.Println("\nYour CLI is outdated. A new release can be found here:", rel.Location)
```
kytrinyx: Maybe _There is a more recent version of the CLI here:_ instead of saying that yours is outdated. Or maybe _A newer version of the CLI (vX.Y.Z) can be downloaded here:_

cmd/debug.go
```diff
@@ -1,23 +1,48 @@
...
+		}
+		fmt.Printf("Exercism CLI Latest Release: %s\n", rel.Version())
```
kytrinyx: I really like having this line in the debug info.

cmd/debug.go
```diff
@@ -1,23 +1,48 @@
...
+
+func (r *release) Version() string {
+	return strings.TrimPrefix(r.TagName, "v")
+}
+
 // Debug provides information about the user's environment and configuration.
 func Debug(ctx *cli.Context) {
 	defer fmt.Printf("\nIf you are having trouble and need to file a GitHub issue (https://github.com/exercism/exercism.io/issues) please include this information (except your API key. Keep that private).\n")
 
+	client := http.Client{Timeout: 5 * time.Second}
```
lcowell: I came by to suggest a timeout, but you're way ahead of me :+1: This is great.

cmd/debug.go
Tonkpils: Sounds good, I'll change it to the latter one! Thanks!

PR #196, opened 2015-06-18
Upgrade command

### LINKED ISSUE #61: Add `exercism upgrade` or `exercism update`
...

Tonkpils:
This is a working prototype of the `upgrade` command for the CLI.

It will install the latest version of the CLI under the same path as the executable that called the command. 

One thing that it does not support yet that I'd like to fix before this gets merged:

1. **ARM version upgrades**
  Golang has no way of checking the ARM version during runtime
    - A possible solution would be to set the ARM version during the build process with `-ldflags` which would allow us to set the binary's OS/ARCH/ARMV that was used during build. 

Also, I wrote the windows portion a bit blindly since I don't have a windows machine to test it on. I'll make sure it works properly when I get a chance. 

I'd also like to include a generic message to file an issue here if the install fails with the `debug` info but not too sure how the message should read. Thoughts?

Once this is merged it fixes #61

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -266,6 +266,7 @@ func (c *Client) Tracks() ([]*Track, error) {
 	return payload.Tracks, nil
 }
 
+// Skip sends a request to exercism to skip the exercise given language and slug
 func (c *Client) Skip(language, slug string) error {
 	url := fmt.Sprintf("%s/api/v1/iterations/%s/%s/skip?key=%s", c.APIHost, language, slug, c.APIKey)
 
--- a/bin/build-all
+++ b/bin/build-all
@@ -5,51 +5,75 @@ set -e -x
 echo "Creating release dir..."
 mkdir -p release
 
-echo "Creating darwin/386 binary..."
-GOOS=darwin GOARCH=386 go build -o out/exercism exercism/main.go
-cd out
-tar cvzf ../release/exercism-mac-32bit.tgz exercism
-cd ..
+# variables as defined by "go tool nm"
+OSVAR=github.com/exercism/cli/cmd.BuildOS
+ARCHVAR=github.com/exercism/cli/cmd.BuildARCH
+ARMVAR=github.com/exercism/cli/cmd.BuildARM
 
-echo "Creating darwin/amd64 binary..."
-GOOS=darwin GOARCH=amd64 go build -o out/exercism exercism/main.go
-cd out
-tar cvzf ../release/exercism-mac-64bit.tgz exercism
-cd ..
+createRelease() {
+	os=$1
+	arch=$2
+	arm=$3
 
-echo "Creating linux/386 binary..."
-GOOS=linux GOARCH=386 go build -o out/exercism exercism/main.go
-cd out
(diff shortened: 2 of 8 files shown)


kytrinyx:
This is so cool! I'm at work, so can't read through it now. I think setting the arm version with a build flag is perfect.

Tonkpils:
Ok, I re-wrote the `bin/build-all` script to make it set the variables during build. I've tested the upgrade command in linux, windows and mac. Also, if you see any improvements that can be done to the build script let me know.

api/api.go
```diff
@@ -266,6 +266,7 @@ func (c *Client) Tracks() ([]*Track, error) {
 	return payload.Tracks, nil
 }
 
+// Skip sends a request to exercism to skip the exercise given language and slug
```
kytrinyx: Yay, doc comments :)

bin/build-all
```diff
@@ -5,51 +5,75 @@ set -e -x
...
+createRelease linux 386
+createRelease linux amd64
+
+# ARM Releases
+createRelease linux arm 5
+createRelease linux arm 6
+
+# Windows Releases
+createRelease windows 386
+createRelease windows amd64
```
kytrinyx: This whole thing is so much nicer!

cmd/upgrade.go
```diff
@@ -0,0 +1,209 @@
...
+// Upgrade command allows the user to upgrade to the latest CLI version
+func Upgrade(ctx *cli.Context) {
+	client := http.Client{Timeout: 5 * time.Second}
+	rel, err := checkLatestRelease(client)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// TODO: This checks strings against string
+	// Version 2.2.0 against release 2.1.0
```
kytrinyx: The TODO comment is a bit unclear. Do you mean that we should check semver and only auto upgrade minor or patch versions automatically?

cmd/upgrade.go
```diff
@@ -0,0 +1,209 @@
...
+	// will trigger an upgrade...
+	// should probably parse semver and compare
+	if rel.Version() == ctx.App.Version {
+		fmt.Println("Your CLI is up to date!")
+		return
+	}
+
+	// current executable path
+	dest, err := osext.Executable()
```
kytrinyx: I was wondering how we would do that. Very nice!

cmd/upgrade.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+type asset struct {
+	ID          int    `json:"id"`
+	Name        string `json:"name"`
+	ContentType string `json:"content_type"`
+}
+
+func (a *asset) Download() (*bytes.Reader, error) {
+	downloadURL := fmt.Sprintf("https://api.github.com/repos/exercism/cli/releases/assets/%d", a.ID)
```
kytrinyx: Oh, wow. This is way better than what I'm doing in the bash install script (which mucks around with location headers).

kytrinyx:
This looks great. I'm going to merge it in (the only thing that I was wondering about is a comment, and that can be tweaked later, I think).

PR #200, opened 2015-07-23
Add a guard to the unsubmit command.

kytrinyx:
The unsubmit command does not take any arguments, but it would be
easy to assume that it does, considering how most of the other commands
work.

This changes the command so that it stops with a warning if someone
tries to pass an argument.

This change is in response to https://github.com/exercism/exercism.io/issues/2446
where someone tried unsubmitting a submission by passing the file path
to the unsubmit command, which blithely unsubmitted the most recent
submission--period.

It's confusing and unfriendly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The exercism CLI follows [semantic versioning](http://semver.org/).
 
 ## Next Release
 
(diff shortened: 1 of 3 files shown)


Tonkpils:
This looks great!

kytrinyx:
Thanks, @Tonkpils!

PR #208, opened 2015-08-03
avoid ambiguous or unresolvable exercism paths

### LINKED ISSUE #205: Instructions to submit exercise could be improved.
I was trying exercism in the morning, submitting doesn't work.
 I had a relative path to the exercism folder
Maybe put a message when you set a path that it should be a full path and not a relative path.

lcowell:
Prepend any relative path with the current working directory so the path can be absolute and will be resolvable no matter what directory you're in.

Fixes #205

### DIFF
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -23,8 +23,7 @@ func Configure(ctx *cli.Context) {
 	dir := ctx.String("dir")
 	api := ctx.String("api")
 
-	err = c.Update(key, host, dir, api)
-	if err != nil {
+	if err = c.Update(key, host, dir, api); err != nil {
 		log.Fatalf("Error updating your configuration %s\n", err)
 	}
 
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -22,7 +22,11 @@ func Configure(ctx *cli.Context) {
 	host := ctx.String("host")
 	dir := ctx.String("dir")
 	api := ctx.String("api")
-	c.Update(key, host, dir, api)
+
+	err = c.Update(key, host, dir, api)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/configure.go
```diff
@@ -22,7 +22,11 @@ func Configure(ctx *cli.Context) {
 	host := ctx.String("host")
 	dir := ctx.String("dir")
 	api := ctx.String("api")
-	c.Update(key, host, dir, api)
+
+	err = c.Update(key, host, dir, api)
```
Tonkpils: Can we scope this in the `if` statement like so.

```Go
if err := c.Update(...); err != nil {
```

config/config.go
```diff
@@ -89,13 +90,18 @@ func (c *Config) Update(key, host, dir, xapi string) {
 
 	dir = strings.TrimSpace(dir)
 	if dir != "" {
-		c.SetDir(dir)
+		err = c.SetDir(dir)
+		if err != nil {
```
Tonkpils: Ditto for this, we don't have to declare the `var err error` above if we scope the error in here.

config/config.go
```diff
@@ -250,16 +256,31 @@ func (c *Config) setDefaults() error {
 // or defaults to the home exercism directory
 func (c *Config) SetDir(path string) error {
 	home, err := c.homeDir()
+	var dir string
```
Tonkpils: Let's move this to L263 to not allocate it if there's an error on `homeDir`.

lcowell:
Thanks @Tonkpils - I've implemented your suggestions.

Tonkpils:
This is a great fix! Thanks!

kytrinyx:
+1000!

PR #214, opened 2015-08-30
Project Updates

Tonkpils:
This removes a couple of instructions on how to deal with outdated errors and things that you'd otherwise run into if you're running old versions of Go.

This stems off from #213. 

I'd like to get some feedback if possible.

Another thing that I'd like to discuss as part of this PR is vendoring of packages that we need. Go 1.5 shipped with experimental vendoring which could make development much easier. We could give it a try now or use a more established tool like `gb` or `godep` and re-iterate when vendoring is 100% in Go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: go
 sudo: false
 
 go:
-  - 1.4.1
   - 1.5
   - tip
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@ that if you're doing the Haskell problems on exercism you don't need a working
 Python or Ruby environment simply to fetch and submit exercises.
 
 ## Dependencies
-Go version 1.3 or higher
+
+Go version 1.5 or higher
 
 ## Installing Go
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -22,15 +23,11 @@ Follow the directions on http://golang.org/doc/install
 1. `go get github.com/exercism/cli/exercism`
 1. `cd $GOPATH/src/github.com/exercism/cli`
 1. `git remote set-url origin https://github.com/<your-github-username>/cli`
-1. `go get -t ./...`*)
-1. Make sure $GOPATH/bin is on your path (you may need something like `export PATH=$PATH:/projects/goprojects/bin`)
-1. Open a separate terminal window to your project directory and run the command `glitch`
+1. `go get -t ./...`
+1. Make sure `$GOPATH/bin` is on your path (you may need something like `export PATH=$PATH:/projects/goprojects/bin`)
```
Tonkpils: I'm not sure that we need this step for development.

.travis.yml
```diff
@@ -4,6 +4,7 @@ sudo: false
 
 go:
   - 1.4.1
```
Tonkpils: Should we remove 1.4.1?

README.md
lcowell: :+1:

.travis.yml
lcowell: Yes, and let's update the readme to say that we're assuming they are using 1.5.

lcowell:
I've used `gb` for a couple of projects and it seemed to work well. `gb` supports cross compilation as well which would clean up our build process.
http://getgb.io/news/gb-version-0.1.0-released/

kytrinyx:
Sure, I could go either way -- experimental 1.5 vendoring, or `gb`. I've only tried using `godeps` once, and wasn't super happy with how the whole process felt. That could be a failing in me, though, not the library :)

Tonkpils:
Cool, I'll update the PR with the suggestions when I have a chance over the weekend.

Issue #223, opened 2015-10-13
Implement new command: exercism status [$language]


kytrinyx:
Instead of implementing a `current` and a `skipped` command, let's have a `status` command that lists the following:

For each language where you have submitted at least one iteration of one exercise:

* list the (one) exercise that you most recently submitted, and when
* list the exercises that you have fetched but not submitted (could be multiple)
* list the exercises that you have skipped

We will need to implement an endpoint in the exercism/exercism.io api (currently at `lib/api/routes/...`).

If someone calls `exercism status $language` for a language where they have not submitted anything, we should explicitly tell them that they've not submitted anything in that language.

This issue supersedes #210 and #137.

devonestes:
I might be interested in taking this one on! Considering this is dependent on the API endpoint, and that's in a different repo, would it be better to take care of that endpoint first and then develop the CLI command once the endpoint is merged?

Tonkpils:
@devonestes you could also implement the API and then develop on the CLI against your local instance of exercism with the API changes.

If you have any questions don't hesitate reaching out.

kytrinyx:
This is now implemented, and will be a part of the next release.

PR #228, opened 2015-10-30
WIP - Adding `exercism status $language` command

devonestes:
This is the first commit in my attempt to implement the new `status` feature as described in #223. Since this is my first time contributing to this project, I'm pushing this up now mainly to get feedback on if this is the right way to go with this feature before I go clean things up and move on to next steps. At the moment the I am basically mocking out an API response so I can make sure the command works as intended.

Below is a screenshot of the output to the user as it stands now.
status_screenshot

I'd love any feedback anyone has to offer. Specifically, I'd like to know if the API endpoint I'm looking to use looks like a good choice, if the API response format (in comments right now in `api.go` seems like a good choice, and also if there is any standardized format for outputting the information to the user that I should be using instead of the format that I have implemented now.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -273,3 +273,28 @@ func (c *Client) Skip(trackID, slug string) error {
 
 	return errors.New(pe.Error)
 }
+
+// Status sends a request to exercism to fetch the user's
+// completion status for the given language track.
+func (c *Client) Status(trackID string) (*StatusInfo, error) {
+	url := fmt.Sprintf("%s/api/v1/tracks/%s/status?key=%s", c.APIHost, trackID, c.APIKey)
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := c.Do(req, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+
+	var si StatusInfo
+	if err := json.NewDecoder(res.Body).Decode(&si); err != nil {
+		return nil, err
+	}
+
+	return &si, nil
+}
--- a/api/status_info.go
+++ b/api/status_info.go
@@ -1,0 +1,64 @@
+package api
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+)
+
+const dateFormat = "January 2, 2006"
+
+// StatusInfo contains information about a user's status on a particular language track.
+type StatusInfo struct {
+	TrackID         string `json:"track_id"`
+	Recent          *Recent
+	FetchedProblems *Slugs `json:"fetched"`
+	SkippedProblems *Slugs `json:"skipped"`
+}
+
+// Recent contains information about the user's most recently submitted exercise on a particular language track.
+type Recent struct {
+	Problem     string `json:"problem"`
+	SubmittedAt string `json:"submitted_at"`
+}
+
+// Slugs is a collection of slugs, all of which are the names of exercises.
+type Slugs []string
+
+func (r *Recent) String() string {
+	submittedAt, err := time.Parse(time.RFC3339Nano, r.SubmittedAt)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	return fmt.Sprintf(" - %s (submitted on %s)", r.Problem, submittedAt.Format(dateFormat))
+}
+
+func (s *StatusInfo) String() string {
+	if len(*s.FetchedProblems) == 0 && s.Recent.Problem == "" {
+		return fmt.Sprintf("\nYou have yet to begin the %s track!\n", s.TrackID)
+	}
+
+	msg := `
+Your status on the %s track:
+
+Most recently submitted exercise:
+%s
+
+Exercises fetched but not submitted:
+%s
+
+Exercises skipped:
+%s
+`
+
+	return fmt.Sprintf(msg, s.TrackID, s.Recent, s.FetchedProblems, s.SkippedProblems)
+}
+
+func (s Slugs) String() string {
+	for i, problem := range s {
+		s[i] = fmt.Sprintf(" - %s", problem)
+	}
+	return strings.Join(s, "\n")
+}
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -1,0 +1,32 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/codegangsta/cli"
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
+)
+
+// Status is a command that allows a user to view their progress in a given
+// language track.
+func Status(ctx *cli.Context) {
+	c, err := config.New(ctx.GlobalString("config"))
+	if err != nil {
+		log.Fatal(err)
+	}
+	args := ctx.Args()
+
+	if len(args) != 1 {
+		log.Fatal("Usage: exercism status LANGUAGE")
+	}
+
+	client := api.NewClient(c)
+	status, err := client.Status(args[0])
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Println(status)
+}
--- a/exercism/main.go
+++ b/exercism/main.go
@@ -30,6 +30,7 @@ const (
 	descOpen      = "Opens exercism.io to your most recent iteration of a problem given the track ID and problem slug."
 	descDownload  = "Downloads a solution given the ID of the latest iteration."
 	descList      = "Lists the available problems for a language track, given its ID."
+	descStatus    = "Fetches information about your progress with a given language track."
 
 	descLongRestore  = "Restore will pull the latest revisions of exercises that have already been submitted. It will *not* overwrite existing files. If you have made changes to a file and have not submitted it, and you're trying to restore the last submitted version, first move that file out of the way, then call restore."
 	descLongDownload = "The submission ID is the last part of the URL when looking at a solution on exercism.io."
@@ -166,6 +167,12 @@ func main() {
 			ShortName: "li",
 			Usage:     descList,
 			Action:    cmd.List,
+		},
+		{
+			Name:      "status",
+			ShortName: "st",
+			Usage:     descStatus,
+			Action:    cmd.Status,
 		},
 	}
 	if err := app.Run(os.Args); err != nil {


Tonkpils:
:+1: This looks great so far

api/api.go
```diff
@@ -292,3 +292,44 @@ func (c *Client) Skip(language, slug string) error {
...
+		return nil, err
+	}
+
+	res, err := c.Do(req, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusNoContent {
```
kytrinyx: So a 204 means that it's all good but you haven't done anything? That seems a little bit cryptic.

api/api.go
```diff
@@ -292,3 +292,44 @@ func (c *Client) Skip(language, slug string) error {
...
+	}
+	//
+	//
+	//
+	//
+	//
+	// API response format
+	// {
+	//   "recent":"leap",
+	//   "submitted":"Jan 1, 2015",
```
kytrinyx: I think we should have the track id in the response for completeness, even though you have to know the language in order to make the API call.

Also, let's make "submitted" a timestamp called "submitted_at", and use the iso8601 format for it. The API shouldn't have to know when we want timestamps vs when we just want the date.

Lastly, I think we should put the recent/submitted together so that it's clear what the submitted timestamp means:

```
{
  "track_id": "go",
  "recent": {
    "problem": "leap",
    "submitted_at": "2015-01-01T14:31:57Z"
  },
  "fetched": ["gigasecond", "raindrops"],
  "skipped": ["clock", "poker"]
}
```

api/api.go
```diff
@@ -292,3 +292,44 @@ func (c *Client) Skip(language, slug string) error {
...
+	//   "fetched":["gigasecond", "raindrops"],
+	//   "skipped":["clock", "poker"]
+	// }
+
+	recent := "leap"
+	submittedOn := "Jan 1, 2015"
+	fetched := []string{"gigasecond", "raindrops"}
+	skipped := []string{"clock", "poker"}
```
kytrinyx: Go ahead and fake out the response by creating an actual JSON string and decoding it into a struct--that will get you one step closer without actually having to have an API endpoint.

api/api.go
```diff
@@ -292,3 +292,44 @@ func (c *Client) Skip(language, slug string) error {
...
+
+	return [][]string{[]string{recent, submittedOn}, fetched, skipped}, nil
```
kytrinyx: Is there a particular reason to return a slice of slice here? I think this should just be a custom struct.

cmd/status.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"log"
+
+	"github.com/codegangsta/cli"
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
+)
+
+// Status command allows a user to view their progress in a given language track
```
kytrinyx: According to the [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) wiki, doc comments should be full sentences, ending with a period.

cmd/status.go
```diff
@@ -0,0 +1,53 @@
...
+		log.Fatal(err)
+	}
+
+	printResponse(language, status)
+}
+
+func printResponse(language string, status [][]string) {
+	recent := status[0]
+	fetched := status[1]
+	skipped := status[2]
```
kytrinyx: Yeah, using a custom struct type for status will make this all much simpler to deal with.

api/api.go
```diff
@@ -292,3 +292,44 @@ func (c *Client) Skip(language, slug string) error {
...
+
+	res, err := c.Do(req, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusNoContent {
+		msg := fmt.Sprintf("You have not yet begun the %v track", language)
+		return nil, errors.New(msg)
```
kytrinyx: This isn't really an error, it's a status of "not started yet".

cmd/status.go
```diff
@@ -0,0 +1,53 @@
...
+
+	fmt.Println("\nExercises fetched but not submitted")
+	for _, exercise := range fetched {
+		fmt.Println(exercise)
+	}
+	fmt.Println("\nExercises skipped")
+	for _, exercise := range skipped {
+		fmt.Println(exercise)
+	}
+	fmt.Println("")
```
kytrinyx: If we have a custom status type, then we can move the printing/formatting into a String() function and do a `fmt.Println(status)`.

api/api.go
```diff
@@ -292,3 +303,30 @@ func (c *Client) Skip(language, slug string) error {
 
 	return errors.New(pe.Error)
 }
+
+// Status sends a request to exercism to fetch the user's
+// completion status for the given language track.
+func (c *Client) Status(language string) (*StatusInfo, error) {
+	c.XAPIHost = "http://localhost:9393"
```
devonestes: This is for testing on my local machine with the API serving mock data, and needs to be removed once this is in good enough shape and ready for merging.

cmd/status.go
```diff
@@ -0,0 +1,54 @@
...
+	client := api.NewClient(c)
+	status, err := client.Status(language)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	printResponse(status)
+}
+
+func printResponse(status *api.StatusInfo) {
```
devonestes: Still need to move this functionality from this standalone function to a method on StatusInfo. My major question, though, is where I should define that method - it can't be defined here because the type is defined in the `api` package, but I feel like adding it to the `api.go` file will be clutter. Should I maybe create a separate file for the type and endpoint processing of that particular command?

devonestes:
Ok, I think this is nearly there. I have pretty much everything working (I think), and I just need to implement the actual API endpoint. Right now I've been testing it on my local with the API serving up mock data for a language which a user has started, and also for a language which the user has not started.

The current output looks like this for both states:
status_update

If there is any more feedback either about functionality or about ways to improve the code, I'd love to hear it! If you think this is ready to roll, @kytrinyx, then I'll remove the last bit of code for the testing and re-push this up for final approval.

Tonkpils:
For consistency, could you add a colon at the end of each section? It would also improve output readability if the items in each section were preceded by a dash or some sort of bullet point for listing. Also, with receiving the iso format of the date, we could format it now to display something more user friendly. Of course, these comments are subject to feedback as well so let me know what you think.

api/api.go
kytrinyx: For reference, you can edit the config file to edit the xapi host (by default, that's `~/.exercism.json`, unless you have one of the environment variables defined).

kytrinyx:
+1 to all of @Tonkpils's suggestions. I really liked the formatting of the date that you had previously, I just wanted the API to be consistent about using timestamps.

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
 	State string `json:"state"`
 }
 
+// StatusInfo contains information about a user's status on a particular language track.
+type StatusInfo struct {
+	Language string `json:"track_id"`
```
kytrinyx: Do we call this `Language` elsewhere, instead of `TrackID`?

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
...
+	Recent   struct {
+		Problem   string `json:"problem"`
+		Submitted string `json:"submitted_at"`
```
kytrinyx: I would call this `At` or `SubmittedAt` -- my naming convention for timestamps is `*_at` and my naming convention for dates is `*_on`.

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
...
+	}
+	Fetched []string `json:"fetched"`
+	Skipped []string `json:"skipped"`
```
kytrinyx: Lately I've been trying to name collections as the name of what's in them, in the plural. That might not be a good choice here, because we'll end up with two long names where the suffix is repeated. I'm open to suggestions, though.

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
...
+}
+
+func (s *StatusInfo) String() {
+	if len(s.Fetched) == 0 && s.Recent.Problem == "" {
+		fmt.Println("\nYou have yet to begin the", s.Language, "track!")
+	} else {
```
kytrinyx: In Go the idiom is to ditch the else if there is a return in the `if` block (or if you could have a return in the if block). This minimizes indentation. [Effective Go](https://golang.org/doc/effective_go.html#if) says:

> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

Here there's no behavior that happens after you've printed this line, which would let you minimize the indentation in the next part.

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
 	State string `json:"state"`
 }
 
+// StatusInfo contains information about a user's status on a particular language track.
+type StatusInfo struct {
+	Language string `json:"track_id"`
+	Recent   struct {
+		Problem   string `json:"problem"`
```
kytrinyx: If we make this a proper type, then Recent could be a pointer to that type.

What do you think @Tonkpils -- would that make it a bit cleaner?

api/api.go
```diff
@@ -37,6 +39,43 @@ type SubmissionInfo struct {
...
+	} else {
+		fmt.Println("\nYour status on the", s.Language, "track:")
+
+		fmt.Println("\nMost recently submitted exercise")
+
+		submittedAt, err := time.Parse(time.RFC3339Nano, s.Recent.Submitted)
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Printf("%v (submitted on %v)\n", s.Recent.Problem, submittedAt)
```
kytrinyx: If you make `Recent` a type, then it could do some of its own reporting.

cmd/status.go
```diff
@@ -0,0 +1,32 @@
...
+	args := ctx.Args()
+
+	if len(args) != 1 {
+		msg := "Usage: exercism status LANGUAGE"
+		log.Fatal(msg)
+	}
+
+	language := args[0]
+	client := api.NewClient(c)
+	status, err := client.Status(language)
```
kytrinyx: `language` is only ever used once. I would probably just dereference the slice as you pass it to `Status`.

api/api.go
devonestes: That seemed to be the variable used when describing the language when the API endpoint URLs were being build in many of the functions - specifically `Fetch` (line 83), `Submission` (line 155), `List` (line 240) & `Skip` (line 309). However, after doing a search in the project I now see that `TrackID` is used in the other custom types created for `Problem` and `Submission`. Maybe it would be beneficial to change all of those variables using `language` in the functions for API calls to `trackId` to be consistent?

api/api.go
devonestes: I think `FetchedProblems` and `SkippedProblems` are good and descriptive names here. I'll go with that!

api/api.go
devonestes: @kytrinyx @Tonkpils My reasoning behind not making this a separate type now was that I didn't see any current use for that type outside of this, and by removing it I though it would be both premature optimization and potentially confusing. However, the performance benefit of using the pointer instead of the value seems like it might be worth the change. I'll await further input before pulling that out, though.

api/api.go
devonestes: This is another good reason to make `Recent` its own type. I think give this and the input above, I'll go ahead and do that now.

devonestes:
Thank you @kytrinyx & @Tonkpils for the great feedback so far! I feel like this has been a really great Go learning experience for me, and I'm really grateful that you've taken the time to give so much consideration to my code!

I'm starting to feel like it's in pretty good shape now. I've implemented all the recommended changes, and here's a screenshot of the output as it stands now:
status_update_2

api/api.go
```diff
@@ -292,3 +292,28 @@ func (c *Client) Skip(language, slug string) error {
 
 	return errors.New(pe.Error)
 }
+
+// Status sends a request to exercism to fetch the user's
+// completion status for the given language track.
+func (c *Client) Status(language string) (*StatusInfo, error) {
```
devonestes: I guess the one thing that might still want to change is if we want to rename these variables across all the `Client` methods to be `trackId` or something like that to be consistent with the field names in the various types that use that information.

api/api.go
kytrinyx: Yeah, let's do that in a separate commit. How about just making this one `TrackID` for now? (also, note: `TrackID`, not `TrackId`. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

api/status_info.go
```diff
@@ -0,0 +1,60 @@
...
+	Problem     string `json:"problem"`
+	SubmittedAt string `json:"submitted_at"`
+}
+
+func (r *Recent) String() string {
+	submittedAt, err := time.Parse(time.RFC3339Nano, r.SubmittedAt)
+	if err != nil {
+		log.Fatal(err)
+	}
+	year, month, day := submittedAt.Date()
```
kytrinyx: This should use https://golang.org/pkg/time/#Time.Format -- let's define our own pkg const.

api/status_info.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return fmt.Sprintf(" - %v (submitted on %v %v, %v)", r.Problem, month, day, year)
```
kytrinyx: This should use `%s` for the problem, and then also `%s` for the formatted time, once it's using a layout.

api/status_info.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (s *StatusInfo) String() string {
+	if len(s.FetchedProblems) == 0 && s.Recent.Problem == "" {
+		return fmt.Sprintf("\nYou have yet to begin the %v track!\n", s.TrackID)
```
kytrinyx: This should be `%s` as well.

api/status_info.go
```diff
@@ -0,0 +1,60 @@
...
+Exercises skipped:
+%v
+`
+
+	return fmt.Sprintf(msg, s.TrackID, s.Recent, formatProblems(s.FetchedProblems), formatProblems(s.SkippedProblems))
+}
+
+func formatProblems(problems []string) string {
+	for i, problem := range problems {
+		problems[i] = fmt.Sprintf(" - %v", problem)
```
kytrinyx: `%s` here too.

api/status_info.go
```diff
@@ -0,0 +1,60 @@
...
+%v
+
+Exercises skipped:
+%v
+`
+
+	return fmt.Sprintf(msg, s.TrackID, s.Recent, formatProblems(s.FetchedProblems), formatProblems(s.SkippedProblems))
+}
+
+func formatProblems(problems []string) string {
```
kytrinyx: We could make an alias type `type Slugs []string` and implement a `String()` on that which has this behavior. Then both `FetchedProblems` and `SkippedProblems` could be `Slugs`.

kytrinyx:
This is looking really great. I have a few more small nitpicks, then if you wouldn't mind squashing the commits we can go ahead and merge.

I don't always need PRs to be a single commit, but I do want each commit to be an atomic change, and there was a bit of churn here that doesn't seem relevant to the history.

devonestes:
@kytrinyx @Tonkpils Ok, changes made and all is working well. Let me know if there's any further feedback or if we're good to go! In the meantime, I'm going to start working on getting the API endpoint to serve up actual data and not just the mock data that I've been using to test this.

api/api.go
kytrinyx: s/language/trackID/

api/status_info.go
```diff
@@ -0,0 +1,64 @@
...
+	SkippedProblems *Slugs `json:"skipped"`
+}
+
+// Recent contains information about the user's most recently submitted exercise on a particular language track.
+type Recent struct {
+	Problem     string `json:"problem"`
+	SubmittedAt string `json:"submitted_at"`
+}
+
+// Slugs is a collection of slugs, all of which are the names of exercises
```
kytrinyx: This should end with a period.

kytrinyx:
Two tiny nitpicks; aside from that, this looks good to me. We should wait to merge until we have the API endpoint, since people can install using `go get`, and they would have an unexpected 404.

api/api.go
devonestes: @kytrinyx I don't understand what you mean with your comment above. Can you elaborate a bit, please?

Tonkpils: Substitute language with trackID

devonestes: Ahh, got it. Thanks @Tonkpils!

devonestes:
@kytrinyx Ok, the nitpicks have been addressed, and i also have the PR up for the (very rough) implementation of the API endpoint, which you can see here: https://github.com/exercism/exercism.io/pull/2603

api/api.go
```diff
@@ -47,13 +47,12 @@ func (c *Client) Fetch(args []string) ([]*Problem, error) {
...
-		language := args[0]
-		url = fmt.Sprintf("%s/v2/exercises/%s?key=%s", c.XAPIHost, language, c.APIKey)
+		trackID := args[0]
+		url = fmt.Sprintf("%s/v2/exercises/%s?key=%s", c.XAPIHost, trackID, c.APIKey)
 	case 2:
-		language := args[0]
+		trackID := args[0]
 		slug := args[1]
-		url = fmt.Sprintf("%s/v2/exercises/%s/%s", c.XAPIHost, language, slug)
-	default:
```
kytrinyx: Was losing the `default` case intentional?

api/api.go
```diff
@@ -198,8 +197,8 @@ func (c *Client) Submit(iter *Iteration) (*Submission, error) {
 }
 
 // List available problems for a language.
-func (c *Client) List(language string) ([]string, error) {
-	url := fmt.Sprintf("%s/tracks/%s", c.XAPIHost, language)
+func (c *Client) List(trackID string) ([]string, error) {
+	url := fmt.Sprintf("%s/tracks/%s", c.XAPIHost, trackID)
```
kytrinyx: I don't want to change variable names for things that are not relevant to the Status command. That mixes in unrelated changes, and it makes the git history more confusing (and the code harder to review).

api/api.go
devonestes: Totally not - I think I must have made a mistake in resolving the merge conflicts I had. All is restored and up to date now, though!

kytrinyx:
Merged in 62266effe2944740305029623122a63cb849a7f0

Thank you, @devonestes!

PR #229, opened 2015-10-31
Prefer EXERCISM_CONFIG_FILE to XDG_CONFIG_HOME

rwz:
Seems like it's currently impossible to override the config path using `EXERCISM_CONFIG_FILE` when `XDG_CONFIG_HOME` is set up.


kytrinyx:
Yeah, that wouldn't work. Thanks so much!

PR #231, opened 2015-11-02
Fix incorrect download path

kytrinyx:
When downloading a submission by uuid, we were using the language name
(e.g. C++ or Objective-C) rather than the track ID (cpp or objective-c).


Tonkpils:
:tada:

PR #233, opened 2015-11-02
Remove unsubmit command

kytrinyx:
The user's API key is used to identify the user, but this is not a secure method
of authorization. Until we use SSL and/or a better method of signing keys, we
should only allow deleting iterations through the web interface, where people are
authorized correctly.


kytrinyx:
Note: I left the command itself, adding a message about how to delete iterations now that the command doesn't work.

PR #235, opened 2015-11-02
Overhaul comments for clarity and correctness.

kytrinyx:
Some of the comments were not full sentences ending with a period.
See https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences.

Also, we were a bit inconsistent with the use of language vs track and
exercise vs problem.

### DIFF
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -18,11 +18,11 @@ const (
 )
 
 var (
-	// Home by default will contact the location of your home directory
+	// Home by default will contact the location of your home directory.
 	Home string
 
-	// XdgConfigHome will contain $XDG_CONFIG_HOME if it exists
-	XdgConfigHome   string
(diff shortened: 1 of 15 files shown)


paths/paths.go
```diff
@@ -18,10 +18,10 @@ const (
 )
 
 var (
-	// Home by default will contact the location of your home directory
+	// Home by default will contact the location of your home directory.
 	Home string
 
-	// XdgConfigHome will contain $XDG_CONFIG_HOME if it exists
+	// XdgConfigHome will contain $XDG_CONFIG_HOME if it exists.
```
Tonkpils: Shouldn't `XDG` be capitalized both in the comment and the constant?

kytrinyx: Yeah, it sure should. Good catch!

kytrinyx:
Fixed.

Tonkpils:
:+1: Looks great!

PR #241, opened 2015-11-03
Clarify domain concepts in README

kytrinyx:
This is a start. I'm still working on cleaning up the language/track id confusion throughout the codebase, and we may also find other things that are confusing.

Related to https://github.com/exercism/exercism.io/issues/2532

I've been normalizing to "problem", but we may want to switch to "exercise" which has fewer negative connotations when used out of context.


Tonkpils:
This helps a lot, I was a bit often confused on the terminology. This is great.

PR #247, opened 2015-11-06
Rename function for clarity

kytrinyx:
The submission method on the api client doesn't return a submission, but rather a url.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -109,8 +109,8 @@ func (c *Client) Submissions() (map[string][]SubmissionInfo, error) {
 	return payload, nil
 }
 
-// Submission gets the latest submitted exercise for the given language track id and problem slug.
-func (c *Client) Submission(trackID, slug string) (*Submission, error) {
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -109,8 +109,8 @@ func (c *Client) Submissions() (map[string][]SubmissionInfo, error) {
 	return payload, nil
 }
 
-// Submission gets the latest submitted exercise for the given language track id and problem slug.
-func (c *Client) Submission(trackID, slug string) (*Submission, error) {
+// SubmissionURL gets the url of the latest iteration on the given language track id and problem slug.
+func (c *Client) SubmissionURL(trackID, slug string) (*Submission, error) {
```
Tonkpils: Are we sure this is only returning the URL? We're actually returning a `Submission` struct so it contains all those fields.

kytrinyx: Pretty much:

https://github.com/exercism/exercism.io/blob/master/lib/api/routes/submissions.rb#L18

PR #248, opened 2015-11-09
add package for logging output

lcowell:


lcowell:
This will provide us with 2 levels of output, normal output and debugging output. With the Verbose flag enabled, debugging output will also be shown to the user. Once we're happy with how this is looking, I'll integrate it with the rest of the app.

kytrinyx:
I think I like where this is going. I'm hesitating over one thing: do we really want to use the log package to tell the user things? That seems like not logging to me. Now that I write it out, it seems like debug info isn't quite logging either, so maybe this is just a question of semantics.

lcowell:
Semantics are important. The better the naming, the less we have to think when we're reading code. What do you think about changing the package name to `out` or `output`?

```
out.Println("hi")
out.Debugf("%s\n", "ok")
```

We could also keep the user related output going through the `fmt` package and make a `debug` package. So we'd have `debug.Println` and that would only print if verbose (or perhaps should be rename to debug) flag is set.

I'm kind of leaning toward the `debug` package as it doesn't interfere with normal user output.

kytrinyx:
Yeah, either of these feel like good approaches to me. I'm with you on leaning a bit towards using `fmt.Print` directly for normal user output, and then having a `debug` package for the verbose output.

I'm thinking the usual stuff would go to STDOUT and the debug stuff would go to STDERR.

lcowell:
This should now reflect the changes we'd talked about. I've also wired up the verbose flag to activate the debug output.

Tonkpils:
:+1: Looks good to me.

lcowell:
merged e0718312d28850d4dfb10f53b6f8daa3df34e072

PR #253, opened 2015-11-21
Add request and responses to debugging output

lcowell:
Helps with debugging situations like this:
https://github.com/exercism/xgo/issues/202

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -8,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/debug"
 )
 
 const (
@@ -54,10 +56,13 @@ func (c *Client) NewRequest(method, url string, body io.Reader) (*http.Request, 
 
 // Do performs an http.Request and optionally parses the response body into the given interface.
 func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
+	debug.Println("Request", req.Method, req.URL)
+
 	res, err := c.client.Do(req)
 	if err != nil {
 		return nil, err
 	}
+	debug.Printf("Response StatusCode=%d\n", res.StatusCode)
 
 	switch res.StatusCode {
 	case http.StatusNoContent:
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -71,7 +74,12 @@ func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
...
 		if v != nil {
 			defer res.Body.Close()
-			if err := json.NewDecoder(res.Body).Decode(v); err != nil {
+
+			var bodyCopy bytes.Buffer
+			body := io.TeeReader(res.Body, &bodyCopy)
+
+			if err := json.NewDecoder(body).Decode(v); err != nil {
+				debug.Println("res.Body:", bodyCopy.String())
 				return nil, fmt.Errorf("error parsing API response - %s", err)
```
Tonkpils: We should change this message to not return the error from the JSON decoder and add that error to the debug output. This message could be much friendlier and direct the user to either re-run the command with debug output or to file a ticket. If we get a request-id on the response back, we should display that as part of the friendly message. Thoughts?

kytrinyx: Agreed.

I'm going to look at the API part right now.

kytrinyx: OK, so it looks like this is when we do `api.Submissions()` after the fetch, the API is returning an error as JSON:

```
# http://exercism.io/api/v1/exercises?key=
{
  error: "Sorry—we can't figure out who you are. Double-check your API key in your exercism.io account page."
}
```

I'll change the API to return a zero-value for this call.

kytrinyx: Ugh. This is kind of awful: the usual response from the API is in this format:

```json
{
    "go": [
        {
            "slug": "hello-world",
            "state": "archived"
        }
    ],
    "ruby": [
        {
            "slug": "hello-world",
            "state": "archived"
        }
    ]
}
```

kytrinyx: OK, deploying API with reasonable empty state.

lcowell: Fetching without an api key now just fetches the exercise. :+1: 

Getting an error response from the server, now looks like this:
```
> ./cli fetch go leapa
verbose logging enabled
NewRequest: GET http://x.exercism.io/v2/exercises/go/leapa
StatusCode: 404
res.Body: {"error":"We don't have problem 'leapa' in 'go'"}
2015/11/21 11:16:59 unable to fetch problems (HTTP: 404) - We don't have problem 'leapa' in 'go'
```
We are displaying the error provided by the json response. Were any other changes required?

I'm wondering if we should always show the body response in verbose mode regardless of whether we have an error or not? It does add a lot of noise, but could be useful as we make changes to the erecism.io api. It could also be useful if, for example, someone has a proxy server that is mangling the response.

kytrinyx: How was the verbose logging activated in the above command? It doesn't look like a flag was passed to `./cli`.

I think you're right that we should always show the body in verbose mode, that would be immensely helpful. Maybe with a newline after it so that the "usual" output is separated a bit.

lcowell: I over-condensed the above. This is the real command:

    go build ./exercism/main.go && ./main --verbose -c config.json fetch go leapa

Cool. I'll space out the responses.

kytrinyx: Ah, good! I was worried I was missing something. Thanks.

PR #254, opened 2015-11-22
Ditch the demo command

kytrinyx:
The demo was great when we had two or three languages, but now that we have almost
30 it makes no sense to download the first problem in each track.

The website no longer shows the demo command, but rather encourages people
to download the first problem in a given track explicitly.

We will also be changing the fetch command to not fetch things in unstarted
tracks, which is even more reason to not have someone implicitly start all
the tracks.


lcowell:
Looks good to me. Should we indicate to users that the demo command has been removed and suggest the preferred method of trying exercism out? I'm not sure if demo was used enough to warrant the warning.

kytrinyx:
I'd rather just make it go silently away. My thinking is: if you used it, you probably only used it once, and then never thought about it again.

lcowell:
Good point. I agree.

PR #256, opened 2015-11-22
clean up build scripts

lcowell:
These scripts are no longer used in the current build-all script and should be removed.


kytrinyx:
Yepp, good thinking!

Issue #187, opened 2015-04-25
We lie about fetching the next exercise...


kytrinyx:
In `cmd/submit.go` we say _To get the next exercise, run "exercism fetch" again_ whenever you submit. This is not actually true. We only unlock the next exercise if this is the first iteration of a submission.

I've modified the response from the server to include an `iteration` key, which is an `int`. When `iteration` is 1, then we can say that a new problem is available.

... And of course even that might be a lie, because if you're at the end of a track, we don't actually tell you that.

https://github.com/exercism/exercism.io/issues/2315

Tonkpils:
I'm not 100% familiar with the design of the iterations on the server how feasible would it be to actually have a count of total exercises for a track and a count of submitted exercises for a user? That would allow us to tackle two issues. I can probably spend some time to take a look at it but I haven't been in the codebase in a while. Thought maybe you'd have a better idea.

kytrinyx:
We can know how many issues there are in a track very easily--the Problem API already has that information and can send the number along when fetching.

The number of things they've completed is another matter--that's the exercism.io repo. But... what if every time they fetch we stick the name of the track/problem in a redis set? Or we could even do a redis hash of problem/track => fetch count, which might be interesting.

It's super cheap to do, it would keep it on the same server.

On the other hand, we don't have their user id. Hm. At that point we'd have to use their key to get their user ID, which sucks. So maybe it's easier to just make a call to the exercism.io api to get the number. We can make the two calls in parallel.

kytrinyx:
BTW - I haven't deployed yet. I'm super jet-lagged and don't want to accidentally break something when I don't have the brain power to fix it. I'll deploy the `iterations` thing tomorrow.

Tonkpils:
> On the other hand, we don't have their user id. Hm. At that point we'd have to use their key to get their user ID, which sucks. So maybe it's easier to just make a call to the exercism.io api to get the number. We can make the two calls in parallel.

What if the we stuck the submission count in a redis hash with the API key as the hash key that way we could just match the API key with the submission count on fetch. 

I don't want to clutter this issue with what seems to be a separate discussion. Want to move it to a more appropriate place?

kytrinyx:
Yeah, let's move the discussion to a new issue.

kytrinyx:
To clarify this issue--it's just a matter of changing the messaging:

If the iteration number we get back when having submitted is 1, then we're unlocking the next exercise.

wadtech:
[...]

PR #257, opened 2015-12-01
Only display "to get the next exercise" message if the server responds with the correct iteration value.

wadtech:
Hi all,

I've had a go at #187, hopefully I've understood the issue properly.

I checked exercism.io for the added iteration key on the submit endpoint and I tested it against my own account using my local build of the cli.

Any and all feedback welcome, I'm trying to get into Go more.

The main thing I'm not so sure about is the `+=` string concatenation, I'm not too sure that's the best way to go about it in idomatic Go.

Thanks,
Pete

### DIFF
--- a/api/submission.go
+++ b/api/submission.go
@@ -10,4 +10,5 @@ type Submission struct {
 	Username      string            `json:"username"`
 	ProblemFiles  map[string]string `json:"problem_files"`
 	SolutionFiles map[string]string `json:"solution_files"`
+	Iteration     int               `json:"iteration"`
 }
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -92,8 +92,13 @@ func Submit(ctx *cli.Context) {
 	msg := `
 Submitted %s in %s.
 Your submission can be found online at %s
+`
 
+	if submission.Iteration == 1 {
+		msg += `
 To get the next exercise, run "exercism fetch" again.
 `
+	}
+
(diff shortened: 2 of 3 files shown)


Tonkpils:
This looks great. Ideally, we would have tests that would check to make sure that the text is only displayed on iteration 1. However, I'm still thinking of how to approach testing the commands themselves and their output. 

> The main thing I'm not so sure about is the += string concatenation, I'm not too sure that's the best way to go about it in idomatic Go.

For this case, `+=` is actually just fine since we're only doing it once and the strings themselves are not lengthy. In cases where we are doing many string concatenations and the strings can become very lengthy or we're looking for performance rather than convenience, then `bytes.Buffer` is the way to go.

wadtech:
Thanks! I also lamented the lack of tests but didn't know how to approach it.

Tonkpils:
> Thanks! I also lamented the lack of tests but didn't know how to approach it.

No problem! Thanks for your contribution! If you have any ideas on how to approach testing the `cli` commands I'm all open to suggestions. I may reach out to the owner of the package to see his thoughts on it.

kytrinyx:
>  I'm still thinking of how to approach testing the commands themselves and their output.

I've been thinking about that as well. We could definitely start creating tests that

1. create a fake context value and pass it to a command,
1. uses the testhttp package to fake out the calls to the API, and
1. uses example style tests to assert on the output

Issue #226, opened 2015-10-26
adjust behavior of exercism fetch to avoid noise on the file system


kytrinyx:
Suggestion from @frenata in #2573

------

What if exercism fetch only pulled in exercises for language folders that already exist? This would require that to get started with a new language you'd have to first manually specify it exercism fetch go , but not afterwards.

That way you only get exactly the language exercises that you want. A message to that effect when issuing the command with an empty exercism folder, and I'd guess that the workflow would be a little smoother, avoiding the issue of forgetting to specify a language and ending up with a bunch of subfolders for languages the user isn't interested in.

frenata:
I'm poking around in the codebase to see if I can solve/implement this myself, rather than just ask someone else to. Heh. In the meantime, I ran into an entirely different issue when running the tests...

kytrinyx:
Fun! Ask if you have any questions, we're always happy to help point people in the right direction. And sometimes if the stuff that comes up is too embarrassing, it's a good opportunity to refactor.

devonestes:
:+1: on this feature - I'd also be happy to dive in and see if there are ways I can help, too!

Tonkpils:
I like this idea. This should be fairly simple to implement and I'd be more than happy to pair or help out with whoever wants to tackle this.

lcowell:
I'd be up for pairing. I'd have a couple hours saturday or sunday morning/early afternoon.

kytrinyx:
Any updates on this one? If things have gotten too hectic, feel free to "unclaim" it (just add a comment saying you're too busy or whatever) and then someone else might grab it.

devonestes:
If nobody minds then I think I'll start work on this week. Don't want to steal anyone's thunder, but I'd really love to see this feature added!

lcowell:
:+1: Go for it!

frenata:
I've been busy traveling and visiting family, or I'd have taken a serious shot at this feature--I've not had really much time for exercism at all! I'll try to keep an eye on your work though, devonestes, and if I can carve out the time may do the work myself as well, if only for my own edification.

devonestes:
Ok, after starting some exploration of this, I think there are three ways to solve the problem.

1) Get a list of all the folders that exist in the user's exercises directory and make API calls to the existing endpoint for each of those folders. This is basically the equivalent of looping over the folders on the user's computer and executing `exercism fetch TRACK` for each of those folders.

[...]

PR #258, opened 2015-12-01
Changes `exercism fetch` behavior to remove filesystem noise

devonestes:
This commit changes the behavior of `exercism fetch` when called without any arguments. Instead of fetching and writing exercises for all tracks, it will now fetch and update only tracks for which the user has folders in their local environment. If a user calls `exercism fetch` without having any language tracks in their exercises folder, a prompt is shown to the user to tell them how to list all available tracks and how to fetch their first track for a language of their choice.

### DIFF
--- a/user/homework.go
+++ b/user/homework.go
@@ -2,6 +2,7 @@ package user
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/exercism/cli/api"
@@ -54,6 +55,27 @@ func (hw *Homework) Save() error {
 			return err
 		}
 	}
+	return nil
+}
+
+// RejectMissingTracks removes any items that are part of tracks the user
+// doesn't currently have a folder for on their local machine. This
+// only happens when a user calls `exercism fetch` without any arguments.
+func (hw *Homework) RejectMissingTracks(dirMap map[string]bool) error {
+	items := []*Item{}
+	for _, item := range hw.Items {
+		dir := filepath.Join(item.dir, item.TrackID)
+		if dirMap[dir] {
+			items = append(items, item)
+		}
+	}
+	if len(items) == 0 {
+		return fmt.Errorf(`
+You have yet to start a language track!
+View all available language tracks with "exercism tracks"
+Fetch exercises for your first track with "exercism fetch TRACK_ID"`)
+	}
+	hw.Items = items
 	return nil
 }
 
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"log"
+	"path/filepath"
 
 	"github.com/codegangsta/cli"
 	"github.com/exercism/cli/api"
@@ -31,7 +32,23 @@ func Fetch(ctx *cli.Context) {
 		log.Fatal(err)
 	}
 
+	dirs, err := filepath.Glob(filepath.Join(c.Dir, "*"))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	dirMap := make(map[string]bool)
+	for _, dir := range dirs {
+		dirMap[dir] = true
+	}
 	hw := user.NewHomework(problems, c)
+
+	if len(ctx.Args()) == 0 {
+		if err := hw.RejectMissingTracks(dirMap); err != nil {
+			log.Fatal(err)
+		}
+	}
+
 	if err := hw.Save(); err != nil {
 		log.Fatal(err)
 	}
--- a/user/homework_test.go
+++ b/user/homework_test.go
@@ -1,0 +1,47 @@
+package user
+
+import (
+	"testing"
+
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
+)
+
+func TestRejectMissingtracks(t *testing.T) {
+	problem1 := &api.Problem{
+		TrackID:  "go",
+		Slug:     "clock",
+		Language: "Go",
+		Name:     "Clock",
+	}
+	problem2 := &api.Problem{
+		TrackID:  "ruby",
+		Slug:     "clock",
+		Language: "Ruby",
+		Name:     "Clock",
+	}
+	dirMap := map[string]bool{
+		"/tmp/go":      true,
+		"/tmp/haskell": true,
+	}
+	emptyDirMap := make(map[string]bool)
+
+	hw := NewHomework([]*api.Problem{problem1, problem2}, &config.Config{Dir: "/tmp"})
+	err := hw.RejectMissingTracks(dirMap)
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if len(hw.Items) == 2 {
+		t.Error("Should have rejected the Ruby problem but did not reject any problems!")
+	}
+
+	if len(hw.Items) == 1 && hw.Items[0].TrackID == "ruby" {
+		t.Error("Should have rejected the Ruby problem and rejected the Go problem instead!")
+	}
+
+	if err := hw.RejectMissingTracks(emptyDirMap); err == nil {
+		t.Error("Should have returned error because user hasn't started any tracks but didn't!")
+	}
+}


devonestes:
This is my rough first attempt at a fix for #226. Any feedback is much appreciated!

user/homework.go
```diff
@@ -57,6 +59,31 @@ func (hw *Homework) Save() error {
...
+			}
+		}
+		hw.Items = existingHW
+		if len(hw.Items) == 0 {
+			fmt.Println(`
+You have yet to start a language track!
+View all available language tracks with "exercism tracks"
+Fetch exercises for your first track with "exercism fetch LANGUAGE"
+`)
+			os.Exit(0)
```
kytrinyx: A couple things (philosophically):

1. I think we should only use `os.Exit` for a value other than 0 (in the case of an error), and
2. I don't think we should ever have `os.Exit` in packages other than `cmd`.

user/homework.go
```diff
@@ -57,6 +59,31 @@ func (hw *Homework) Save() error {
 	return nil
 }
 
+// SaveAfterFetch filters out problems for tracks the user has yet to start
+// before saving them all to disk.
+func (hw *Homework) SaveAfterFetch(numArgs int) error {
+	if numArgs == 0 {
```
kytrinyx: My first thought when seeing this was _can we minimize indentation here_?

How about introducing a `hw.RejectItems()` function or something like that? Maybe it could take the arguments, and exit immediately if there's nothing to reject:

```
func (hw *Homework) RejectSomething(...) {
    if nothingToDoHere {
        return
    }
    // do the work
}

func (hw *Homework) SaveAfterFetch(...) error {
     hw.RejectStuff(...)
     if ln(hw.Items) == 0 {
        return err // that you made up
     }
     return hw.Save()
}
```

user/homework.go
```diff
@@ -57,6 +59,31 @@ func (hw *Homework) Save() error {
...
+		var existingHW []*Item
+		for _, item := range hw.Items {
```
kytrinyx: Rather than `existingHW` let's just call them `items`.

kytrinyx:
I like where this is going. A few suggestions, but nothing major, I think.

devonestes:
@kytrinyx thanks for the feedback! All your comments have been addressed, and I made a couple other small changes to be more consistent with the existing style in `homework.go`.

user/homework.go
```diff
@@ -57,6 +59,36 @@ func (hw *Homework) Save() error {
...
+// SaveAfterFetch filters out problems for tracks the user has yet to start
+// before saving them all to disk.
+func (hw *Homework) SaveAfterFetch(numArgs int) error {
+	hw.RejectUnstartedTracks(numArgs)
+	if len(hw.Items) == 0 {
+		fmt.Println(`
+You have yet to start a language track!
+View all available language tracks with "exercism tracks"
+Fetch exercises for your first track with "exercism fetch LANGUAGE"`)
+	}
```
kytrinyx: Instead of `LANGUAGE` this should be `TRACK_ID` (it's super confusing, I'm so sorry!).

Should we return that as an error instead of printing it here?

user/homework.go
```diff
@@ -57,6 +59,36 @@ func (hw *Homework) Save() error {
...
+	return hw.Save()
+}
+
+// RejectUnstartedTracks removes any items that are part of tracks the user
+// has yet to start and doesn't have a folder for on their local machine. This
+// only happens when a user calls `exercism fetch` without any arguments.
+func (hw *Homework) RejectUnstartedTracks(numArgs int) {
```
kytrinyx: OK, now I'm nitpicking _really_ hard! They're not necessarily unstarted in this scenario, they're just... I don't know... inactive? If I started a track but then deleted the directory off my machine, it would not write new exercises to it.

I'm also thinking the `numArgs` is going to confuse me down the line when I've forgotten this discussion.

I don't know the right solution, so I'm just going to think out loud.

The arguments are arguments to the command when the user calls `exercism fetch`. We're using the argument count as a flag: either do something or don't do it (right? Or am I forgetting something?).

If it's a flag, then it's a flag for... filtering?

A boolean makes a good flag, but I'm not sure what to name it. `isSomething`? But I'm not sure what makes a good _something_ in this context.

Any suggestions?

kytrinyx:
This is looking really clean. I'm worried about that one name, but that's it.

user/homework.go
devonestes: I didn't really think that this warranted an error, since when an error is returned from this that will bubble up to a `log.Fatal` call and exit in that fashion. I thought that might be a little jarring for the user, so I decided to go this route. However, I'm happy to change this to an error if you think that's more intention revealing!

kytrinyx: Hm. Yeah. I see what you mean.

I think it _is_ an error, though. They asked to fetch something, and we're telling them: something's not right.

devonestes: That's a good point. If the user asked to do something and it couldn't be done, then that's totally an error. Ok, switching over to an error now!

user/homework.go
kytrinyx: How about `filterMode bool` for the argument? It doesn't feel quite right, but it's a bit closer.

devonestes: Good call on using a boolean instead. How about `isCalledWithoutArgs` or `isFetchedWithoutArgs`? Both are kind of long, but I can't think of anything shorter to describe the behavior we're flagging here.

kytrinyx: The thing about _is called without args_ is that it doesn't say what it _means_ to be fetched without args.

devonestes:
@kytrinyx I think we're nearly there. I still don't love the name of `isFetchedWithoutArgs`, but until someone can think of something better that's what there. Maybe something will come to me after a good night's sleep!

kytrinyx:
I like `filterMode` better than `isFetchedWithoutArgs`, to be honest. filterMode is about what we're deciding (to filter or not) whereas isFetchedWithoutArgs is one step away. The reader has to make the connection between calling without arguments and this meaning that we're going to filter.

devonestes:
I sort of like the `isSomething` name. `filterMode` is a great name when it's next to an `if`, but outside of that context it seems a little odd to me. How about `shouldFilter`?

kytrinyx:
Yeah I like the `is...` thing better, too. `shouldFilter` is better, and it reads better than `filterMode` in the context as well. Let's go with that. If someone thinks of something better, holler!

devonestes:
@kytrinyx Great, `shouldFilter` it is! Let me know if there's anything else that jumps out at you that you think should be changed and I'll get right on it.

lcowell:
Looks great. Does`RejectMissingTracks` need to be exported? It makes sense to have `SaveAfterFetch` exported because it's the replacement entry point for `Save`.

I think it would be good to see a test to verify this new behaviour. It could to be a little difficult because you're calling `os.Stat` deeper in your code. I'm wondering if we should pass a list of known language to `SaveAfterFetch`, so that we don't have to set up a whole directory in order to test this.

devonestes:
@lcowell Good call on not exporting `RejectMissingTracks`!

I agree that a couple of tests would be a good thing, but I think the list of known languages would be a real hassle to keep up to date. Dealing with creating and deleting directories for the tests won't be that bad, so I'll just go that route for now.

lcowell:
To get the list of user languages, isn't it just a matter of getting a listing of all the directories in your exercism language? Something like:

`filepath.Glob(filepath.Join(exercismDir, "*"))`

devonestes:
@lcowell Sorry, I thought when you said 'Pass a list of known languages' you meant that we'd keep a list of the available languages on exercism.io in a constant or something and pass that list. My misunderstanding!

Yes, we could totally go the way you describe to solve this problem (that's the way I originally thought of doing it), but it was faster and easier for me to write the tests with the code in the shape it is in now rather than refactoring. If there's a particular problem or reason why it would be better to refactor then I'd love to hear it and would be glad to refactor!

lcowell:
I think the motivation is to isolate the parts of the system that need to talk to the OS. This improves testability and reduces brittleness in the system. In a system where everything can talk to everything it becomes more difficult to reason about what's happening or what went wrong.

kytrinyx:
@lcowell excellent call on passing the list of languages to `SaveAfterFetch()` so we can do an isolated test on that. @devonestes Do you see how that would work?

devonestes:
@kytrinyx @lcowell Your reasoning makes total sense - the separation of concerns benefit is definitely worth a refactor. I'll get on in two shakes of a lamb's tail :sheep:

devonestes:
Ok, let me know what you think about the way I've refactored this, and thanks again for the idea!

lcowell:
This looks great. Is there some reason we can't handle `shouldFilter` in `SaveAfterFetch`? If we do that `rejectMissingTracks` will do exactly what it says it does *and* we avoid passing an argument where we don't need to. I think passing `shouldFilter` in made more sense when `rejectMissingTracks` was exported.

devonestes:
@lcowell No reason, really. @kytrinyx had recommended that structure in a previous comment (here: https://github.com/exercism/cli/pull/258#discussion_r46345180), so I went with that since it didn't really bother me either way.

kytrinyx:
I like @lcowell's suggestion better :) (sorry!)

devonestes:
@kytrinyx @lcowell No worries - that's why more opinions are better than one!

devonestes:
Ok, updated and ready for review!

lcowell:
Looks great. Nice work @devonestes !

kytrinyx:
Agreed, now that you mention it :) Gawd I love code review.

Now that you mention it, should we just have one `Save()` function and one `Reject` or `Filter` function that we call from the `cmd` package? (I'm so sorry to keep thinking out loud about this, it really does look good!).

lcowell:
I like code review too. I learn so much, whether I'm the one giving feedback or the one writing code.

In the context of homework, which is something that is submitted, I think that `Filter` makes more sense than `Reject`.

I see a couple of approaches to implementing this:
```go
  hw.Filter(dirMap)
  err := hw.Save()
```

or if we have `Filter` return a copy of the homework:

```go
  err := hw.Filter(dirMap).Save()
```

kytrinyx:
Yepp, I agree.

devonestes:
Well right now in `homework` we have the following code:

```go
// HWFilter is used to categorize homework items.
type HWFilter int

// SummaryOption allows selective display of summary items.
type SummaryOption HWFilter

const (
	// HWAll represents all items in the collection.
	HWAll = iota
	// HWUpdated represents problems where files have been added.
	HWUpdated
	// HWNew represents newly fetched problems.
	HWNew
	// HWNotSubmitted represents problems that have not yet been submitted for review.
	HWNotSubmitted
)

...

// ItemsMatching returns a subset of the set of problems.
func (hw *Homework) ItemsMatching(filter HWFilter) []*Item {
	items := []*Item{}
	for _, item := range hw.Items {
		if item.Matches(filter) {
			items = append(items, item)
		}
	}
	return items
}
```

Since there's already a bit of framework for filtering `hw.Items`, should I try and integrate this filtering in with that? I avoided that at first since it seemed like it was more hassle than it was worth, but if we think that might be helpful I'll take a crack at it.

lcowell:
Can you briefly describe how would you implement this? The only ways I can think of would add complexity instead of reduce. But hey, I'm only one coffee deep, so I might be missing something obvious.

devonestes:
@lcowell I totally agree that it would add complexity, but if the goal is to have one `Save()` function and one `Filter()` function, then I don't think that this code can exist alongside the single `Filter()` function without causing confusion. My concern is that by generalizing the names and not incorporating all of the behavior that might fall under those names, that maintainers later on will get confused.

kytrinyx:
I don't want to add complexity.

Here's what I was thinking:

```
// in cmd/fetch.go
if len(ctx.Args()) == 0 {
  if err := hw.FilterMissingTracks(dirMap); err != nil {
    log.Fatal(err)
  }
}
if err := hw.Save(); err != nil {
  log.Fatal(err)
}
```

devonestes:
Ahh, I understand now. Crystal clear - I'll get on it now!

kytrinyx:
I really like how this ended up. Thanks @devonestes and @lcowell for working through all the details!

Issue #252, opened 2015-11-16
Clarify domain concepts throughout the codebase


kytrinyx:
We've spent some time coming up with terminology that works for the project. Here's what we've landed on:
- **Track** -  track is in a language (e.g. C++) and has an ID (e.g. `cpp`)
- **Problem** - a generic description of a problem to be solved, can be implemented in any language
- **Exercise** - a language-specific implementation of a problem. It has a test suite in that language.
- **Solution** - an author's worked answer to an exercise. It may have 1 or more iterations.
- **Iteration** - a single version of an author's solution to an exercise.

At the moment we've got some old names in the codebase. I think `Problem` is used in the old sense (it should now be `Exercise`) and `Submission` should become `Iteration`. There might be other confusing things in there as well.

devonestes:
I might be able to take this one on. Would it be better to submit commits one at a time rather than all at once? For example, one commit would update all instances in the codebase where something should be called a Track, the next commit for Problem, etc.

I'm thinking it might be easier to review the code if each commit only focused on one change. I could of course squash them down before merging.

kytrinyx:
Yes -- please take little bits and pieces at a time. I might even suggest adding a comment here to say which piece you're tackling, then if someone else wants to grab another piece meanwhile they could.

It's much, much easier to review if each commit is one type of change, and I would not want them to be squashed. I might even suggest having different pull requests (much easier to review that way).

It wouldn't even have to be all about the same name, but perhaps all about one name in one area of the code, or all about one struct throughout the codebase.

I've got a lot of experience with refactoring and overwhelmingly, smaller is better unless it's a very straight-forward change that just happens to be in 800 places.

devonestes:
Sounds good! The first thing that's jumped out at me is the use of `Submission` instead of `Iteration` in `api.go`, so I'll start there with the updating.

kytrinyx:
+1 sounds great

bordeltabernacle:
Is this something that still needs looking at?

kytrinyx:
I've actually been thinking that it would be nice to review the entire CLI user interface, normalize it, tweak the API endpoints to also be normalized, and then use spf13/cobra as the framework. If I have time to think through it more thoroughly I'll open an issue to start a discussion about what that might look like.

bordeltabernacle:
[...]

PR #260, opened 2015-12-02
Updating terminology for Iteration in api.go & elsewhere

devonestes:
This is the first PR towards updating some terminology in the codebase as referenced in #252. I've changed a few names here and there and hopefully made things a little more clear at the expense of some longer names.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -25,9 +25,9 @@ type PayloadProblems struct {
 	PayloadError
 }
 
-// PayloadSubmission represents metadata about a successful submission.
-type PayloadSubmission struct {
-	*Submission
+// PayloadSubmittedIteration represents metadata about a submission of an iteration.
+type PayloadSubmittedIteration struct {
+	*SubmittedIteration
 	PayloadError
 }
 
@@ -109,15 +109,15 @@ func (c *Client) Submissions() (map[string][]SubmissionInfo, error) {
 	return payload, nil
 }
 
-// SubmissionURL gets the url of the latest iteration on the given language track id and problem slug.
-func (c *Client) SubmissionURL(trackID, slug string) (*Submission, error) {
+// IterationURL gets the url of the latest iteration on the given language track id and problem slug.
+func (c *Client) IterationURL(trackID, slug string) (*SubmittedIteration, error) {
 	url := fmt.Sprintf("%s/api/v1/submissions/%s/%s?key=%s", c.APIHost, trackID, slug, c.APIKey)
 	req, err := c.NewRequest("GET", url, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	var payload Submission
+	var payload SubmittedIteration
 	if _, err := c.Do(req, &payload); err != nil {
 		return nil, err
 	}
@@ -125,16 +125,16 @@ func (c *Client) SubmissionURL(trackID, slug string) (*Submission, error) {
 	return &payload, nil
 }
 
-// Download fetches a solution by submission key and writes it to disk.
-func (c *Client) Download(submissionID string) (*Submission, error) {
-	url := fmt.Sprintf("%s/api/v1/submissions/%s", c.APIHost, submissionID)
+// Download fetches an iteration by iteration key and writes it to disk.
+func (c *Client) Download(iterationID string) (*SubmittedIteration, error) {
+	url := fmt.Sprintf("%s/api/v1/submissions/%s", c.APIHost, iterationID)
 
 	req, err := c.NewRequest("GET", url, nil)
 	if err != nil {
 		return nil, err
 	}
 
-	payload := &PayloadSubmission{}
+	payload := &PayloadSubmittedIteration{}
 	res, err := c.Do(req, payload)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -25,9 +25,9 @@ type PayloadProblems struct {
 	PayloadError
 }
 
-// PayloadSubmission represents metadata about a successful submission.
-type PayloadSubmission struct {
-	*Submission
+// PayloadSubmittedIteration represents metadata about a submission of an iteration.
+type PayloadSubmittedIteration struct {
+	*SubmittedIteration
```
Tonkpils: I don't think there's a time when the CLI will ever handle an iteration that is not Submitted. This should probably be simply `Iteration` along with removing `Submitted` from the other structs

devonestes: I thought the same thing, but there's already an `Iteration` type defined in `iteration.go`. The two types are really similar and could probably be combined with no real consequences, but since that was a much more significant refactor than this, I felt that it would be better to keep two different types rather than trying to merge them now.

After writing all that, though, I'm starting to think that merging those two types might be a good idea. What do you think?

Tonkpils: Taking a bit closer look, `Iteration` is the payload sent when submitting a new iteration or a local iteration. `Submission` is the already submitted solution which just seems to me like.

Reading the #252, seems like actually `Submission`, should actually be `Solution` since a solution can have more than one iteration and `Submission` actually contains the number of iterations. 

Thoughts? @kytrinyx

kytrinyx:
I think they're both iterations--do we ever handle more than one iteration at a time for a solution on our end? I don't think we do.

We're using one on restore and one on submit. It sounds like they should really be the same struct. Are there values that we have in one context but not another? It looks like the ProblemFiles, but that seems fine.

We should probably call `Submission.Iteration` `Iteration.Number` for clarity.

I'm so glad you're tackling this. The codebase is going to be so much easier to reason about once we get this sorted out!

kytrinyx:
Just checking in. Did this get too weird and ambiguous?

devonestes:
@kytrinyx No, I wanted to focus on the other PR I had open so I was only working on one thing at a time, and then things got a little hectic in the pre-holiday rush at work, but now that my other PR is closed I'll be back on this one today!

kytrinyx:
One thing at a time sounds like a great idea. I'm trying to enforce that in my own life. Let me know if you need anything from me.

devonestes:
@kytrinyx @Tonkpils So I'm torn on something with this. Since both `Iteration` and `SubmittedIteration` are used to parse JSON from the API, we're going to have to update the API endpoints to merge those two structs into one, right? If so, do you think that's a good way to go with this, or is it a bit of feature creep to expand this task out to include that refactoring?

kytrinyx:
What are the differences in the payloads?

devonestes:
Much of the data is similar, but the keys in the JSON returned from the API are all different for the same data. We'd have to do a little normalization of the data returned from the API and also make sure the names of the keys were all the same in order for those two structs to be merged into one. It might be worth it, or maybe it won't...

kytrinyx:
It's worth it. If you could list the differences, I'd be happy to normalize the API (write it to both keys so that the old and new clients can both handle the payloads).

devonestes:
Ok, here are the two type definitions that we'll be merging together, including the JSON keys that they're mapped to.

```go
type Iteration struct {
	Key      string            `json:"key"`
	Code     string            `json:"code"`
	Dir      string            `json:"dir"`
	TrackID  string            `json:"language"`
	Problem  string            `json:"problem"`
	Solution map[string]string `json:"solution"`
}

type SubmittedIteration struct {
	URL           string            `json:"url"`
	TrackID       string            `json:"track_id"`
	Language      string            `json:"language"`
	Slug          string            `json:"slug"`
	Name          string            `json:"name"`
	Username      string            `json:"username"`
	ProblemFiles  map[string]string `json:"problem_files"`
	SolutionFiles map[string]string `json:"solution_files"`
	Iteration     int               `json:"iteration"`
}
```
Right now I'm thinking the Iteration type will look like this:

```go
type Iteration struct {
	IterationURL    string            `json:"iteration_url"`
	TrackID       	string            `json:"track_id"`
	Slug          	string            `json:"slug"`
	Username      	string            `json:"username"`
	UserAPIKey      string
	Code     				string
	Dir      				string            `json:"dir"`
	ProblemFiles  	map[string]string `json:"problem_files"`
	SolutionFiles 	map[string]string `json:"solution_files"`
	IterationNumber	int               `json:"iteration_number"`
}
```

I'm happy to make the changes to the API as well if you like. I was just unsure if we wanted this work to spill over into changing the API as well.

Tonkpils:
Is the `user_api_key` supposed to be in the payload?

devonestes:
@Tonkpils At the moment that's how it's done... I think it's mostly so that it can be used to send iterations when they're complete since that is needed in the API endpoint for the POST request. I don't think it's a great way to handle it, though...

kytrinyx:
Let's call it `URL` not `IterationURL` and `Number` not `IterationNumber` to avoid the stutter (the type is called iteration already).

We shouldn't need both `Code` and `SolutionFiles`. We also shouldn't need `ProblemFiles`... it should probably just all be `Files`.

Agreed, we shouldn't have the api key as part of the iteration. Isn't it currently sent as a query param? If not, should we change that?

devonestes:
So I was messing around with this last night a bit, and I'm starting to feel a little worried that I might break something in doing what is now a kind of significant refactor without a ton of test coverage. Would it be ok with you all if I put this on the back burner and worked a bit to try and get some more test coverage? I would just feel a lot more confident in my refactoring if I knew I wasn't breaking anything significant while I was working.

kytrinyx:
Absolutely! Getting some good test coverage would be fantastic.

devonestes:
So I've been running into some blockers working on this - I just need to read up a bit more on some stuff (regarding testing, mainly) and come back to it, I think. If you don't mind, I'm going to close this PR, do some reading, and come back to this with a fresh mind in a day or two after I've had some time to sort of start from the top again and think through things...

kytrinyx:
Yeah, totally! Usually when I run into blockers like this I end up having to tackle things from a completely different angle anyhow.

Issue #197, opened 2015-06-23
Improve error message when track / problem cannot be identified


kytrinyx:
A lot of people have trouble with the submit command, because they've put their solution files outside of the exercism working directory.

https://github.com/exercism/exercism.io/issues/2335

Perhaps we can improve the error message to explain what was expected as opposed to what actually happened ("you gave me this... but I'm looking for something like this.").

anxiousmodernman:
Got your email about 24 pull requests. 'Tis the season! :gift: 

Here is a draft of a more verbose error message this situation based on my paths.

```
2015/11/29 22:16:58 Unable to submit - Unable to identify track and file.

This usually occurs when you try to submit a file located outside of your configured exercises directory.

Your current directory is /Users/coleman/random-directory; exercism is looking for your exercises directory at /Users/coleman/exercism.

Try running `exercism submit <filename>` from inside /Users/coleman/exercism/<language>/<problem>.

For example, to submit the JavaScript "hello-world.js" problem, run `exercism submit hello-world.js` from the directory /Users/coleman/exercism/javascript
```

This is verbose, and maybe it's overkill, but based on [exercism/exercism.io#2335](https://github.com/exercism/exercism.io/issues/2335) it seems like more info is better than less.

To get the nice output for all OS's, we will probably need string templating + a helper function, so we could do something like this in `api/iteration.go`

```go

	// All the files should be within the exercism path.
	for _, filename := range filenames {
		if !iter.isValidFilepath(filename) {
			msg := getUnidentifiableTrackMessage(dir)
			return nil, errors.New(msg)
		}
	}
```

If anyone has comments, let me know. An alternative to this message might be introducing some kind of a subcommand/flag for `exercism submit`, so an error might instead look like

```
2015/11/29 22:16:58 Unable to submit - Unable to identify track and file.
Please run `exercism submit --help` for more info
```

kytrinyx:
I think you're right that more information is better than less. If they're getting the error, then we may as well show them the error message up front rather than making them re-run the command.

We can know for sure whether or not the problem is that they're outside of their configured directory, since we have both paths. If that's not the problem, let's show a more generic error and ask them to submit an issue.

What are the suggested ways to fix this?

1. Change their configuration. Can we guess what it should be and suggest that?
2. Move their code to the correct place. This involves understanding the DIR/track_id/slug thing.

A couple of things. Would using whitespace to show the diff between current/configured help?

```
Configured:  /Users/coleman/exercism
[...]

PR #261, opened 2015-12-12
Better error message when track/problem can't be identified

anxiousmodernman:
I've created two error messages for two scenarios.

1. If `submit` is called outside of the configured exercises directory, use the `msgSubmitCalledFromWrongDir` template
2. If `submit` is called *inside* the configured exercises directory, but the path seems weird, use `msgGenericPathError` template. I was able to simulate this error like this:

```
cd ~/exercism   # my exercism home
mkdir tmp
cd tmp
touch Clock.go
exercism submit Clock.go
```

Happy to rework any of this. :santa: :christmas_tree: :gift: 

I used `text/template` because the messages were quite long. The `iterationError` type is not necessary, strictly speaking, but I'd like to know what the rest of the team thinks.

It was mentioned in #197 that we could _suggest_ the proper path. I am unsure how to do that without having some kind of lookup based on file extension and the exercise slug. Even if I know that the file they're trying to submit is JavaScript, how can I tell it's the `hello-world` problem when they're submitting from a bad path/location?

### DIFF
--- a/api/iteration.go
+++ b/api/iteration.go
@@ -37,7 +37,7 @@ inside {{ .Configured }}
 For example, to submit the JavaScript "hello-world.js" problem, run
 "exercism submit hello-world.js" from this directory:
 
-{{ .Configured }}/javascript/helloworld
+{{ .Configured }}/javascript/hello-world
 
 You can see where exercism is looking for your files with "exercism debug".
 
@@ -71,14 +71,6 @@ type Iteration struct {
 	Solution map[string]string `json:"solution"`
 }
 
-type iterationError struct {
-	message string
-}
-
-func (i iterationError) Error() string {
-	return i.message
-}
-
 // NewIteration prepares an iteration of a problem in a track for submission to the API.
 // It takes a dir (from the global config) and a list of files which it will read from disk.
 // All paths are assumed to be absolute paths with symlinks resolved.
@@ -96,7 +88,7 @@ func NewIteration(dir string, filenames []string) (*Iteration, error) {
 	// All the files should be within the exercism path.
 	for _, filename := range filenames {
 		if !iter.isValidFilepath(filename) {
-			// User has run exercism submit in the wrong directory
+			// User has run exercism submit in the wrong directory.
 			return nil, newIterationError(msgSubmitCalledFromWrongDir, iter.Dir)
 		}
 	}
@@ -106,7 +98,7 @@ func NewIteration(dir string, filenames []string) (*Iteration, error) {
 
 	segments := strings.Split(path, string(filepath.Separator))
 	if len(segments) < 4 {
-		// Submit called from inside exercism directory, but the path
+		// Submit was called from inside exercism directory, but the path
 		// is still bad. Has the user modified their path in some way?
 		return nil, newIterationError(msgGenericPathError, iter.Dir)
 	}
@@ -131,7 +123,7 @@ func (iter *Iteration) RelativePath() string {
 }
 
 // isValidFilepath checks a files's absolute filepath and returns true if it is
-// within the configured exercise directory
+// within the configured exercise directory.
 func (iter *Iteration) isValidFilepath(path string) bool {
 	if iter == nil {
 		return false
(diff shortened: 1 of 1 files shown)


api/iteration.go
```diff
@@ -16,11 +18,48 @@ const (
...
+Configured directory:	{{ .Configured }}
+
+Try re-running "exercism fetch". Then move your solution file to the correct
+exercise directory for the problem you're working on. It should be somewhere
+inside {{ .Configured }}
+
+For example, to submit the JavaScript "hello-world.js" problem, run
+"exercism submit hello-world.js" from this directory:
+
+{{ .Configured }}/javascript/helloworld
```
kytrinyx: s/helloworld/hello-world

api/iteration.go
```diff
@@ -48,15 +96,19 @@ func NewIteration(dir string, filenames []string) (*Iteration, error) {
 	// All the files should be within the exercism path.
 	for _, filename := range filenames {
 		if !iter.isValidFilepath(filename) {
-			return nil, errUnidentifiable
+			// User has run exercism submit in the wrong directory
```
kytrinyx: This comment should end in a period, per the [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences) wiki.

api/iteration.go
```diff
@@ -48,15 +96,19 @@ func NewIteration(dir string, filenames []string) (*Iteration, error) {
...
+			return nil, newIterationError(msgSubmitCalledFromWrongDir, iter.Dir)
```
kytrinyx: I don't think we're getting any value from a custom type here. All we're doing is returning an error string, right?

api/iteration.go
```diff
@@ -16,11 +18,48 @@ const (
...
-	errUnidentifiable = errors.New("unable to identify track and problem")
-	errNoFiles        = errors.New("no files submitted")
-	utf8BOM           = []byte{0xef, 0xbb, 0xbf}
+	errNoFiles = errors.New("no files submitted")
+	utf8BOM    = []byte{0xef, 0xbb, 0xbf}
 )
 
+var msgSubmitCalledFromWrongDir = `Unable to identify track and file.
+
+It seems like you've tried to submit a solution file located outside of your
```
kytrinyx: Excellent call on the text/template!

api/iteration.go
```diff
@@ -48,15 +96,19 @@ func NewIteration(dir string, filenames []string) (*Iteration, error) {
...
 		}
 	}
 
 	// Identify the language track and problem slug.
 	path := filenames[0][len(dir):]
+
 	segments := strings.Split(path, string(filepath.Separator))
 	if len(segments) < 4 {
-		return nil, errUnidentifiable
+		// Submit called from inside exercism directory, but the path
```
kytrinyx: s/Submit called/Submit was called/ (for a complete sentence)

api/iteration.go
```diff
@@ -78,6 +130,8 @@ func (iter *Iteration) RelativePath() string {
 	return filepath.Join(iter.Dir, iter.TrackID, iter.Problem) + string(filepath.Separator)
 }
 
+// isValidFilepath checks a files's absolute filepath and returns true if it is
+// within the configured exercise directory
```
kytrinyx: This should end with a period.

api/iteration.go
```diff
@@ -109,3 +163,30 @@ func readFileAsUTF8String(filename string) (*string, error) {
 	s := string(decodedBytes)
 	return &s, nil
 }
+
+// newIterationError executes an error message template to create a detailed
+// message for the end user. A custom error type is returned.
+func newIterationError(msgTemplate string, configured string) error {
```
kytrinyx: oooh. Yeah, sorry. I get it.

I think we can still use the default errorString type for this, even though we have a helper method.

kytrinyx: Can simplify: `newIterationError(msgTemplate, configured string)`

kytrinyx:
This looks great! I really like the clarity of these error messages, and I think using a text/template was a really good call. I have a few minor comments.

api/iteration.go
anxiousmodernman: Did not even know this resource existed. Thank you!

anxiousmodernman:
I think this is done.

kytrinyx:
Thanks so much for taking the time to work through this so thoroughly!

Issue #77, opened 2014-04-12
Encourage improvement after submission


dpashkevich:
The `README` for the beer song exercise has an important section that reminds people of the end purpose of all the exercises:

```
## For bonus points

Did you get the tests passing and the code clean? If you want to, these are some additional things you could try:

* Remove as much duplication as you possibly can.
* Optimize for readability, even if it means introducing duplication.
* If you've removed all the duplication, do you have a lot of conditionals? Try replacing the conditionals with polymorphism, if it applies in this language. How readable is it?

Then please share your thoughts in a comment on the submission. Did this experiment make the code better? Worse? Did you learn anything from it?
```

Why not print it in the CLI tool after the user submits his next exercise?

eToThePiIPower:
Is anyone still interested in this issue? I was directed here from 24pullrequests and this sounds like a good place to start out.

kytrinyx:
Yeah, definitely still interested in this one. Up for grabs if you want it.

eToThePiIPower:
Great. I've got a preliminary version done. For the moment I thought it'd be best to only show it on the first iteration for each exercise (to keep down the annoyance factor), and hard-wrapped the lines to 80 characters.

Still working on a rotation/randomization scheme to keep some variety as you suggested. I think randomization will be easier to implement without having to keep track of the state somewhere.

kytrinyx:
Yeah, I think you're right -- we don't need to show it on subsequent iterations.

eToThePiIPower:
So I've got a simple random rotation going on the "header" to provide some variety (per your suggestion for adding variety). Currently:
* For bonus points
* Don't stop now: The fun's just begun
* Some tips to continue

It should be easy to add more (suggestions for creative phrases would be appreciated)

With enough of a selection of creative header phrases, I feel this would be enough variation. Randomizing the whole tips probably wouldn't be worth the extra complexity, at least at this point.

kytrinyx:
Agreed--let's start with this and see how it feels in action.

Tonkpils:
Going to close this since #262 was merged

PR #262, opened 2015-12-18
Encourage improvements after submission

eToThePiIPower:
As per issue #77, the displays extra hints after the users submit the first iteration of each problem.

The hints encourage users on ways to refactor, and then to participate in the community by commenting on their submission.

Variation is added through randomizing the opening phrase to keep users from automatically ignoring the message. Additional and more creative phrases here would be appreciated.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -3,10 +3,10 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"math/rand"
 	"os"
 	"path/filepath"
-  "math/rand"
-  "time"
+	"time"
 
 	"github.com/codegangsta/cli"
 	"github.com/exercism/cli/api"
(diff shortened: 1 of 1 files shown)


cmd/submit.go
```diff
@@ -5,6 +5,8 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+  "math/rand"
+  "time"
```
kytrinyx: Would you make sure this is run through `gofmt`, please?

cmd/submit.go
```diff
@@ -97,6 +99,28 @@ Your submission can be found online at %s
...
+
+* Remove as much duplication as you possibly can.
+* Optimize for readability, even if it means introducing duplication.
+* If you've removed all the duplication, do you have a lot of conditionals? Try
+  replacing the conditionals with polymorphism, if it applies in this language.
+  How readable is it?
+
+Then please share your thoughts in a comment on the submission. Did this
+experiment make the code better? Worse? Did you learn anything from it?
 `
```
kytrinyx: Should we use a text/template for this?

cmd/submit.go
```diff
@@ -97,6 +99,28 @@ Your submission can be found online at %s
...
+		msg += fmt.Sprintf("## %s", phrases[rand.Intn(len(phrases))])
+		msg += `
+
+Did you get the tests passing and the code clean? If you want to, these are some
+additional things you could try:
+
+* Remove as much duplication as you possibly can.
+* Optimize for readability, even if it means introducing duplication.
+* If you've removed all the duplication, do you have a lot of conditionals? Try
+  replacing the conditionals with polymorphism, if it applies in this language.
```
kytrinyx: Let's not mention polymorphism. How about suggesting trying to get rid of conditionals without suggesting a tactic?

cmd/submit.go
eToThePiIPower: Looking into text/template, it seems like it'll make the code less readable overall. It either:
* Adds overhead for creating the template plus a data structure (struct/map), while still requiring all the text in the source file, or
* Creates a run-time dependency on an external file using `ParseFiles`, which doesn't really make sense.

If anything it makes sense to define the static text as a const definite at the end of the file, eg:
```
    msg += fmt.Sprintf("## %s", phrases[rand.Intn(len(phrases))])
    msg += tips
...
const tips = `
(tips here)
`
```

kytrinyx: Yes, I think you're right. Defining it as a const would be better.

kytrinyx:
Thanks!

In the future, after pushing a fix, would you please add a comment that it has been done? I don't get notifications when someone pushes to a PR.

kytrinyx:
Now that this has been in production for a bit... what do you think?

Tonkpils:
So, I've used it a bit since this was implemented and I feel that after
first or second submission it just becomes noise. It's good encouragement
initially, but I stopped reading it after my second submission.

kytrinyx:
That's been my impression as well.

I think we should just take it back out, and keep working towards making the conversations on the site better.

PR #264, opened 2015-12-27
Fix version flag to use --version and -v.

### LINKED ISSUE #263: Version is broken
Right now if you say `exercism --version` it outputs the entire usage message.

This changed in f999e69e

```
    disambiguate debug and verbose flags
    
    The debug boolean flag and the debug command were ambiguous. Changing
    the `debug` boolean flag to `verbose` makes which options you're
    referring to more obvious.
```

@lcowell You deliberately added the app.HideVersion here. Is it safe to take it back off?

I'm thinking this would make sense:

````
--version, v
--verbose # no shorthand
```

Tonkpils:
-v was reserved for verbose at https://github.com/exercism/cli/commit/f999e69e5290cec6c5c9933aecc6fddfad8cf019  but then removed https://github.com/exercism/cli/commit/4aa6a8665386a7ec5f4aff811b64f9f3aa862f0c because it caused issues with the cli package.

Fixes #263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The exercism CLI follows [semantic versioning](http://semver.org/).
 
 ## Next Release
 * **Your contribution here**
+* [#264](https://github.com/exercism/cli/pull/264) Fix version flag to use --version and --v - [@Tonkpils](https://github.com/Tonkpils)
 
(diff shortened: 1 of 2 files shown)


kytrinyx:
OK, excellent. I'm going to release a new version so that the homebrew test will work properly. Thanks!

PR #266, opened 2016-01-01
Fix grammar in post-submit message

vassilevsky:


kytrinyx:
Oh no. I can't believe I missed that. Thank you!

PR #269, opened 2016-01-10
Tweak API key flag in 'configure' usage statement

kytrinyx:
We're referring to /account/key elsewhere which is only the api key stuff. This keeps the CLI consistent with the rest of the messaging.

### DIFF
--- a/exercism/main.go
+++ b/exercism/main.go
@@ -80,7 +80,7 @@ func main() {
 				},
 				cli.StringFlag{
 					Name:  "key, k",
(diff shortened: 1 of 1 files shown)


Tonkpils:
Just for reference, we still show the API key in the `/account` page on the front-end. We should maybe remove it from there so that it's in an unified place.

kytrinyx:
Agreed. Actually, let's link to it from `/account` instead of showing it... in case people go looking for it. (I'll take care of it).

PR #271, opened 2016-01-13
Add debug API call timing.

### LINKED ISSUE #268: Add latency check on Debug command
It would be good to have a measure of the latency to the GitHub API and the Exercism Apps and display them in the debug command.

Tonkpils:
Also, increased the timeout to 20 seconds which gives slow clients a bit more time to complete the calls

Fixes #268 

Any feedback on this would be great. It's a bit of a mess since I added a lot of code to handle concurrent API ping checks. We are also checking against Github's API to ensure it's not just our API that is timing out. 

The output of the command now looks like this:

```sh
**** Debug Information ****
Exercism CLI Version: 2.2.3
Exercism CLI Latest Release: 2.2.3
OS/Architecture: darwin/amd64
Build OS/Architecture /
Home Dir: /Users/leo
Config file: /Users/leo/.exercism.json
API Key: 888c953db23817e4db310f10bf476222af806cc9
Exercises Directory: /Users/leo/code/exercisms
Testing API endpoints reachability
	* GitHub API: https://api.github.com/ [connected] 287.315532ms
	* XAPI: http://x.exercism.io [connected] 575.072309ms
	* API: http://exercism.io [connected] 863.000295ms

If you are having trouble and need to file a GitHub issue (https://github.com/exercism/exercism.io/issues) please include this information (except your API key. Keep that private).
```

### DIFF
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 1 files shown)


kytrinyx:
+1 on both increasing the timeout, and checking against GitHub's API. I also like getting more data out of the ping.

LGTM!

PR #273, opened 2016-01-21
Guard against submitting _spec.* files and READMEs

### LINKED ISSUE #272: Guard against submitting the wrong file
I was moving a little too fast tonight and [accidentally submitted `hamming_spec.lua` instead of `hamming.lua`](http://exercism.io/submissions/7f35bb6eb9684a48ae17583c929f9888). It would be nice if the CLI tool were smart enough to know if the user is accidentally submitting a README or test cases in lieu of a solution.

Would it make sense for the CLI tool to check the filename of the file you're submitting and guard against certain filenames? We could check for things like: 

- anything containing "README"
- (name of the exercise) + "_spec.lua"
- (name of the exercise) + "_test.clj"

daveyarwood:
Closes #272.

I see that we allow users to submit test files if they include a `--test` flag (#189) -- should we do the same to allow submitting READMEs? If so, I propose we change the `--test` flag to a `--force` flag and let it apply to READMEs and any other files we may want to complain about in the future.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"path/filepath"
 	"regexp"
+	"strings"
 )
 
 const (
@@ -21,4 +22,8 @@ func isTest(path string) bool {
 		return true
 	}
 	return regexp.MustCompile(`[\._-]?([tT]est|[sS]pec)`).MatchString(name)
+}
+
+func isREADME(path string) bool {
+	return strings.Contains(path, "README")
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -17,8 +17,8 @@ func isTest(path string) bool {
 
 	file := filepath.Base(path)
 	name := file[:len(file)-len(ext)]
-	if name == "test" {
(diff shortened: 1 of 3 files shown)


Tonkpils:
Nope, I don't think we ever want to allow submitting the README. This looks great :)

cmd/cmd.go
```diff
@@ -17,8 +18,12 @@ func isTest(path string) bool {
 
 	file := filepath.Base(path)
 	name := file[:len(file)-len(ext)]
-	if name == "test" {
+	if name == "test" || name == "spec" {
```
kytrinyx: perl doesn't have test or spec in the name, for what it's worth. Languages that have a different pattern for the test files have this defined in the language config, so we could expose it via the API (and perhaps write it to a dotfile somewhere that we can load and check.

daveyarwood: I like that idea, seems much safer.

kytrinyx:
OK cool, I'll add an issue for it.

kytrinyx:
https://github.com/exercism/x-api/issues/95

Issue #277, opened 2016-01-23
`download` fails with submissions with directory structure


petertseng:
The recent change in https://github.com/exercism/exercism.io/commit/2dc984d4030c2826af317b1d287a8fc9e9f9ce20 caused directory structures to be preserved in submissions. But this causes downloads to fail. Example:

```
$ exercism download 002951e89f1e477db457ac70ae4001d1
2016/01/22 07:09:18 Unable to write file hand/hand.go: open /home/pt/exercism/solutions/petertseng/go/poker/002951e89f1e477db457ac70ae4001d1/hand/hand.go: no such file or directory
```

As you can see, it's failed to create the `hand` directory (the directory `/home/pt/exercism/solutions/petertseng/go/poker/002951e89f1e477db457ac70ae4001d1` got created just fine, that's no problem).

Issue #276, opened 2016-01-22
should files from `download` be created non-executable instead of executable?


petertseng:
Currently they're created executable, as https://github.com/exercism/cli/blob/master/cmd/download.go#L42 and https://github.com/exercism/cli/blob/master/cmd/download.go#L49 would have it. Need to think about whether that should be the case. Is the idea that some language tracks' submission files are meant to be executed directly?

However, when doing `fetch`, files are written 0644: https://github.com/exercism/cli/blob/master/user/item.go#L71 . And given that the Ruby track has an [automated feedback for when Rubys submissions have shebangs](https://github.com/exercism/rikki/blob/master/comments/analyzer/ruby/shebang/shebang.md), it makes it seem more likely those 0755 should be change to 0644.

Any specific reason to keep at 0755?

kytrinyx:
I know in Ruby the test files are executable so that people have the option of saying `ruby path/to/testfile` or just `path/to/testfile`. But I don't actually have very strong feelings about it.

We could use the shebang to check whether a file should be 0644 or 0755.

petertseng:
So you are thinking "if first line in file is shebang, make it 0755, else make it 0644?"

I could see that making sense. If we are in agreement, I can probably make it happen, though I'm a bit unclear on how I'd test.

petertseng:
My tentative test plan is: Make my change, build client from git, and just download a few submissions and see if they get created with the right modes.

Any place where it'd be appropriate to add an automated test for this?

kytrinyx:
Yeah, that was what I was thinking. Building from git is how I change things that I don't have automated tests for. If we move the code that writes the file out of `cmd/download.go` and onto a method on `Submission` then it should be fairly straight forward to write a test for it.

Tonkpils:
I don't like to encourage people to simply run an executable they've fetched from a remote. I think we should remove the shebang and simply have them execute the file with `ruby /path/to/test`. Thoughts?

kytrinyx:
Yeah, good point.

I can't remember the arguments for adding it in Ruby. I think if they want to make it executable, they're welcome to do it themselves.

OK, for now, let's make it 0644 on everything, and then I'll open a ticket on the Ruby track to remove the shebangs and add some documentation to let people know they can chmod and shebang themselves if they want.

petertseng:
OK, everything 0644 then. In that case since there's no new logic added I am not really motivated to add new automated tests; I'll just test by building from git.

[...]

PR #278, opened 2016-01-23
Download: Create files with 0644, support directory structures

petertseng:
This fixes two issues with downloads #276 and #277. More detailed explanations in the corresponding commit messages.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -46,6 +46,10 @@ func Download(ctx *cli.Context) {
 
 	for name, contents := range submission.SolutionFiles {
 		filename := strings.TrimPrefix(name, strings.ToLower("/"+submission.TrackID+"/"+submission.Slug+"/"))
+		dir := filepath.Dir(filename)
+		if err := os.MkdirAll(fmt.Sprintf("%s/%s", path, dir), 0755); err != nil {
+			log.Fatal(err)
+		}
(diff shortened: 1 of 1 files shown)


cmd/download.go
```diff
@@ -39,14 +39,18 @@ func Download(ctx *cli.Context) {
...
+		if err := ioutil.WriteFile(fmt.Sprintf("%s/%s", path, name), []byte(contents), 0644); err != nil {
 			log.Fatalf("Unable to write file %s: %s", name, err)
 		}
 	}
 
 	for name, contents := range submission.SolutionFiles {
 		filename := strings.TrimPrefix(name, strings.ToLower("/"+submission.TrackID+"/"+submission.Slug+"/"))
-		if err := ioutil.WriteFile(fmt.Sprintf("%s/%s", path, filename), []byte(contents), 0755); err != nil {
+		dir := filepath.Dir(filename)
+		if err := os.MkdirAll(fmt.Sprintf("%s/%s", path, dir), 0755); err != nil {
```
petertseng: I don't know how similar this code path is starting to get with https://github.com/exercism/cli/blob/master/user/item.go#L51 - maybe could use some unification? I'm unsure of where I'd put the code though, so I'd take suggestions on that.

kytrinyx: Let's just leave it for now. We can do some refactoring in a second step.

kytrinyx:
Thank you!

PR #282, opened 2016-01-26
Remove random encouragement after submitting

kytrinyx:
The post-submit message turned out to not work. After seeing it
the first time it just becomes noise, and you end up hunting for the
important information.

This simplifies the message, removing unnecessary noise. The only
information necessary is: language track, problem name, and the
URL of your iteration.

I put the url of the iteration alone on its own line instead of saying
"your submission can be found online at URL", since the extra verbiage
is superfluous, and it also prevents double-clicking on the line to
highlight the url and paste it easily into the browser.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -3,10 +3,8 @@ package cmd
 import (
 	"fmt"
 	"log"
-	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


Tonkpils:
:+1:

PR #283, opened 2016-01-26
Output configuration after calling configure command

kytrinyx:
$ exercism configure --key=abc123

    Configuration written to /Users/kytrinyx/.exercism.json
      --key=abc123
      --dir=/Users/kytrinyx/exercism
      --host=http://localhost:4567
      --api=http://x.exercism.io

By extension this will also give you your current config values if you
call configure without any flags. It's faster than the debug command
(because of the latency checks), and it's useful during development
when you might be talking to a local version of the app in one moment
and then need to switch to the production version the next.

### DIFF
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -35,6 +35,9 @@ func Configure(ctx *cli.Context) {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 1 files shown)


Tonkpils:
:+1: :sparkles:

PR #284, opened 2016-01-28
Update the homebrew release instructions

kytrinyx:
They moved exercism into the "real" homebrew, so we should no longer try to work with the homebrew-binary repository.


Tonkpils:
:+1: great catch

kytrinyx:
You're fast!

PR #290, opened 2016-02-03
Fix `upgrade` command on some Linux systems

### LINKED ISSUE #288: `exercism upgrade` fails with `text file busy`
I just tried to upgrade the tool from the command-line and got the following error:

```
$ exercism upgrade
Downloading exercism-linux-64bit.tgz
2016/02/01 09:30:13 open /home/jbaiter/.bin/exercism: text file busy
```

jbaiter:
Previously the still open executable file was directly written to, which caused the program to abort with a `text file busy` message on some systems.

Now the following approach is taken:

1. Move the old file to a backup path
2. Try to write to the original path
3. If opening/writing the original path fails: Move backup to old location
4. If upgrade is successful: Remove backup file

### DIFF
--- a/cmd/release_utils.go
+++ b/cmd/release_utils.go
@@ -85,19 +85,25 @@ func installTgz(source *bytes.Reader, dest string) error {
 			return err
 		}
 
-		// Remove the file before overwriting it.
+		// Move the old version to a backup path that we can recover from
+		// in case the upgrade fails
+		destBackup := dest + ".bak"
 		if _, err := os.Stat(dest); err == nil {
-			os.Remove(dest)
+			os.Rename(dest, destBackup)
 		}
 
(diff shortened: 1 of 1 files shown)


kotp:
It really helps to have the "fixes #288" in the body of the commit as compared to the title or summary as it links there in the web gui.  This is more a complaint to Github itself than anyone that puts that in the title though.

jbaiter:
Sorry for that, I somewhat mindlessly retyped the original commit title.

kotp:
It's the functionality that that statement, when combined into master branch, it will auto-close the issue, so it is useful.  Unsure if it closes from the first line or not, but in the body on the web interface, it links.  So a couple of functional things for having it.  I think it should link in the title as well, not just in the body, but that may just be me.

Tonkpils:
Thanks for this PR! I'll take a look at it soon.

Tonkpils:
I tested this on Linux and OSX and works great

PR #291, opened 2016-02-20
Cleanup CHANGELOG user links

Tonkpils:
This will allow us to clean simply add the reference to the username rather than copying the link to their profile every time if they have already contributed.


kytrinyx:
Nice! Thanks for that.

Issue #296, opened 2016-03-03
exercism status returns an internal server error when there's no api key set locally


mrageh:
When a user does not have their api key set and `exercism debug` returns:
```
**** Debug Information ****
Exercism CLI Version: 2.2.4
Exercism CLI Latest Release: 2.2.4
OS/Architecture: darwin/amd64
Build OS/Architecture /
Home Dir: /Users/Adam
Config file: /Users/Adam/.exercism.json
API Key:
Exercises Directory: /Users/Adam/programming/cisms/
Testing API endpoints reachability
	* API: http://exercism.io [connected] 170.057673ms
	* GitHub API: https://api.github.com/ [connected] 322.724326ms
	* XAPI: http://x.exercism.io [connected] 649.753655ms
```

If that user tries to check the status of a language with `exercism status language`, the server returns a 500 I think:
```
2016/03/03 09:36:45 an internal server error was received.
Please file a bug report with the contents of 'exercism debug' at: https://github.com/exercism/exercism.io/issues
```

Is this a bug and should we fix it?

I'd be happy to take a stab at it by making changes to `exercism/cli`.

Tonkpils:
The server should definitely not blow up with an internal server error if the API key is not provided. It should return with an unauthorized or something similar in which case the error message on our end would be that response. 

One thing I noticed, if the API Key is not set, `exercism debug` should probably display a message stating that it should be set to access some functionality. 

As far as ensuring all the commands that require API key on the server end have an API key present, I believe was discussed. The consensus was that we should let the server tell us that in case the API endpoints change requirements for API key presence or not.

kytrinyx:
Yeah, the server side needs to be fixed for this. I'm not sure what's going on, because the API supposedly checks for the presence of the key. I'm looking into this now.

kytrinyx:
OK, that was quick.

```
http://localhost:4567/api/v1/tracks/ruby/status # status 401
```

```
http://localhost:4567/api/v1/tracks/ruby/status?key= # status 500
```

kytrinyx:
Fixed in exercism/exercism.io@d0c061575be65f03ceddcf096d7577f9351663e7

mrageh:
Wow @kytrinyx that was quick :+1: 
Closing this issue as it was just (fixed)[https://github.com/exercism/exercism.io/commit/d0c061575be65f03ceddcf096d7577f9351663e7]

Tonkpils:
[...]

PR #297, opened 2016-03-04
Enhance debug output when api key is absent

mrageh:
Previously if the API key was not, some of the features of exercism cli would
not work, for example exercism status.

This commit just prompts the user to set their API key to access the full
features of exercism cli, when its not set.

This fix was suggested in https://github.com/exercism/cli/issues/296 by @Tonkpils

### DIFF
--- a/cmd/debug.go
+++ b/cmd/debug.go
@@ -64,10 +64,14 @@ func Debug(ctx *cli.Context) {
 
 	if configured {
 		fmt.Printf("Config file: %s\n", c.File)
-		fmt.Printf("API Key: %s\n", c.APIKey)
+		if c.APIKey != "" {
+			fmt.Printf("API Key: %s\n", c.APIKey)
+		} else {
+			fmt.Println("API Key: Please set your API Key to access all of the CLI features")
+		}
 	} else {
 		fmt.Println("Config file: <not configured>")
-		fmt.Println("API Key: <not configured>")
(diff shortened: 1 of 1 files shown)


cmd/debug.go
```diff
@@ -64,10 +64,14 @@ func Debug(ctx *cli.Context) {
...
-		fmt.Printf("API Key: %s\n", c.APIKey)
+		if c.APIKey != "" {
+			fmt.Printf("API Key: %s\n", c.APIKey)
+		} else {
+			fmt.Println("API Key: Please set your API Key to access all of the CLI features")
+		}
 	} else {
 		fmt.Println("Config file: <not configured>")
-		fmt.Println("API Key: <not configured>")
+		fmt.Println("API Key: Please set your API Key to access all of the CLI features")
```
Tonkpils: Let's leave this with the `<not configured>` text as it's on the case where the CLI is not configured at all and it makes more sense.

mrageh:
@Tonkpils when I opened the exercism status issue, I didn't have exercism configured. Which meant that when I tried to run exercism status, the server returned a 500. 

The <not configured> message for the api key was not in helpful in the above instance. I think we should always prompt users to set their api key, when it's not set.

Tonkpils:
Right, this was an issue with the server, the server will now return the
appropriate message when the API key is not set which should fix the issue
of exercism status.

Tonkpils:
@mrageh actually it doesn't hurt to mention the message on both instances so :+1:

Issue #295, opened 2016-03-02
Output track-specific message at the end of the list command


kytrinyx:
Right now we always say `exercism fetch ruby matrix` at the end of the list command.

I suggest that we always use the track id, and the first problem in the list as the example.

```
$ exercism list elixir

....

e.g.:

exercism fetch elixir hello-world
```

PR #299, opened 2016-03-17
Use track ID and first problem on exercism list output

Tonkpils:
Fixes https://github.com/exercism/cli/issues/295


kytrinyx:
Nice!

PR #301, opened 2016-03-21
Return error message for unknown track status

### LINKED ISSUE #298: Unknown track gives incorrect error message
When you try to fetch the status of an unknown track, the error message is a bit odd:

<pre>
$ exercism status haskel

You have yet to begin the haskel track!
</pre>

neslom:
Fixes #298 

This is my first attempt at a contribution here, so I hope I didn't miss anything too obvious.. :smile:


api/api.go
```diff
@@ -270,6 +270,10 @@ func (c *Client) Status(trackID string) (*StatusInfo, error) {
 
 	defer res.Body.Close()
 
+	if _, err := c.List(trackID); err != nil {
+		return nil, err
+	}
+
```
kytrinyx: Do we not need to do anything with the result of the List?

cmd/status.go
```diff
@@ -23,9 +23,14 @@ func Status(ctx *cli.Context) {
...
-	status, err := client.Status(args[0])
+	trackID := args[0]
+	status, err := client.Status(trackID)
 	if err != nil {
-		log.Fatal(err)
+		if err == api.ErrUnknownTrack {
+			log.Fatalf("There is no track with ID '%s'.", trackID)
+		} else {
+			log.Fatal(err)
+		}
```
kytrinyx: This looks right to me.

kytrinyx:
The fix to the command looks good, I just have the one question about the change in the API.

api/api.go
neslom: No, we just need to return an error here if the Track doesn't exist. 

I looked into the API a bit and it looks like the `/tracks/<trackID>/status` endpoint returns the same response for both valid and invalid tracks, so we just need to first check that the track exists and return an error if it doesn't instead of parsing the response.

If there's no error here, then `Status` looks at the fetched problems and recent problems, so we don't need the `List` results.

kytrinyx: /tracks/:track_id/status is the endpoint for a problem named "status" which we don't have.

neslom: Sorry, I should have been more specific.

This is the endpoint I was referencing:

`http://exercism.io/api/v1/tracks/elixir/status?key=<someKey>`

kytrinyx: Oh, you mean http://exercism.io/api/v1/tracks/:id/status?key=$KEY

kytrinyx: OK, I think that the fix is to improve that endpoint so that it says whether or not the track exists. Are you comfortable in Ruby? If not, no worries, I can do this easily.

neslom: I'll fork that repo right now :+1:

neslom:
Hey @kytrinyx, I submitted https://github.com/exercism/exercism.io/pull/2806 to address this. I can remove the changes I made in the `api` package now :+1:

kytrinyx:
Oh, cool! Yeah, this is better.

neslom:
Thanks @Tonkpils! I hope I can help out again!

PR #302, opened 2016-03-23
add check if the submitted filename is a directory

alebaffa:
I tried to add some tests, but I didn't know if to add a new one into_ paths/paths_test.go_ (so basically to testing the function _IsDir()_), or create a new _cmd/submit_test.go_. 
Let me know if you need a test.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -9,6 +9,7 @@ import (
 	"github.com/codegangsta/cli"
 	"github.com/exercism/cli/api"
 	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/paths"
 )
 
 // Submit posts an iteration to the API.
@@ -58,6 +59,10 @@ func Submit(ctx *cli.Context) {
 
 		if isREADME(filename) {
 			log.Fatal("You cannot submit the README as a solution.")
+		}
+
+		if paths.IsDir(filename) {
+			log.Fatal("Please specify each file that should be submitted, e.g. `exercism submit file1 file2 file3`.")
 		}
 
 		file, err := filepath.Abs(filename)


cmd/submit.go
```diff
@@ -60,6 +61,10 @@ func Submit(ctx *cli.Context) {
 			log.Fatal("You cannot submit the README as a solution.")
 		}
 
+		if paths.IsDir(filename) {
+			log.Fatal("You cannot submit a directory. Submit only the solution file.")
```
kytrinyx: You can submit more than one file.

How about:

```
Please specify each file that should be submitted, e.g. `exercism submit file1 file2 file3`.
```

kytrinyx:
No, don't worry about a test. Eventually we'll be adding tests for the commands, but it's going to take some thought to get it right.

alebaffa:
Just for my curiosity: what are the considerations you are making for the tests for the commands? (don't know if it's the right place to ask it.. :P )

Tonkpils:
> what are the considerations you are making for the tests for the commands?

We are using a 3rd party library to handle the CLI interface. In order to test the commands we have to mock the argument passed into the command functions which would require a change of signature of all the functions to pass an interface of our own. However, the CLI package takes a specific struct which doesn't allow for this kind of change. 

Unless that's changed in the CLI package, that's the one thing stopping us from writing tests around the commands.

kytrinyx:
This looks good to me. Would you mind squashing the commits, please?

alebaffa:
Voila :)

kytrinyx:
Lovely, thank you!

mhelmetag:
Fixes #302

Issue #305, opened 2016-04-04
Solved exercises showed as "not submitted"


lucasloisp:
When I run `exercism status haskell`
I get
```
Most recently submitted exercise:
 - accumulate (submitted on April 4, 2016)

Exercises fetched but not submitted:
 - accumulate
```
I've already solved 'accumulate', and it shows in my dashboard, but I can't seem to get the tool to recognize it.
Note, this has never happened with other exercises.

lucasloisp:
I realize now this has been discussed on issue #303, my bad.

Issue #303, opened 2016-03-29
Incorrect track status information


behrtam:
`exercism status scala` gives me totally incorrect information. For example in the *scala* track, I have submitted [`Bob`](http://exercism.io/exercises/cb2f63d644cf499a939e9795eb186326), [`Hamming`](http://exercism.io/exercises/6b72f7e37f4f4b8ba9bb9d219ec52bac), [`Word Count`](http://exercism.io/exercises/0e4612a1870d4b839ca8376dce6539c4) and [`Anagram`](http://exercism.io/exercises/3a722130d44948459fd539ae0829f99c) but my status looks like this:

```
Your status on the scala track:

Most recently submitted exercise:
 - anagram (submitted on February 10, 2016)

Exercises fetched but not submitted:
 - anagram
 - bob
 - hamming
 - nucleotide-count
 - word-count

Exercises skipped:
```

I don't know if it's checkt online or locally but I didn't delete them and even `restore` doesn't change anything.

**** Debug Information ****
Exercism CLI Version: 2.2.4
Exercism CLI Latest Release: 2.2.4
OS/Architecture: darwin/amd64
Build OS/Architecture /
Home Dir: /Users/tmo
Config file: /Users/tmo/.exercism.json
Exercises Directory: /Users/tmo/exercism
Testing API endpoints reachability
	* API: http://exercism.io [connected] 255.35963ms
	* GitHub API: https://api.github.com/ [connected] 382.065413ms
	* XAPI: http://x.exercism.io [connected] 394.421512ms

kytrinyx:
This is checked on the API. There's an open pull request in the works to fix it: https://github.com/exercism/exercism.io/pull/2810

I'm going to close this as a duplicate of https://github.com/exercism/exercism.io/issues/2804

behrtam:
Sorry didn't check the PRs.

kytrinyx:
No worries, there's no easy way to keep track of everything everywhere in this project!

PR #307, opened 2016-04-13
Pass API key when fetching individual exercises

kytrinyx:
We're tracking what people fetch, and if we don't pass the key
we can't know who is fetching it.

Tangentially related to #303 and #305


kytrinyx:
My bad. Fixing.

Tonkpils:
🎉 cool!

PR #312, opened 2016-04-29
Add newline when printing usage

jppunnett:
A little thing, but annoying nonetheless: added newline to avoid command prompt being shifted to right when running `exercism list` and it displays usage.


kytrinyx:
Nice catch, thank you!

PR #317, opened 2016-05-07
Update dependency: codegangsta/cli

manusajith:


lcowell:
Thanks for the contribution. How did you upgrade codegangsta/cli ? I got a different version when I upgraded using `godep update github.com/codegangsta/cli`. See my Godep.json [here](https://github.com/exercism/cli/compare/master...upgrade_codecangsta_cli).

I'm wondering if you could also fix the deprecation warnings that you'd get when you run the app. eg. `exercism debug` will produce:
> DEPRECATED Action signature.  Must be `cli.ActionFunc`.  This is an error in the application.  Please 
> contact the distributor of this application if this is not you.  See 
> https://github.com/codegangsta/cli/blob/master/CHANGELOG.md#deprecated-cli-app-action-signature

I'm thinking for the first round of changes we just get the app compiling. We now have the option of returning an error from the action and cg/cli will display the error and exit. cg/cli also has a new error type `cli.ExitError`, this allows us to set an exit value on exit. I think that's a bigger change that belongs in its own PR.

manusajith:
@lcowell I will have a look at this tonight, and will keep you updated.

lcowell:
Looks good. If the first 3 commits are all about `updating codegangsta/cli`, I think it would be best if you could squash them in to a single commit.

Tonkpils:
Thank you for your contribution!

PR #319, opened 2016-05-09
Fix related solutions link

jbranchaud:
The first part of the link is missing an initial forward slash.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -100,5 +100,5 @@ func Submit(ctx *cli.Context) {
 	}
 
 	fmt.Printf("Your %s solution for %s has been submitted. View it here:\n%s\n\n", submission.Language, submission.Name, submission.URL)
-	fmt.Printf("See related solutions and get involved here:\n%stracks/%s/exercises/%s\n\n", c.API, iteration.TrackID, iteration.Problem)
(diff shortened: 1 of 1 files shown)


Tonkpils:
Thanks for this! 

I think we should write a method that returns a standardized URL from the configuration or at least it standardizes it when it reads it from the configuration file (with or without `/` suffix) so that we don't run into these situations in the future and since this can be set on the configuration file.

kytrinyx:
Agreed. @jbranchaud do you want to tweak this PR to handle both cases?

jbranchaud:
Yes, I'll try to update the PR accordingly.

@Tonkpils what is "the configuration" that you are referring to above?

Tonkpils:
the `.exercism.json` file contains the API urls used by the client and gets passed around through all the commands as the `Configure` struct.

kytrinyx:
This is becoming pretty critical to solve—we get new bug reports about it on a daily basis.

@jbranchaud do you still want to take a stab at this? If so, would you be able to do it in the next few days? If not I think one of the @exercism/cli-committers needs to just go ahead and do it.

Tonkpils:
@kytrinyx I was about to pick off this commit and add the extra functionality if that's ok with @jbranchaud

kytrinyx:
@Tonkpils I'd appreciate that. Once that's done I'll cut a new release.

Tonkpils:
Closing this in favor of https://github.com/exercism/cli/pull/328

PR #322, opened 2016-05-18
Explain how to run the tests locally

jbranchaud:
I'm not sure if I got the wording quite right, but I think it's useful to note (in the `README`) how to run the tests locally.

I got this from the `.travis.yml` file. Perhaps there is a different, preferred way of running the tests for this project?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Follow the directions on http://golang.org/doc/install
 
 Please provide tests for the changes where possible.
 
+To run the tests locally, use `go test $(go list ./... | grep -v /vendor/)`.
+
 At the moment the CLI commands are not tested, so if you're adding a new
 command don't worry too hard about tests.
 


lcowell:
Thanks! This definitely works and is the only way I'd know to do ti too. It feels a little onerous. Using a tool like `gb` would allow us to simply say `gb test`. Anyone else know a simpler way to call the tests?

Tonkpils:
@jbranchaud @lcowell `go test ./...` works just fine for me. It doesn't seem like `vendor` folder includes any test files maybe `godep` handles this when vendoring?

Tonkpils:
Yeah, it seems you have to explicitly vendor with a `-t` flag when saving deps if you want to include test files with Godep so the instructions should just be `go test ./...` https://github.com/tools/godep#how-to-use-godep-with-a-new-project

Tonkpils:
Closed by https://github.com/exercism/cli/commit/4edbbc88f48c4dfd872060febc94d6d6c3a110b9

PR #323, opened 2016-05-20
add a slash between the API url and the rest of the track path

rousisk:
Hi folks,

Just installed the cli and tried a golang exercise. After submitting my first solution, I got the following output:

```
See related solutions and get involved here:
http://exercism.iotracks/go/exercises/hello-world
```

As you can see, there is a missing slash between the tld and the rest of the path. 

This patch, hopefully, fixes that. Since this is my first touch with golang, I wasn't sure how to test my change :-)


Tonkpils:
@rousisk thanks for taking the time to work on this. This is actually a duplicate of https://github.com/exercism/cli/pull/319

rousisk:
I checked the issues first but not the open PRs... Sorry for the duplication!

PR #324, opened 2016-05-27
update codegangsta/urfave url

blackerby:
Minor, but seems worth doing: https://github.com/urfave/cli#cli


kytrinyx:
I agree that we should do this. You will need to update Godeps and the vendored library as well.

Tonkpils:
Closing this since #333 was merged

Issue #327, opened 2016-05-29
`exercism submit` prints incorrect link for related solutions


AlexLov:
Example:

~~~
~/exercism/go/leap/ ❯ exercism submit leap.go
Your Go solution for Leap has been submitted. View it here:
http://exercism.io/submissions/7a94905ea232476b894ca9aed87e2e2b

See related solutions and get involved here:
http://exercism.iotracks/go/exercises/leap
~~~

Where http://exercism.iotracks/go/exercises/leap should be http://exercism.io/tracks/go/exercises/leap
Lost '/' between exercism.io and tracks.

I've fixed it by changing api address in ~/.exercism.json:
from 
"api": "http://exercism.io"
to
"api": "http://exercism.io/"

But submit command should print correct link in both cases.

Tonkpils:
We are working on a fix for this https://github.com/exercism/cli/pull/319

PR #328, opened 2016-05-29
append solution URL path consistently

Tonkpils:
URL.Path adds the trailing '/' between the hostname and the path if it doesn't exist

Fixes https://github.com/exercism/cli/issues/327


kytrinyx:
Thank you!

Issue #335, opened 2016-06-26
exercism fetch TRACK_ID --all


kytrinyx:
I've gotten a few requests for the ability to download all the exercises for a languages before getting on a long flight.

What do you think of adding a flag `--all` to the fetch command?

This wouldn't have to be fancy; I'd be totally OK with a first iteration that just loops through the full list.

guilhermeepaixaoo:
This flag is for download all exercises from the language ?

Tonkpils:
@guilhermeepaixaoo from the implementation it seems like it

Tonkpils:
Fixed by #336

PR #336, opened 2016-07-01
Add '--all' flag to 'fetch' command

neslom:
Fetch all exercises for a given track

As mentioned in issue #335

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -19,19 +19,25 @@ func Fetch(ctx *cli.Context) error {
 	client := api.NewClient(c)
 
 	args := ctx.Args()
-	problems, err := client.Fetch(args)
+	var problems []*api.Problem
 
 	if ctx.Bool("all") {
 		if len(args) > 0 {
 			trackID := args[0]
-			problems = fetchAll(trackID, client)
+			p, err := client.FetchAll(trackID)
+			if err != nil {
+				log.Fatal(err)
+			}
+			problems = p
 		} else {
 			log.Fatalf("You must supply a track to fetch all exercises")
 		}
-	}
-
-	if err != nil {
-		log.Fatal(err)
+	} else {
+		p, err := client.Fetch(args)
+		if err != nil {
+			log.Fatal(err)
+		}
+		problems = p
 	}
 
 	submissionInfo, err := client.Submissions()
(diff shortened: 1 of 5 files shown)


neslom:
Hey @kytrinyx! I was looking through the issues and thought I would take a quick stab at this. I'll be offline until Tuesday, but let me know what you think!

cmd/fetch.go
```diff
@@ -18,7 +18,18 @@ func Fetch(ctx *cli.Context) error {
 	}
 	client := api.NewClient(c)
 
-	problems, err := client.Fetch(ctx.Args())
+	args := ctx.Args()
+	problems, err := client.Fetch(args)
```
Tonkpils: We shouldn't perform this fetch if the `all` flag is set to true. What if we moved it to the `else` of the following `if`?

cmd/fetch.go
```diff
@@ -71,3 +82,20 @@ func setSubmissionState(problems []*api.Problem, submissionInfo map[string][]api
 
 	return nil
 }
+
+func fetchAll(trackID string, client *api.Client) []*api.Problem {
```
Tonkpils: Could we make it so that this function returns the list of problems and the error and `log.Fatal` on the caller if the error is not nil?

Also, what do you think about adding this to the `api.Client` in case we add an API in the future to fetch a list of exercises?

neslom:
Sounds good, @Tonkpils. I'm out of town for the holiday weekend, but will follow up on this when I get back Monday or Tuesday.

cmd/fetch.go
kytrinyx: +1 for returning the error from here.

neslom:
Hey @Tonkpils, I updated the PR with your suggestions. I started to add a test for `FetchAll`, but wanted to get your input on this before moving forward. `FetchAll` is essentially just calling `List` and then `Fetch` for each item returned by `List`, so I was unsure if we needed to add a test for this. Also, it looks like the fixtures for `tracks

Tonkpils:
@neslom I think having tests would be great but if it's too difficult then it's ok for now. The code seems fine. Once we iterate over this and have an actual API then we can add the fixture and simulate that request for the test.

neslom:
hey @Tonkpils, I ended up coming up with a test for this! let me know if I did something weird (I'm quite new to Go). I might have gone overboard with the assertions, but I explained my reasoning in the commit message

Tonkpils:
That looks great! Thank you so much for your contribution!

kotp:
When does this become available on the client (via `exercism upgrade` hopefully)?

Tonkpils:
@kotp we'd need to cut a release out. I'll get to that tonight

kytrinyx:
@Tonkpils if you run into any trouble holler—I've tried my best to document the release process, but if I've missed anything let me know and we'll improve the documentation along the way.

PR #339, opened 2016-07-22
Don't run status command if API key is missing.

ests:
Running `exercism status ruby` command when API is missing (f.i. I've deleted config file) will make actual HTTP request and cause PANIC: `%!v(PANIC=runtime error: invalid memory address or nil pointer dereference)`. I believe it should behave in the same way like `submit` command (prevent any requests and print message).


Tonkpils:
Great, thanks for this!

PR #344, opened 2016-08-21
Be a XDG-friendly tool

narqo:
Search for config in common paths (e.g. `~/.config/exercism.json`, ~/.exercism.json`).

In case `$XDG_CONFIG_HOME` is set, tests of paths package fail as config home was not mocked properly.

The patch simplifies paths' internals and introduces a more flexible way to mock globals of `paths` package.

### DIFF
--- a/paths/paths.go
+++ b/paths/paths.go
@@ -11,7 +11,7 @@ import (
 const (
 	// File is the default name of the JSON file where the config written.
 	// The user can pass an alternate filename when using the CLI.
-	File = ".exercism.json"
+	File = "exercism.json"
 	// DirExercises is the default name of the directory for active users.
 	// Make this non-exported when handlers.Login is deleted.
 	DirExercises = "exercism"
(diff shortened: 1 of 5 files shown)


kytrinyx:
Nice catch, thanks @narqo!

@Tonkpils @lcowell Would one of you give this a second look?

paths/paths.go
```diff
@@ -62,16 +63,17 @@ func Exercises(path string) string {
...
 
-// Recalculate sets exercism paths based on Home.
-func Recalculate() {
-	if Home == "" {
-		home, err := findHome()
-		if err != nil {
-			panic(err)
-		}
+// Recalculate sets exercism paths
+func Recalculate(home, configHome string) {
```
Tonkpils: It seems this functions whole purpose is to setup the `Home` and `ConfigHome` and only used within `init`. What about just dropping the function all together and setting the variables directly on the tests and doing this same logic from init?

lcowell: I agree, this function doesn't have a lot of value right now, especially given that we now have other inputs for doing the testing.  I think we could justify keeping it if we gave it a better name. Maybe `assignConfigHome` or `resolveConfigHome` (if we assign `paths.ConfigHome` outside the function). if we don't fold it in to `init`, I don't think it needs to be exported.

lcowell:
Tests pass. Looks great!

narqo:
Thank you for the feedback, I'll fix the PR anytime soon.

But, to be honest, I'm not very happy about the logic of the `paths` module, which I wanted to discuss.

To my taste, it feels unnatural, that `$XDG_CONFIG_HOME` is considered as an exclusive option for `ConfigHome` path. IMHO, the way "XDG-friendly" tool should operate, is to check if the config is within the possible directories. E.g., a tool should check

- `$XDG_CONFIG_HOME/exercism.json`
- `$HOME/.config/exercism.json`
- `$HOME/.exercism.json`

during the initialization process. At least, tools like git and others do it in this way.

Tonkpils:
@narqo we are always open for discussion and thanks for bringing up your concern. 

If I understand you correctly, you mean we should be checking for the existence of the `exercism.json` file under those paths based on priority rather than just assigning the `ConfigHome` altogether and assuming the file will be there?

That sounds reasonable to me, would you be open to making those changes? 

@lcowell @kytrinyx what are your thoughts on that?

narqo:
> That sounds reasonable to me, would you be open to making those changes?

Sure, it'd be nice brain training.

kytrinyx:
> That sounds reasonable to me

Yeah, same here.

narqo:
🆙 

I've updated the logic of `paths` and `config` modules. Since now on `path.Config()` returns a path based on `ConfigHome`, which is set to `$XDG_CONFIG_HOME` or `~/.config` if the former isn't set. If the path doesn't exist, `config` module uses `paths.DefaultConfig` as a value for `config.File`. 

I believe the changes match to the current behaviour of the CLI.

kytrinyx:
I believe the current behavior has three possibilities in this order of priority:

1. `EXERCISM_CONFIG_FILE`
1. `XDG_CONFIG_HOME`
1. `~/.exercism.json` (I think, though `~/.config/exercism.json` would be friendlier)

paths/paths_test.go
```diff
@@ -74,10 +82,3 @@ func TestConfig(t *testing.T) {
 		assert.Equal(t, tc.expectedPath, actual, tc.desc)
 	}
 }
-
-func TestXDGConfig(t *testing.T) {
```
narqo: This test is useless now, as the behaviour of the `paths.Config()` only depends on `ConfigHome`, which was tested above.

narqo:
> I believe the current behavior has three possibilities in this order of priority:

Yeah, thank you, I've forgotten to mention this. The way `github.com/urfave/cli` works , the `$EXERCISM_CONFIG_FILE` would always be the most preferable (and exclusive) option if it was set.

Tonkpils:
Thanks for the work on this @narqo, I'm going to go ahead and merge this and update the CHANGELOG!

kytrinyx:
Thank you @narqo!

narqo:
I'm totally glad I was helpful

PR #346, opened 2016-09-05
iteration: fall back to UTF-8 (not windows-1252) if encoding uncertain

### LINKED ISSUE #309: Files that have Unicode chars only after the 1024th byte get their Unicode mangled
The situation is that I had a file whose first 1024 bytes didn't contain a Unicode character, but then a character after the 1024th byte did. An example of such a file is:

http://exercism.io/submissions/1e341848768141cf8eba94c6af6e55a7

Submitting this file mangles the Unicode character.

In contrast, this file has Unicode in the first 1024 bytes, so it is good (even the Unicode that appears after the first 1024 bytes is good)

http://exercism.io/submissions/dce4e3ddf0294034ad987ce7b86cdb38

(These are just example submissions in Hello World, but this affected my submission for a real exercise too, Counter in xgo)

I tracked this down to `readFileAsUTF8String` in `api/iteration.go`. This uses the https://godoc.org/golang.org/x/net/html/charset#DetermineEncoding function to determine the encoding, which reads the first 1024 bytes.

I'm not really sure what's the right solution here. I know that function was created for https://github.com/exercism/cli/pull/182 to solve https://github.com/exercism/exercism.io/issues/2303 so there obviously is a legitimate reason behind all this, I guess maybe now we just need to figure out how to deal with this case as well. I don't yet have a good solution, so I'll file this first and sleep on it for a bit.

petertseng:
As discussed in #309: Since #184 we have been using DetermineEncoding to
deal with the case of UTF-16 files.
That was a reasonable fix for exercism/exercism.io#2303.

DetermineEncoding only looks at the first 1024 bytes of a file. If it
can't determine an encoding, it defaults to windows-1252.

This causes undesirable behaviour for files with Unicode characters but
also only ASCII in their first 1024 characters - they get interpreted as
windows-1252, mangling the Unicode characters.

This commit takes advantage of the fact that DetermineEncoding reports
whether it is *certain* about its encoding guess. If it is uncertain, we
default to UTF-8 instead of windows-1252.

Note that if DetermineEncoding sees UTF-16 BOMs, it will declare that it
is certain. Therefore, behaviour for UTF-16 files is preserved (existing
tests would have caught it if behaviour were accidentally altered).

A new fixture file is attached that tests this case - the test fails
without the attached code change.

I find it unlikely that DetermineEncoding would have returned anything
other than UTF-16, UTF-8, or windows-1252 since it was made to examine
HTML documents and thus examine the content-type (we always pass
text/plain) and the meta tags (unlikely to be present in a non-HTML
Exercism submission).

The risk of this change is that anyone who **actually** wanted to submit
in windows-1252 will now be unable to, but I doubt that anyone is in
this constituency, and discussion in #309 seems to be in favor of 
nudging them toward UTF-8 anyway.

Closes #309

### DIFF
--- a/api/iteration.go
+++ b/api/iteration.go
@@ -137,7 +137,16 @@ func readFileAsUTF8String(filename string) (*string, error) {
 		return nil, err
 	}
 
-	encoding, _, _ := charset.DetermineEncoding(b, mimeType)
+	encoding, _, certain := charset.DetermineEncoding(b, mimeType)
+	if !certain {
+		// We don't want to use an uncertain encoding.
+		// In particular, doing that may mangle UTF-8 files
+		// that have only ASCII in their first 1024 bytes.
+		// See https://github.com/exercism/cli/issues/309.
+		// So if we're unsure, use UTF-8 (no transformation).
+		s := string(b)
+		return &s, nil
+	}
 	decoder := encoding.NewDecoder()
 	decodedBytes, _, err := transform.Bytes(decoder, b)
 	if err != nil {
--- a/api/iteration_test.go
+++ b/api/iteration_test.go
@@ -18,6 +18,7 @@ func TestNewIteration(t *testing.T) {
 		filepath.Join(dir, "python", "leap", "lib", "three.py"),
 		filepath.Join(dir, "python", "leap", "utf16le.py"),
 		filepath.Join(dir, "python", "leap", "utf16be.py"),
+		filepath.Join(dir, "python", "leap", "long-utf8.py"),
 	}
 
 	iter, err := NewIteration(dir, files)
@@ -32,8 +33,8 @@ func TestNewIteration(t *testing.T) {
 		t.Errorf("Expected problem to be leap, was %s", iter.Problem)
 	}
 
-	if len(iter.Solution) != 5 {
-		t.Fatalf("Expected solution to have 5 files, had %d", len(iter.Solution))
+	if len(iter.Solution) != 6 {
+		t.Fatalf("Expected solution to have 6 files, had %d", len(iter.Solution))
 	}
 
 	expected := map[string]struct {
@@ -45,6 +46,7 @@ func TestNewIteration(t *testing.T) {
 		filepath.Join("lib", "three.py"): {prefix: "# three"},
 		"utf16le.py":                     {prefix: "# utf16le"},
 		"utf16be.py":                     {prefix: "# utf16be"},
+		"long-utf8.py":                   {prefix: "# The first 1024", suffix: "👍\n"},
 	}
 
 	for filename, code := range expected {
--- a/api/iteration_test.go
+++ b/api/iteration_test.go
@@ -33,7 +33,7 @@ func TestNewIteration(t *testing.T) {
 	}
 
 	if len(iter.Solution) != 5 {
-		t.Fatalf("Expected solution to have 3 files, had %d", len(iter.Solution))
+		t.Fatalf("Expected solution to have 5 files, had %d", len(iter.Solution))
 	}
 
 	expected := map[string]string{
--- a/api/iteration_test.go
+++ b/api/iteration_test.go
@@ -36,12 +36,15 @@ func TestNewIteration(t *testing.T) {
 		t.Fatalf("Expected solution to have 5 files, had %d", len(iter.Solution))
 	}
 
-	expected := map[string]string{
-		"one.py": "# one",
-		"two.py": "# two",
-		filepath.Join("lib", "three.py"): "# three",
-		"utf16le.py":                     "# utf16le",
-		"utf16be.py":                     "# utf16be",
+	expected := map[string]struct {
+		prefix string
+		suffix string
+	}{
+		"one.py": {prefix: "# one"},
+		"two.py": {prefix: "# two"},
+		filepath.Join("lib", "three.py"): {prefix: "# three"},
+		"utf16le.py":                     {prefix: "# utf16le"},
+		"utf16be.py":                     {prefix: "# utf16be"},
 	}
 
 	for filename, code := range expected {
@@ -49,8 +52,11 @@ func TestNewIteration(t *testing.T) {
 			t.Errorf("Iteration content is not valid UTF-8 data: %s", iter.Solution[filename])
 		}
 
-		if !strings.HasPrefix(iter.Solution[filename], code) {
-			t.Errorf("Expected %s to contain `%s', had `%s'", filename, code, iter.Solution[filename])
+		if !strings.HasPrefix(iter.Solution[filename], code.prefix) {
+			t.Errorf("Expected %s to start with `%s', had `%s'", filename, code.prefix, iter.Solution[filename])
+		}
+		if !strings.HasSuffix(iter.Solution[filename], code.suffix) {
+			t.Errorf("Expected %s to end with `%s', had `%s'", filename, code.suffix, iter.Solution[filename])
 		}
 	}
 }
--- a/fixtures/iteration/python/leap/long-utf8.py
+++ b/fixtures/iteration/python/leap/long-utf8.py
@@ -1,0 +1,31 @@
+# The first 1024 bytes of this file need to contain only ASCII characters.
+# After the first 1024 bytes, then there should be a non-ASCII character.
+#
+# Explanation:
+# We use golang.org/x/net/html/charset.DetectEncoding to guess file encoding.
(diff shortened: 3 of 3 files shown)


api/iteration_test.go
```diff
@@ -32,16 +34,24 @@ func TestNewIteration(t *testing.T) {
...
 
-	if len(iter.Solution) != 5 {
-		t.Fatalf("Expected solution to have 3 files, had %d", len(iter.Solution))
+	if len(iter.Solution) != 6 {
+		t.Fatalf("Expected solution to have 6 files, had %d", len(iter.Solution))
 	}
 
+	var buf bytes.Buffer
+	buf.WriteString("# ")
+	for i := 0; i < 1024; i++ {
```
petertseng: nah, let's use `strings.Repeat` instead.

petertseng: done

fixtures/iteration/python/leap/long-utf8.py
```diff
@@ -0,0 +1,2 @@
+# aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
petertseng: This makes me a bit uncomfortable.

It's important for this test that there be 1024 bytes with only ASCII, followed by non-ASCII (in this case the thumb up).

This is not at all apparent from reading this file. You have to look at the commit message. I would like to make it more apparent from reading the file.

That however complicates the test process. The tests only support exact string matches. Let's say I put explanatory text inside this file. But then the test would have to expect that same explanatory text.

But I might argue that the test is expecting too much. All I care about for this file is that the thumbs up survives intact.

So I propose that I do the following:

* Instead of a single linne 1024 a's, put an explanatory comment in this file explaining the required characteristics (first 1024 bytes ASCII, followed by non-ascii), and some filler text (I'll grab some lorem ipsum).
* In the test file, support arbitrary functions for file validation in addition to exact string matching.
* In the test file, just expect that the thumbs up survives intact in long-utf8.py (the other files will remain on exact string matching because no reason not to)

petertseng: I have done the proposed.

Note that I was inaccurate - the existing tests look for prefixes. So I just also allowed suffixes too.

api/iteration.go
```diff
@@ -137,7 +137,16 @@ func readFileAsUTF8String(filename string) (*string, error) {
 		return nil, err
 	}
 
-	encoding, _, _ := charset.DetermineEncoding(b, mimeType)
+	encoding, _, certain := charset.DetermineEncoding(b, mimeType)
+	if !certain {
```
petertseng: The consequence of replacing this line with `if true`:

```
--- FAIL: TestNewIteration (0.00s)
	iteration_test.go:54: Iteration content is not valid UTF-8 data: ÿþ# utf16le
	iteration_test.go:58: Expected utf16le.py to start with `# utf16le', had `ÿþ# utf16le'
	iteration_test.go:54: Iteration content is not valid UTF-8 data: þÿ# utf16be
	iteration_test.go:58: Expected utf16be.py to start with `# utf16be', had `þÿ# utf16be'
FAIL
FAIL	github.com/exercism/cli/api	0.019s
```

And the consequences of replacing with `if false` (as if I never wrote this code):

```
--- FAIL: TestNewIteration (0.00s)
	iteration_test.go:61: Expected long-utf8.py to end with `👍
		', had `# The first 1024 bytes of this file need to contain only ASCII characters.
		# After the first 1024 bytes, then there should be a non-ASCII character.
		# Explanation: DetectEncoding checks the first 1024 bytes of a file.
		# If it can't determine the encoding and saw no non-ASCII characters,
		# it declares the file to have windows-1252 encoding.
		# This mangles the submitted file if it should have been UTF-8.
		# We test to make sure we use UTF-8 for such files, instead of windows-1252.
		
		lipsum = """
		Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam condimentum vitae
		ipsum eget tempor. Morbi sed ex quis orci vulputate cursus quis non massa.
		Vestibulum quam nibh, elementum in justo in, venenatis tristique nisl. Morbi
		sagittis elit id velit ultricies, sed rutrum augue posuere. Donec nec nulla nec
		eros fringilla pellentesque. Duis at dictum justo. Nunc ut magna felis. Aliquam
		volutpat, lectus et molestie porttitor, est orci malesuada erat, ac pretium
		eros ligula vel erat. Nullam venenatis dui eget sapien semper lobortis. Aenean
		ac eros eget neque porta auctor in nec erat. Phasellus ac nulla ac turpis
		porttitor auctor. Etiam eget posuere diam, ac feugiat lacus. Curabitur ornare
		justo ut nulla congue, vitae posuere erat venenatis. Aliquam pulvinar eleifend
		faucibus.
		
		Etiam justo sem, faucibus malesuada purus a, ultrices efficitur ex.
		Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac
		turpis egestas. Duis maximus dapibus mattis. Quisque sem ex, convallis eu
		ultricies posuere.
		"""
		
		# ðŸ‘�
		'
FAIL
FAIL	github.com/exercism/cli/api	0.021s
```

fixtures/iteration/python/leap/long-utf8.py
```diff
@@ -0,0 +1,28 @@
+# The first 1024 bytes of this file need to contain only ASCII characters.
```
petertseng: Of two minds. On the one hand, this is a test that knows a little too much about the implementation of the code it's testing (1024 bytes).

On the other hand, this is a real thing that happened and mangled my submission and someone else's, so it is not unwarranted.

Tonkpils:
These changes seem ok to me though I'd be more comfortable if @lcowell or @kytrinyx double checked on these. The test makes me comfortable enough to merge this though I'd feel more confident with a second pair of eyes on this.

lcowell:
Good explanation of the changes and the test coverage looks good too. Code looks good. The tests pass on my OSX(10.12) machine. Thanks @petertseng for this PR.

kytrinyx:
I'd say let's go for it.

kytrinyx:
Going for it :)

Issue #340, opened 2016-07-29
improper slug on case-insensitive filesystems


hynkle:
On case-insensitive filesystems (e.g. OS X's default), it's possible for an incorrect slug for the current exercise to be gotten from the current working directory.

Suppose an exercise exists in `~/exercism/ruby/pangram`, but you got into that directory by doing `cd ~/exercism/Ruby/pangram`.  This causes the capitalized `Ruby` in the path to get used as the slug here in NewIteration: https://github.com/exercism/cli/blob/master/api/iteration.go#L106  I discovered this upon doing `exercism submit pangram.rb`, which caused an `unable to submit (HTTP: 400) - unknown problem (track: Ruby, slug: pangram, path: )` error.

(If you're wondering how anyone might have ever done this on accident:  My boyfriend's exercises path is ~/Documents/Code/exercism, and the capital D and C made him accidentally capitalize the other directory names as well while he was `cd`ing.)

I'm not super familiar with the exercism ecosystem, but if it's safe to assume that slugs will always be lowercase, a possible fix for this could be to use `ToLower` on the line linked above.

kytrinyx:
> possible fix for this could be to use ToLower on the line linked above.

Yepp—that's a safe assumption and a good solution. Thanks!

Tonkpils:
Are there currently or will there be any slugs with other casing as well? If not, we should ToLower on the slug as well.

PR #352, opened 2016-10-25
Fix case sensitivity on slug and track ID iteration.

Tonkpils:
fixes https://github.com/exercism/cli/issues/340


kytrinyx:
:heart: thank you!

PR #360, opened 2016-11-25
Gofmt for codegangsta/cli -> urfave/cli change

ferhatelmas:
Probably it should be done while rename was being done but here it is.


kytrinyx:
Ahhh, yepp. Thanks for catching that.

Issue #341, opened 2016-07-29
Feature request:  'download'  should accept full Exercism URLs


Insti:
Feature request: The download option should accept full Exercism URLs

`exercism download http://exercism.io/submissions/45aaa580ea91452bb34076ade70511f2`

Currently it's necessary to copy/paste just the uuid hex digits from the full URL.

kytrinyx:
I like this suggestion.

Insti:
Should this change happen in the CLI or the API?

kytrinyx:
Trade-offs:

In the CLI feels cleaner. In the API means we don't need to release a new CLI.

kotp:
Alternatively, having the digits shown with a command like `exercism dl 45aaa580ea91452bb34076ade70511f2` for each submission page means that you can copy and paste to your terminal/command line.

Insti:
I'd like there to be a button that just let me download the file directly from the webpage.

kotp:
Wouldn't that require registering an "application type" to be able to use the cli in that way?  Send a message to it while the exercise is being requested.

kytrinyx:
> Wouldn't that require registering an "application type" to be able to use the cli in that way?

We wouldn't have to use the CLI to download—we could zip up the exercise+solution and it would download to the default location that the browser uses for downloads.

kotp:
Yes, but would it be able to put it in the location as I am used to it being now?  I think this is based on the local config... Of course I could just choose not to use the button... I like how it works at the moment, though having the hexadecinal number easily placed on my clipboard would be nice. (It works well in my terminal based browser, so really have no strong feelings on this personally.)

Insti:
Just wanted to clarify, my "button to download directly from the webpage" would be a button on the solution page on the website implemented in the exercism.io repository and NOT as an addition to the cli.

kytrinyx:
> Yes, but would it be able to put it in the location as I am used to it being now? 

No, I think that would be fine though.

> having the hexadecinal number easily placed on my clipboard would be nice.

Yeah, having the full command that could be copied to the clipboard with a click would be really nice.

blueelvis:
We can have application specific URLs (for example - exercism://fetch ) and it will open via Exercism. That's all good and easy as well. Read this - https://msdn.microsoft.com/en-us/library/aa767914(v=vs.85).aspx


But, there could be security issues as well. Read this over here - http://stackoverflow.com/a/389271

I also think that having the command being copied to the clipboard would be best.

[...]

PR #364, opened 2017-01-02
'download' should accept full Exercism URLs

verg:
Enhancement from: https://github.com/exercism/cli/issues/341

Allows `exercism download http://exercism.io/submissions/45aaa580ea91452bb34076ade70511f2` and `exercism download exercism.io/submissions/45aaa580ea91452bb34076ade70511f2`

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/exercism/cli/api"
@@ -27,7 +28,8 @@ func Download(ctx *cli.Context) error {
 		os.Exit(1)
 	}
 
-	submission, err := client.Download(args[0])
+	submissionKey := submissionKeyFromURL(args[0])
+	submission, err := client.Download(submissionKey)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -65,4 +67,9 @@ func writeFile(path, contents string) error {
 		return err
 	}
 	return ioutil.WriteFile(path, []byte(contents), 0644)
+}
+
+func submissionKeyFromURL(submission string) string {
+	re := regexp.MustCompile(".*(https*://)?exercism.io/submissions/")
+	return re.ReplaceAllString(submission, "")
 }
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -70,6 +70,6 @@ func writeFile(path, contents string) error {
 }
 
 func submissionKeyFromURL(submission string) string {
-	re := regexp.MustCompile(".*(https*://)?exercism.io/submissions/")
+	re := regexp.MustCompile(".*(https?://)?exercism.io/submissions/")
 	return re.ReplaceAllString(submission, "")
 }


cmd/download.go
```diff
@@ -13,6 +13,11 @@ import (
 	"github.com/urfave/cli"
 )
 
+const (
+	submission_url_prefix        = "exercism.io/submissions/"
+	submission_url_prefix_w_http = "http://" + submission_url_prefix
```
kytrinyx: Would it be possible to use a precompiled regex here that optionally includes http:// (and, potentially, https://)?

cmd/download.go
```diff
@@ -27,7 +32,8 @@ func Download(ctx *cli.Context) error {
 		os.Exit(1)
 	}
 
-	submission, err := client.Download(args[0])
+	submissionID := parseSubmissionID(args[0])
```
kytrinyx: We've got a bit of a mess in how we name things, but I think that we've been calling this the submission key rather than the submission ID.

cmd/download.go
```diff
@@ -66,3 +72,10 @@ func writeFile(path, contents string) error {
 	}
 	return ioutil.WriteFile(path, []byte(contents), 0644)
 }
+
+// Parses Submission ID from url strings
+// e.g "http://exercism.io/submissions/<SUBMISSION_ID>" --> <SUBMISSION_ID>
+func parseSubmissionID(submission string) string {
```
kytrinyx: How about `submissionKeyFromURL` as a name for this? parseSubmissionID sounds like it's parsing the ID, which it's not. A name starting in _parse_ tends to focus on what the function is doing rather than what you get out of it, which can be harder to understand.

The function doesn't need to be commented, since it's not exported.

verg:
Hey Katrina,

Thanks for the awesome feedback!

How's this look? Okay to slurp everything at the beginning with the ".*"?


Some testing:
```go
gore> re := regexp.MustCompile(".*(https*://)?exercism.io/submissions/")
&regexp.Regexp{regexpRO:regexp.regexpRO{expr:".*(https*://)?exercism.io/submissions/", prog:(*syntax.Prog)(0xc4200664b0), onepass:(*regexp.onePassProg)(nil), prefix:"", prefixBytes:[]uint8(nil), prefixComplete:false, prefixRune:0, prefixEnd:0x0, cond:0x0, numSubexp:1, subexpNames:[]string{"", ""}, longest:false}, mu:sync.Mutex{state:0, sema:0x0}, machine:[]*regexp.machine(nil)}
gore> re.ReplaceAllString("http://exercism.io/submissions/fbfb626b2b534e18bfc7d04e1363fce8", "")
"fbfb626b2b534e18bfc7d04e1363fce8"
gore> re.ReplaceAllString("https://exercism.io/submissions/fbfb626b2b534e18bfc7d04e1363fce8", "")
"fbfb626b2b534e18bfc7d04e1363fce8"
gore> re.ReplaceAllString("exercism.io/submissions/fbfb626b2b534e18bfc7d04e1363fce8", "")
"fbfb626b2b534e18bfc7d04e1363fce8"
gore> re.ReplaceAllString("wasdlkexercism.io/submissions/fbfb626b2b534e18bfc7d04e1363fce8", "")
"fbfb626b2b534e18bfc7d04e1363fce8"
gore> re.ReplaceAllString("fbfb626b2b534e18bfc7d04e1363fce8", "")
"fbfb626b2b534e18bfc7d04e1363fce8"
```

NobbZ:
> `".*(https*://)?exercism.io/submissions/"`

The complexity of the resulting automaton is nearly the same, but isn't `".*(https?://)?exercism.io/submissions/"` more correct? Also since we throw the away the match anyway, `".*(?https*://)?exercism.io/submissions/"` could reduce overhead, since there is no additional matching/backtracking group to remember. Dunno if go-re does know about that particular part of syntax.

verg:
@NobbZ Good call. I updated `(https*://)?` to `(https?://)?`.

I'm not very familiar, but Go doesn't appear to support `'(?'`. I get this panic:
```go
panic: regexp: Compile(`.*(?https?://)?exercism.io/submissions/`): error parsing regexp: invalid or unsupported Perl syntax: `(?h`
```
Updated tests: https://play.golang.org/p/oQhgB79AOD

kotp:
Will this support both forms?  The "last part of the URL" and the full URL?

I just tried to download a submission with the (reported) cli being up to date, but it is failing.

kytrinyx:
It sounds like it would be worth having a test for the regex, since it's been a bit tricky to get it working.

nywilken:
> Also since we throw the away the match anyway, ".*(?https?://)?exercism.io/submissions/" could reduce overhead, since there is no additional matching/backtracking group to remember. Dunno if go-re does know about that particular part of syntax.

@NobbZ I believe you are advising that the a non-capturing group be used for matching the http string.  If so @verg the syntax you are looking for is `".*(?:https?://)?exercism.io/submissions/"`. Go uses the Re2 syntax which differs from other RegExp engines. More information on Re2 can be found on [Google's Re2 Wiki](https://github.com/google/re2/wiki/Syntax)

NobbZ:
Basically there where 2 proposals in that overall comment:

* Change quantifier `*` after the `s` of `https` to `?` because the former would allow urls like `httpssssss://…`.
* Change the match-group to a non-match group to reduce complexity of the resulting DFA. This is a premature optimisation though. It may be, that it doesn't actually change anything in the internal representation of the compiled regular expression. This is subject to benchmarks. Those benchmarks should measure compile time of both expressions as well as execution times against a couple of strings.

kytrinyx:
It's been over a year since we heard from the original author on this one. I'm going to go ahead and close the PR.

PR #366, opened 2017-01-22
Add silent argument/option to the configure command to obfuscate output

dmmulroy:
Give users an option obfuscate/hide the output from the configure command.
My personal use case: Running exercism on Travis CI and my API key being publicly view-able in the output/logfile.


kytrinyx:
I really like this. Thanks @dmmulroy!

PR #378, opened 2017-03-19
Replace Client with Interface

rootulp:
- SVG for badge instead of PNG
- Remove style=flat for 24pullrequests badge
- Move badges below title for consistency with other repos


kytrinyx:
Thanks, I appreciate it.

PR #379, opened 2017-03-25
Fix minor formatting issue in RELEASE doc

kytrinyx:
There was an unmatched underscore in the markdown for something which was intended to be italic.


kytrinyx:
I'm going to go ahead and merge this.

PR #398, opened 2017-06-30
Support the two most recent releases of Go

kytrinyx:
1.8 was released recently, let's remove support for 1.6 and add 1.8


Tonkpils:
👍 Ah, thanks for reminding me to read up on changes in Go 1.8 🎉

PR #405, opened 2017-07-23
[WIP|ready 4 review] Selective restore

### LINKED ISSUE #389: Respect TRACK_ID and SLUG options on restore command
The restore command currently has one behavior: restore absolutely everything.

This is not going to work when people have done lots of exercises in lots of tracks.

A better behavior would be to respect passed in arguments for track and exercise:

```
exercism restore javascript
```

and

```
exercism restore rust leap
```

If you have not done the exercise, the API will return a 404 and an error message that we can display.

If someone says `exercism restore` with no arguments, I think we should return an error message and ask them to specify which track to restore.

NobbZ:
This should work already, but I wasn't able to set up a VM containing a pristine environment in which I were able to check its behaviour.

I do hope I have such an environment ready in at least 72 hours and can report back with my experiences.

Until then, feel free to review and comment on the first small thing you see, that I can fix it before it gets a real problem ;)

Will FIX #389

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -93,8 +93,8 @@ func (c *Client) FetchAll(trackID string) ([]*Problem, error) {
 	return problems, nil
 }
 
-// Restore fetches the latest revision of a solution and writes it to disk.
-func (c *Client) Restore() ([]*Problem, error) {
+// RestoreAll fetches the latest revisions of all solutions and writes them to disk.
+func (c *Client) RestoreAll() ([]*Problem, error) {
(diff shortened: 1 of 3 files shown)


NobbZ:
Sorry for not beeing able to take better care of this. But this one is hacked in a hurry and @kytrinyx seems to have done similar in #407, I'll close this one in favor of the other one.

kytrinyx:
Ahhh, I'm sorry I missed this! It's been a chaotic few weeks in my notifications and it slipped through the cracks :(

kytrinyx:
That makes sense. Thanks @NobbZ 👍

PR #407, opened 2017-07-31
Refactor upgrade command

kytrinyx:
I've moved all the upgrade logic into its own package. I added a couple of tests for the "is upgrade needed" logic.

In order to create a separate package named `cli`, I had to alias the urlfave/cli package. I aliased it to `app`.

I tested both the debug command and the upgrade command manually since we don't have test coverage for the commands.

```
# edit the version in exercism/main.go to be lower
go build -o testercism exercism/main.go && ./testercism debug
go build -o testercism exercism/main.go && ./testercism upgrade
```

### DIFF
--- a/cli/asset.go
+++ b/cli/asset.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 )
 
+// Asset is a build for a particular system, uploaded to a GitHub release.
 type Asset struct {
 	ID          int    `json:"id"`
 	Name        string `json:"name"`
--- a/cli/asset.go
+++ b/cli/asset.go
@@ -1,20 +1,19 @@
-package cmd
+package cli
 
 import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 )
 
-type asset struct {
+type Asset struct {
 	ID          int    `json:"id"`
 	Name        string `json:"name"`
 	ContentType string `json:"content_type"`
 }
 
-func (a *asset) download() (*bytes.Reader, error) {
+func (a *Asset) download() (*bytes.Reader, error) {
 	downloadURL := fmt.Sprintf("https://api.github.com/repos/exercism/cli/releases/assets/%d", a.ID)
 	req, err := http.NewRequest("GET", downloadURL, nil)
 	if err != nil {
@@ -34,14 +33,4 @@ func (a *asset) download() (*bytes.Reader, error) {
 	}
 
 	return bytes.NewReader(bs), nil
-}
-
-type release struct {
-	Location string  `json:"html_url"`
-	TagName  string  `json:"tag_name"`
-	Assets   []asset `json:"assets"`
-}
-
-func (r *release) Version() string {
-	return strings.TrimPrefix(r.TagName, "v")
 }
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,176 @@
+package cli
+
+import (
+	"archive/tar"
+	"archive/zip"
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/blang/semver"
+	"github.com/exercism/cli/debug"
+	update "github.com/inconshreveable/go-update"
+)
+
+var (
+	// BuildOS is the operating system (GOOS) used during the build process.
+	BuildOS string
+	// BuildARM is the ARM version (GOARM) used during the build process.
+	BuildARM string
+	// BuildARCH is the architecture (GOARCH) used during the build process.
+	BuildARCH string
+)
+
+var (
+	osMap = map[string]string{
+		"darwin":  "mac",
+		"linux":   "linux",
+		"windows": "windows",
+	}
+
+	archMap = map[string]string{
+		"amd64": "64bit",
+		"386":   "32bit",
+		"arm":   "arm",
+	}
+)
+
+var (
+	HTTPClient       = &http.Client{Timeout: 10 * time.Second}
+	LatestReleaseURL = "https://api.github.com/repos/exercism/cli/releases/latest"
+)
+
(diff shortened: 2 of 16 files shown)


cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+
+	archMap = map[string]string{
+		"amd64": "64bit",
+		"386":   "32bit",
+		"arm":   "arm",
+	}
+)
+
+var (
+	HTTPClient       = &http.Client{Timeout: 10 * time.Second}
```
kytrinyx: I made this a package variable so that debug can set its own client on the package.

cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+	LatestReleaseURL = "https://api.github.com/repos/exercism/cli/releases/latest"
```
kytrinyx: I made this a package variable so that we can override it in the test.

cli/asset.go
```diff
@@ -1,20 +1,19 @@
...
 import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 )
 
-type asset struct {
+type Asset struct {
```
kytrinyx: I moved asset into its own file.

cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+	LatestRelease *Release
+}
+
+func New(version string) *CLI {
+	return &CLI{
+		Version: version,
+	}
+}
+
+func (c *CLI) IsUpgradeNeeded() (bool, error) {
```
kytrinyx: I got rid of the `upgrader` type, and now the `CLI` type can check whether or not it needs an upgrade.

cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+	if c.LatestRelease == nil {
```
kytrinyx: The tests can check the upgrade logic without http calls by setting a `LatestRelease` on the `CLI`.

cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+	cv, err := semver.Make(c.Version)
+	if err != nil {
+		return false, fmt.Errorf("unable to parse current version (%s): %s", c.Version, err)
+	}
+
+	return rv.GT(cv), nil
+}
+
+// Upgrade allows the user to upgrade to the latest version of the CLI.
+func (c *CLI) Upgrade() error {
```
kytrinyx: I moved the `Upgrade()` function directly onto the `CLI` type. The logic is the same as it was, but I moved some of the `fmt.Println`s to keep them in the `cmd` package.

cli/cli.go
```diff
@@ -0,0 +1,176 @@
...
+	bin, err := extractBinary(downloadRC, OS)
+	if err != nil {
+		return err
+	}
+	defer bin.Close()
+
+	return update.Apply(bin, update.Options{})
+}
+
+func extractBinary(source *bytes.Reader, os string) (binary io.ReadCloser, err error) {
```
kytrinyx: Since `extractBinary` doesn't need any data from the `CLI` I changed it to be a function rather than a method. Other than removing the function receiver, I haven't touched the logic in `extractBinary`.

cli/release.go
```diff
@@ -0,0 +1,13 @@
+package cli
+
+import "strings"
+
+type Release struct {
```
kytrinyx: I moved `Release` into its own file.

cmd/configure.go
```diff
@@ -6,13 +6,13 @@ import (
...
 	"github.com/exercism/cli/config"
-	"github.com/urfave/cli"
+	app "github.com/urfave/cli"
 )
 
 // Configure stores settings in a JSON file.
 // If a setting is not passed as an argument, default
 // values are used.
-func Configure(ctx *cli.Context) error {
+func Configure(ctx *app.Context) error {
```
kytrinyx: This is the aliasing that I mentioned, to make room for the new `cli` package.

cli/cli.go
Tonkpils: What do you think about making it part of the `CLI` struct instead since it seems it'll only be used in that context anyway. We could also allow passing it through in the `New` function and make it a private on the struct, I feel it would make it a more explicit on what the struct is using for the underlying client.

kytrinyx: I was thinking about that and kept going back and forth. I think after sleeping on it, I'm fine with the CLI having the client, but I really don't like passing `nil` to functions as the default (it's a personal thing. I know it's not wrong, but I always find myself having to jump to the method definition to see what the nil is about when I come across nils being passed in the codebase).

So maybe a `SetHTTPClient` on the CLI, or even just an exported field on the struct.

kytrinyx:
I've fixed the golint complaints.

@Tonkpils I could go either way on the HTTPClient, curious to hear what you're thinking about that one now.

cli/cli.go
```diff
@@ -0,0 +1,180 @@
...
+
+// New creates a CLI, setting it to a particular version.
+func New(version string) *CLI {
+	return &CLI{
+		Version: version,
+	}
+}
+
+// IsUpgradeNeeded compares the current version to that of the latest release.
+func (c *CLI) IsUpgradeNeeded() (bool, error) {
```
Tonkpils: That'd make sense, I agree the name could be more idiomatic and using `ok` feels a bit better. `IsUpToDate` would be my choice if you asked me 😄

Tonkpils:
> @Tonkpils I could go either way on the HTTPClient, curious to hear what you're thinking about that one now.

Me too, let's leave as is and we can always change it if it becomes odd to deal with down the road.

cli/cli.go
kytrinyx: I like `IsUpToDate` as well. IsCurrent is a bit ambiguous in this context.

kytrinyx:
> let's leave as is and we can always change it if it becomes odd to deal with down the road.

@Tonkpils agreed! OK, cool. I'll make the IsUpToDate change and then merge this. I have another refactoring waiting in the wings.

PR #420, opened 2017-08-05
Nextercism deps

Tonkpils:
If there's a time to do this, now is the time...

This switches us over to `dep` on the nextercism branch.  I had issues trying to get things to work smoothly with the newest version of Godep and I've pretty much ripped out the deps we had with Godep and locked them in with `dep`. Dep ended up adding extra missing dependencies that Godep was not adding and now tests are passing on CI!

One thing I noticed is that the `vendor` folder changeset is much bigger than when we were using Godep, not sure why that is yet but that seems to be a con of committing the vendor folder. I think that with `dep` we could get away with not doing this since it contains a lockfile but there's still caveats https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory


kytrinyx:
I'd love to try this with just the lock file and see if we're OK with it. Thoughts?

Tonkpils:
There may be downsides to that but we can try it and see how it works. If
we run into any issues we could always fix them and move to vendoring the
deps at that point. I'm not near my computer, if you'd like to add a commit
removing the vendor folder and adding to gitignore or I could do it once
I'm back on my computer!

kytrinyx:
> we can try it and see how it works. If
> we run into any issues we could always fix them and move to vendoring the
> deps at that point.

I think that sounds like a plan. I'm going to try it!

kytrinyx:
That worked!

If you like this feel free to squash (or not) and merge into the nextercism branch.

Tonkpils:
We can squash later if needed! 👍 thanks for tackling the removal of vendor and documentation!

robphoenix:
Awesome!! :tada: :tada:

Issue #417, opened 2017-08-04
nextercism: implement configure command


kytrinyx:
_To be merged into the nextercism branch in https://github.com/exercism/cli/pull/410_

Command name: `configure` (alias `c`).
`cmd/configure.go`.

This command accepts flags and writes the values into the user's configuration file.

To start, let's just implement `token` (the API token) and `workspace` (the directory that `download` will download exercises and solutions to).

kytrinyx:
Closed by #424

PR #421, opened 2017-08-05
[nextercism] Handle config directories that don't exist

kytrinyx:
While the normal config home, `~/.config` typically exists, we're
defining the app's config home to be a directory 'exercism' within
that directory, and that will almost certainly not exist.

Came across this problem as I was working on https://github.com/exercism/cli/issues/417

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -46,5 +46,17 @@ func Write(f filer) error {
 	if err != nil {
 		return err
 	}
+	if err := ensureDir(f); err != nil {
+		return err
+	}
 	return ioutil.WriteFile(f.File(), b, os.FileMode(0644))
+}
+
+func ensureDir(f filer) error {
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -46,5 +46,19 @@ func Write(f filer) error {
...
 	}
+	if err := ensureDir(f); err != nil {
+		return err
+	}
 	return ioutil.WriteFile(f.File(), b, os.FileMode(0644))
 }
+
+func ensureDir(f filer) error {
+	dir := filepath.Dir(f.File())
+	if _, err := os.Stat(dir); err == nil || !os.IsNotExist(err) {
```
kytrinyx: Nice! I like that second one.

kytrinyx:
Rebased onto master. Will merge when it goes green.

PR #423, opened 2017-08-06
[nextercism] Normalize the user config before writing it

kytrinyx:
Set home on the config in the normalize method rather than always setting it.

As I was troubleshooting this I discovered a silly bug where I was returning an empty string for the home directory if I found one.

This is related to the work I'm doing for https://github.com/exercism/cli/issues/417


kytrinyx:
Thanks @robphoenix. In terms of timeline—should I go ahead and merge this and you can beef up the tests in separate PRs later?

robphoenix:
Yeah, sounds good @kytrinyx :+1:

Issue #419, opened 2017-08-04
nextercism: implement submit command


kytrinyx:
_To be merged into the nextercism branch in https://github.com/exercism/cli/pull/410_

Command name: `submit` (alias `s`).
`cmd/submit.go`.

```
$ exercism submit
$ exercism submit path/to/directory
$ exercism submit file1 file2 [...]
```

Calling `submit` with no arguments will try to submit the current directory. If the current directory is not an identifiable exercise, we can either ask interactively what track (if multiple) and what exercise. Or we can error out and ask them to specify the path to the directory to submit.

The files are submitted as a multipart form upload as a `PATCH` request to the `/solutions/:uuid` endpoint.

The form field for the files is called `files[]`.

```
--54e5745cb2d1f16f85ecbdb4115271a6b58a4ecb3a98eeb594f259bd98f6
Content-Disposition: form-data; name="files[]"; filename="/file-1.txt"
Content-Type: application/octet-stream

This is file 1.

--54e5745cb2d1f16f85ecbdb4115271a6b58a4ecb3a98eeb594f259bd98f6
Content-Disposition: form-data; name="files[]"; filename="/subdir/file-2.txt"
Content-Type: application/octet-stream

This is file 2.

--54e5745cb2d1f16f85ecbdb4115271a6b58a4ecb3a98eeb594f259bd98f6--
```

Given the following directory structure:

```
.
├── file-1.txt
└── subdir
    └── file-2.txt
```

The contents of the files are:

```
# file-1.txt
This is file 1.

# file-2.txt
This is file 2.
```

And the following sinatra app:

```
# app.rb
require 'sinatra'

post '/upload' do
  request.body.rewind
  request.body.read
end
```

Run the app with:

```
ruby app.rb
```

Then use this curl command to generate a multipart post:

```
curl -v -include --request PATCH -H "Authorization: Bearer a37dee7f-8be9-49ea-9dda-df703784607f" --form files[]=@file-1.txt --form files[]=@subdir/file-2.txt "http://localhost:4567/upload"
```

Note that curl strips off the subdirectory and only sends the file's basename, which is wrong.

kytrinyx:
I've got this one.

nikclayton:
Hi -- why was this closed?

I'd like an "exercism submit" feature that just submits the current directory without needing to specify the files, and chasing through the GitHub issues brought me to the "Nextercism" branch, and then this issue.

[...]

Issue #416, opened 2017-08-04
nextercism: implement open command


kytrinyx:
_To be merged into the nextercism branch in https://github.com/exercism/cli/pull/410_

Command name: `open` (alias `o`).
`cmd/open.go`.

## Open solution by name

```
$ exercism open clock
```

Given an exercise name, it will locate it in the tree under the user's workspace. If there are `clock` exercises (e.g. in multiple tracks), then the command will ask which one to open.

The command uses the functionality in the `browser` package to open the URL.

There is a hidden `.solution` file in the exercise directory, which contains a UUID. This UUID can be used to determine the website URL to open.

## Open solution by path

Given a path to an exercise, it won't search for other exercises with the same name. This will work to open solutions by other people as well.

**UPDATE** The API will provide a solution url when we download, which we'll store in the metadata file for the solution. Then the open command can pass this URL directly to the browser package.

kytrinyx:
Fixed in #429

PR #425, opened 2017-08-06
[nextercism] Find exercises by path or name

kytrinyx:
This lets us pass absolute paths, relative paths, or an exercise name to a command
and locate all the potential matches within the user's workspace.

The open command (#416) and the submit command (#419) both need this behavior.

This also handles workspaces that are a symlink to another directory, as well as exercises that are symlinks. (/cc @wobh).

Most of the files in this pull request are the addition of fixtures supporting the tests for this behavior. I've put all the fixtures for this set of tests inside the `fixtures/locate-exercise` directory. Speaking from painful experience, other tests should not reuse these fixtures. I've found that shared fixtures become a mess. Also, I prefer not to mimic real-world data, but construct a fixture that has the features that I need to test. I try to keep the fixtures minimal, but I would rather have different fixtures to test different things, than to conflate a multiple concepts in a single set of fixtures. E.g. `bat` vs `batten` in one test, and `crane` vs `crane-2` in another, instead of saying `bat`, `bat-2`, and `batten`.

```
fixtures/locate-exercise/
├── equipment
│   └── bat
│       └── .keep
├── food
│   └── squash
│       └── .keep
├── symlinked-workspace -> workspace
└── workspace
    ├── actions
    │   ├── batten
    │   │   └── .keep
    │   ├── date
    │   └── squash
    │       └── .keep
    ├── creatures
    │   ├── bat
    │   │   └── .keep
    │   ├── crane
    │   │   └── .keep
    │   ├── crane-2
    │   │   └── .keep
    │   ├── duck
    │   │   └── .keep
    │   └── horse
    │       └── .keep
    ├── equipment
    ├── food
    │   ├── date -> ../text-files/date
    │   └── squash -> ../../food/squash
    ├── friends
    │   └── alice
    │       └── creatures
    │           ├── bat
    │           │   └── .keep
    │           └── fly
    │               └── .keep
    └── text-files
        ├── date
        └── duck
```

### DIFF
--- a/workspace/errors.go
+++ b/workspace/errors.go
@@ -1,0 +1,29 @@
+package workspace
+
+import "fmt"
+
+// ErrNotInWorkspace signals that the target directory is outside the configured workspace.
+type ErrNotInWorkspace string
+
+// ErrNotExist signals that the target directory could not be located.
+type ErrNotExist string
+
+func (err ErrNotInWorkspace) Error() string {
+	return fmt.Sprintf("%s not within workspace", err)
+}
+
+func (err ErrNotExist) Error() string {
+	return fmt.Sprintf("%s not found", err)
+}
+
+// IsNotInWorkspace checks if this is an ErrNotInWorkspace error.
+func IsNotInWorkspace(err error) bool {
+	_, ok := err.(ErrNotInWorkspace)
+	return ok
+}
+
+// IsNotExist checks if this is an ErrNotExist error.
+func IsNotExist(err error) bool {
+	_, ok := err.(ErrNotExist)
+	return ok
+}
--- a/workspace/workspace.go
+++ b/workspace/workspace.go
@@ -1,0 +1,104 @@
+package workspace
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+var rgxSerialSuffix = regexp.MustCompile(`-\d*$`)
+
+// Workspace represents a user's Exercism workspace.
+// It may contain a user's own exercises, and other people's
+// exercises that they've downloaded to look at or run locally.
+type Workspace struct {
+	Dir string
+}
+
+// New returns a configured workspace.
+func New(dir string) Workspace {
+	return Workspace{Dir: dir}
+}
+
+// Locate the matching directories within the workspace.
+// This will look for an exact match on absolute or relative paths.
+// If given the base name of a directory with no path information it
+// It will look for all directories with that name, or that are
+// named with a numerical suffix.
+func (ws Workspace) Locate(exercise string) ([]string, error) {
+	// First assume it's a path.
+	dir := exercise
+
+	// If it's not an absolute path, make it one.
+	if !filepath.IsAbs(dir) {
+		var err error
+		dir, err = filepath.Abs(dir)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// If it exists, we were right. It's a path.
+	if _, err := os.Stat(dir); err == nil {
+		if !strings.HasPrefix(dir, ws.Dir) {
+			return nil, ErrNotInWorkspace(exercise)
+		}
+
+		src, err := filepath.EvalSymlinks(dir)
+		if err == nil {
+			return []string{src}, nil
+		}
+	}
+
+	// If the argument is a path, then we should have found it by now.
+	if strings.Contains(exercise, string(os.PathSeparator)) {
+		return nil, ErrNotExist(exercise)
+	}
+
+	var paths []string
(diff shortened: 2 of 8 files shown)


workspace/errors.go
```diff
@@ -0,0 +1,19 @@
...
+
+// ErrNotInWorkspace signals that the target directory is outside the configured workspace.
+var ErrNotInWorkspace = errors.New("not in workspace")
+
+// ErrNotExist signals that the target directory could not be located.
+var ErrNotExist = errors.New("not found")
+
+// IsNotInWorkspace checks if this is an ErrNotInWorkspace error.
+func IsNotInWorkspace(err error) bool {
+	return err == ErrNotInWorkspace
```
Tonkpils: I haven't seen this done very much in go. Is this a different way of doing something like `err.(*ErrNotInWorkspace)`? I need to brush up again in error handling 😞

kytrinyx: Oh, no, you're right. It should absolutely be a type assertion.

kytrinyx:
This is failing on a test that passes on my machine. I wonder if this is a difference in symlinks between Mac and Linux.

workspace/workspace.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// First assume it's a path.
+	dir := exercise
+
+	// If it's not an absolute path, make it one.
+	if !filepath.IsAbs(dir) {
+		var err error
+		dir, err = filepath.Abs(dir)
+		if err != nil {
+			return paths, err
```
Tonkpils: The caller should be ignoring that returned value anyway, should we return `nil` for paths instead?

workspace/workspace.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+
+	// If it exists, we were right. It's a path.
+	if _, err := os.Stat(dir); err == nil {
+		if !strings.HasPrefix(dir, ws.Dir) {
+			return paths, ErrNotInWorkspace
+		}
+
+		src, err := filepath.EvalSymlinks(dir)
```
Tonkpils: What kind of errors can we expect from this function? Is there anything worth testing against?

workspace/workspace.go
kytrinyx: paths is still nil here, but it's definitely worth making it more explicit

workspace/workspace.go
kytrinyx: I couldn't think of anything. The only thing is path errors, which would suggest that it's not a symlink or that the symlink doesn't go anywhere. In both of these cases, we just want to go with the original path and return an error if we can't find it.

workspace/errors.go
Tonkpils: Actually, I deleted the comment because I realized that type assertion is done when creating a struct that matches the error interface. Doing this equality check seems to be ok https://play.golang.org/p/g_C1nEq7Ac

Tonkpils:
> This is failing on a test that passes on my machine. I wonder if this is a difference in symlinks between Mac and Linux.

I just pulled the branch and I'm getting similar errors

```sh
--- FAIL: TestLocate (0.00s)
        Error Trace:    workspace_test.go:152
	Error:		Received unexpected error not found
	Messages:	don't be confused by a symlinked file named the same as an exercise

        Error Trace:    workspace_test.go:153
	Error:		Not equal: []string{"/Users/tonkpils/go/src/github.com/exercism/cli/fixtures/locate-exercise/workspace/actions/date"} (expected)
			        != []string(nil) (actual)

			Diff:
			--- Expected
			+++ Actual
			@@ -1,4 +1,2 @@
			-([]string) (len=1 cap=1) {
			- (string) (len=94) "/Users/tonkpils/go/src/github.com/exercism/cli/fixtures/locate-exercise/workspace/actions/date"
			-}
			+([]string) <nil>

	Messages:	don't be confused by a symlinked file named the same as an exercise

FAIL
FAIL	github.com/exercism/cli/workspace	0.015s
```

kytrinyx:
> I just pulled the branch and I'm getting similar errors

Huh. That's good news, then. Maybe my symlink didn't make it into git. One sec!

kytrinyx:
Yes! That was it. Thanks ✨

kytrinyx:
... or maybe not. Still failing. What the heck?

Tonkpils:
> Yes! That was it. Thanks ✨

I think it's still missing the file `date` under actions

kytrinyx:
Wait. Yes. It is. I put the keep file in the wrong place.

kytrinyx:
I switched out the error values with types and type assertions, @Tonkpils.

PR #426, opened 2017-08-07
[nextercism] Create solution metadata

kytrinyx:
We are going to store some metadata about a solution in a hidden file.
This currently only hides the file on Linux/MacOS.
I need to work out the hidden file for Windows, presumably using
`syscall.SetFileAttributes(f, syscall.FILE_ATTRIBUTE_HIDDEN)`.

We're going to use this in the `download` (store the metadata), `submit` (find the solution id), and `open` (look up the web url) commands.

### DIFF
--- a/visibility/hide_file_windows.go
+++ b/visibility/hide_file_windows.go
@@ -1,0 +1,27 @@
+package visibility
+
+import "syscall"
+
+// HideFile sets a Windows file's 'hidden' attribute.
+// This is the equivalent of giving a filename on
+// Linux or MacOS a leading dot (e.g. .bash_rc).
+func HideFile(path string) error {
+	// This is based on the discussion in
+	// https://www.reddit.com/r/golang/comments/5t3ezd/hidden_files_directories/
+	// but instead of duplicating all the effort to write the file, this takes
+	// the path of a written file and then flips the bit on the relevant attribute.
+	// The attributes are a bitmask (uint32), so we can't call
+	// SetFileAttributes(ptr, syscall.File_ATTRIBUTE_HIDDEN) as suggested, since
+	// that would wipe out any existing attributes.
+	ptr, err := syscall.UTF16PtrFromString(path)
+	if err != nil {
+		return err
+	}
+	attributes, err := syscall.GetFileAttributes(ptr)
+	if err != nil {
+		return err
+	}
+
+	attributes |= syscall.FILE_ATTRIBUTE_HIDDEN
+	return syscall.SetFileAttributes(ptr, attributes)
+}
--- a/workspace/solution.go
+++ b/workspace/solution.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/exercism/cli/visibility"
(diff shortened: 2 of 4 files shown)


workspace/solution.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Solution contains metadata about a user's solution.
+type Solution struct {
+	Track       string    `json:"track"`
+	Exercise    string    `json:"exercise"`
+	ID          string    `json:"id"`
+	URL         string    `json:"url"`
+	Handle      string    `json:"handle"`
+	IsRequester bool      `json:"is_requester"`
+	SubmittedAt time.Time `json:"submitted_at"`
```
kytrinyx: I just realized that we will sometimes download a fresh exercise where the solution hasn't been submitted yet. There will still be a _solution_. It has an ID, URL, etc. It just hasn't been submitted, so there's no timestamp.

I think I need a custom time format that can be a pointer.

kytrinyx: Ok, I made time.Time a pointer, even though we have to jump through some hoops to take an address from a time.Time value. We'll see how painful it is and can always change it later if necessary.

kytrinyx:
OK, super not sure about the windows hidden file thing. I've never tried this before. I read the docs and the source code for the `syscall` stuff. I can't test this locally (no Windows), so I kind of expect to have gotten this wrong.

If someone would pull this down on Windows and try running the tests (wait, how do you do `go test $(go list ./... | grep -v vendor)` on windows?) that would be a great help.

If you have the spare cycles to do something like the below in a temporary folder and see what happens, that would be cool too:

```
package main

import (
	"io/ioutil"
	"os"

	"github.com/exercism/cli/visibility"
)

func main() {
	f := `can-you-see-me.txt`
	ioutil.WriteFile(f, []byte(`o hai, didn't see you there`), os.FileMode(0755))
	visibility.HideFile(f)
}
```

Then figure out if `can-you-see-me.txt` is hidden or not (I don't even know how you would find out, sorry!).

Tonkpils:
Does this help? https://github.com/inconshreveable/go-update/blob/master/hide_windows.go I'm not on a windows computer but I know that works as I've tested it on it before.

kytrinyx:
Oh, yeah. Interesting. I would definitely not have thought about that approach. Did you copy/paste to use it? I guess you'd have to since it's not exported. I wish they'd export it, since we're already relying on that package. We could just call it directly.

robphoenix:
I'll give this a go tomorrow on my Windows laptop

kytrinyx:
I added a few more commits to this one. Realized that we're going to want to display solutions to people to get them to disambiguate. E.g. when they've got three different "clock" exercises, one in Go, one in Rust, and one by some other user, if they say `exercism open clock`, we need to get them to pick. So I've fleshed out solution a little bit to give us a `String()` method that will help us with that.

visibility/hide_file_windows.go
```diff
@@ -0,0 +1,27 @@
...
+	ptr, err := syscall.UTF16PtrFromString(path)
+	if err != nil {
+		return err
+	}
+	attributes, err := syscall.GetFileAttributes(ptr)
+	if err != nil {
+		return err
+	}
+
+	attrs |= syscall.FILE_ATTRIBUTE_HIDDEN
```
robphoenix: s/attrs/attributes

kytrinyx: whuuut. This is what I get when I can't lean on the compiler!

kytrinyx:
> I'm not sure if it's me and not knowing the right build instructions, but I had to comment out the hide_file.go contents as it was complaining about the function being redeclared. I suppose that won't be a problem with a binary will it.

I will go look at instructions about build constraints to see if I got it wrong. It shouldn't give you trouble if I do it right, I think.

kytrinyx:
@robphoenix You're right. This built just fine when I just built the binary. I don't understand why it's complaining on your machine.

I'm going to go ahead and merge this, as I really need to get on with the various commands that depend on this one, but I would love to figure the build thing out before too long.

robphoenix:
Cool, I'll have a look into it too, see what I can figure out.

PR #428, opened 2017-08-07
[nextercism] Avoid stack overflow on workspace errors

kytrinyx:
Including an err in a format string calls Error() on it... including err in a format
string _in_ Error() calls Error() which calls Error() which calls Error()...

Yeah. So that happened.


kytrinyx:
FYI I'm going to go ahead and merge this, as I want to rebase the WIP branches onto it.

PR #429, opened 2017-08-07
[nextercism] Implement open command

### LINKED ISSUE #416: nextercism: implement open command

kytrinyx:
This tries to be a little bit clever, but it's not yet smooth enough.

The command can open your own exercises or other people's. For other people's exercises, you need to specify the path to the exercise, as I imagine you'll quickly have a gajillion of these.

The logic here is: given the input, locate the possible matches. If there are no possible matches, we're done. If there are more than one match, we need to disambiguate. But if there are more than one, then it wasn't a search by path, it was a search by name. That means that we found everything by you _and_ other people. So we need to filter out the results that aren't yours.

Finally, provide an interactive selection, and let them pick the one that they want to open.

The error handling could use some work here, I just don't know where to go with it.

Closes #416

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -2,7 +2,11 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
+	"github.com/exercism/cli/browser"
+	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
 )
 
@@ -17,7 +21,60 @@ Pass either the name of an exercise, or the path to the directory that contains
 the solution you want to see on the website.
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("open called")
+		cfg, err := config.NewUserConfig()
+		BailOnError(err)
+		ws := workspace.New(cfg.Workspace)
+
+		if len(args) != 1 {
+			// TODO: usage
+			return
+		}
+
+		paths, err := ws.Locate(args[0])
+		BailOnError(err)
+
+		var solutions []workspace.Solution
+
+		for _, path := range paths {
+			solution, err := workspace.NewSolution(path)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, err.Error())
+				continue
+			}
+			solutions = append(solutions, solution)
+		}
+
+		if len(solutions) == 0 {
+			return
+		}
+
+		if len(solutions) > 1 {
+			var mine []workspace.Solution
+			for _, s := range solutions {
+				if s.IsRequester {
+					mine = append(mine, s)
+				}
+			}
+			solutions = mine
+		}
+
+		sx := workspace.Solutions(solutions)
+		for {
+			prompt := `
+We found more than one. Which one did you mean?
+Type the number of the one you want to select.
+
+%s
+> `
+			s, err := sx.Pick(prompt)
+			if err != nil {
+				fmt.Println(err)
+				continue
+			}
+			browser.Open(s.URL)
+			break
+		}
+
 	},
 }
 
--- a/workspace/solutions.go
+++ b/workspace/solutions.go
@@ -1,0 +1,65 @@
+package workspace
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// Solutions is a collection of solutions to interactively choose from.
+type Solutions []Solution
+
+// Pick lets a user interactively select a solution from a list.
+func (sx Solutions) Pick(prompt string) (Solution, error) {
+	// If there's just one, then we're done here.
+	if len(sx) == 1 {
+		return sx[0], nil
+	}
+
+	fmt.Printf(prompt, sx.Display())
+
+	n, err := sx.ReadSelection(os.Stdin)
+	if err != nil {
+		return Solution{}, err
+	}
+
+	s, err := sx.Get(n)
+	if err != nil {
+		return Solution{}, err
+	}
+	return s, nil
+}
+
+// Display shows a numbered list of the solutions to choose from.
+// The list starts at 1, since that seems better in a user interface.
+func (sx Solutions) Display() string {
+	str := ""
+	for i, s := range sx {
+		str += fmt.Sprintf("  [%d] %s\n", i+1, s)
+	}
+	return str
+}
+
+// ReadSelection reads the user's selection and converts it to a number.
+func (sx Solutions) ReadSelection(r io.Reader) (int, error) {
+	reader := bufio.NewReader(r)
+	text, _ := reader.ReadString('\n')
+	n, err := strconv.Atoi(strings.TrimSpace(text))
+	if err != nil {
+		return 0, err
+	}
+	return n, nil
+}
+
+// Get returns the solution corresponding to the number.
+// The list starts at 1, since that seems better in a user interface.
+func (sx Solutions) Get(n int) (Solution, error) {
+	if n <= 0 || n > len(sx) {
+		return Solution{}, errors.New("can't do that")
+	}
+	return sx[n-1], nil
+}
--- a/workspace/solutions_test.go
+++ b/workspace/solutions_test.go
@@ -1,0 +1,36 @@
+package workspace
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSolutions(t *testing.T) {
+	solutions := []Solution{
+		{Track: "a", Exercise: "foo"},
+		{Track: "b", Exercise: "foo"},
+		{Track: "c", Exercise: "foo"},
+	}
+	sx := Solutions(solutions)
+	display := "  [1] a/foo\n  [2] b/foo\n  [3] c/foo\n"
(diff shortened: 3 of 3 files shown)


kytrinyx:
I've reworked this quite a bit.

First I created a small abstraction over a collection of solutions that lets us interactively ask people to pick one of them. So if you have solved clock in Python, Go, Ruby, Elixir, and Haskell and you say

```
exercism open clock
```

Then you'll get a prompt that says:

```
We found more than one. Which one did you mean?
Type the number of the one you want to select it.

  [1] python/clock
  [2] go/clock
  [3] ruby/clock
  [4] elixir/clock
  [5] haskell/clock

> 
```

Then you type the number, and it opens it in the browser.

A big difference in nextercism is that we store a hidden file with solution metadata (track, url to your solution, etc). If we don't have that file we can't act on anything.

When we first launch, nobody will have these files, so we're probably going to need an `exercism migrate` command that will walk their workspace tree and download all the metadata.

NobbZ:
How will those metadata-stuff act across different workstations?

Lets say, I've two workstations A and B.

I've solved all those things at A, but I am sitting at B right now and while implementing `clock` for scala (random pick) I want to take a look at my haskell solution. Will I be able to open it up from CLI or do I have to search using the browser?

How will other metadata related tasks handle this or similar scenarios?

If this is actually a problem, shouldn't we encourage to have the exercism workspace in a git repository that the nextercism-cli adds and commits to and the user can pick where he wants to push to? Shall I open up a new proposal or is this something that isn't room for on the list right now?

nywilken:
> if this is actually a problem, shouldn't we encourage to have the exercism workspace in a git repository that the nextercism-cli adds and commits to and the user can pick where he wants to push to? Shall I open up a new proposal or is this something that isn't room for on the list right now?

This is a valid point, but I don't necessarily think we should enforce a vcs on the user or on the cli. when a user switches machines there should be an expectation that they will take the steps to get their environment setup and in sync, before starting a new exercise, may it be through vcs, USB, etc.

Instead of a git integration or even the migrate command, what if we added a `sync` command that would sync up a users metadata via the nextercism API. We would have to keep track of 
the metadata (maybe even auto generate based off submission). But the idea is that we build the dependency into the CLI and not on an external integration that may or may not exist for a user.

kytrinyx:
> How will those metadata-stuff act across different workstations?

@NobbZ The way I've been thinking about it, you can say `exercism download clock --track haskell` to download the haskell exercise + your latest solution, which will also include the metadata. At that point you could say `exercism open clock` and see haskell in the list.

If this becomes onerous, I think it's worth considering a `--remote` (`-r`) flag which would ask the API for a URL to open.

> > shouldn't we encourage to have the exercism workspace in a git repository that the nextercism-cli adds and commits to and the user can pick where he wants to push to?

> This is a valid point, but I don't necessarily think we should enforce a vcs on the user or on the cli. 

I agree that we should not dictate a VCS. We considered implementing the CLI via git, but decided not to.

> Instead of a git integration or even the migrate command, what if we added a sync command that would sync up a users metadata via the nextercism API.

That's what the migrate command is. It's a one-time thing that will take your current exercism working tree of exercises and makes sure you have metadata for everything.

> We would have to keep track of the metadata

@nywilken Would you elaborate? I'm not sure I'm following.

robphoenix:
> we're probably going to need an exercism migrate command that will walk their workspace tree and download all the metadata.

Could it be that if `exercism open` is called and there's no metadata found, it mentions this and prompts to download it. And then instead of a one-time `migrate` command, a `sync` command could be used whenever to ensure there is metadata where there should be metadata. Would that solve @NobbZ's scenario? Like doing a `git pull` on a different machine before getting started.

kytrinyx:
That might be smoother for a lot of people, @robphoenix. The problem with the scenario that @NobbZ is describing is that he doesn't have the Haskell exercise locally at all, so we wouldn't know about it.

We would find only exercises that are locally right now, we wouldn't ask the API for the relevant exercises. This is fine most of the time, but for some of the more prolific people who work on a number of different machines, I could see that not having an exercise locally is a likely occurrence.

robphoenix:
Ah, I see. I misunderstood. You wouldn't necessarily want to `sync` all your solutions on the site to the local machine. 

> a `--remote` (`-r`) flag which would ask the API for a URL to open.

In @NobbZ case instead of first downloading the exercise to then open it in a browser, could `exercism open --remote clock` ask the api for all his `clock` exercises on the site and then offer the same menu you describe above? Opening the chosen exercise without downloading it.

nywilken:
> @nywilken Would you elaborate? I'm not sure I'm following.

The impression I got from the migrate command description was that the solution metadata would only exist locally. Is that correct? 

If so, I was saying that we would have to track on the server, maybe in the user's account, or generate from the information we have so that we could push down the metadata data upon calling sync. For tracking, I was thinking that maybe upon an exercise submit we post the metadata as well so that we have all the data, if we are unable to generate it. But it seems like at minimum the track directory for each exercise would be need locally in order to store the metadata in the appropriate locations. 


My line of thinking is basically what @robphoenix called out 
>  a sync command could be used whenever to ensure there is metadata where there should be metadata

The excepting being to only sync metadata not the actual solution files.

cmd/open.go
```diff
@@ -17,7 +21,65 @@ Pass either the name of an exercise, or the path to the directory that contains
...
+		if len(args) != 1 {
+			// TODO: usage
+			return
+		}
+
+		paths, err := ws.Locate(args[0])
+		if err != nil {
+			Bail(err)
+		}
+		fmt.Println(paths)
```
robphoenix: I assume this is for debugging?

workspace/solutions.go
```diff
@@ -0,0 +1,65 @@
...
+// The list starts at 1, since that seems better in a user interface.
+func (sx Solutions) Display() string {
+	str := ""
+	for i, s := range sx {
+		str += fmt.Sprintf("  [%d] %s\n", i+1, s)
+	}
+	return str
+}
+
+// ReadSelection takes reads the user's selection and converts it to a number.
```
robphoenix: delete `takes`

NobbZ:
>>> shouldn't we encourage to have the exercism workspace in a git repository that the nextercism-cli adds and commits to and the user can pick where he wants to push to?
>>
>> This is a valid point, but I don't necessarily think we should enforce a vcs on the user or on the cli.
>
> I agree that we should not dictate a VCS. We considered implementing the CLI via git, but decided not to.

I do not see it as dictating. It is a sensible choice I made based on assumptions:

* Every user at exercism authenticates via github, so he will probably already use git
* User will probably want to use a VCS that is distributed to be able to update their own stuff even when offline
* I haven't found any tool so far that handles conflicts as good as git did.

But I totally accept a "we've already talked about and decided against". I just wanted to make clear that this wasn't a randomly made decission.

---

I still don't get what `migrate` does… Will it “update” my exercism workspace to a nextercism workspace, and I have to do it only once?

Or will this enable me to keep different workspaces on different workstations “in-sync”?

I'm doing quite a dance do keep my three workstations in sync and fail often. Since I've forgotten to `git add && git commit` after `exercism submit` in about 110% of all cases, unmanaged `git` didn't work well for me ;) Currently I try to keep them in sync via a service similar to dropbox. This clashes every now and then after I had some offline session on the train.

So an official way to stay in sync would be very valuable for me and probably others.

kytrinyx:
> The impression I got from the migrate command description was that the solution metadata would only exist locally. Is that correct?

@nywilken no, sorry. I was unclear! The source of truth for the metadata is the API, this is a local cache.

> I totally accept a "we've already talked about and decided against". I just wanted to make clear that this wasn't a randomly made decission.

Yeah, there are good reasons to suggest/choose Git, but when we balanced it against some downsides and complicating factors, we decided not to go down that route.

> I still don't get what migrate does… Will it “update” my exercism workspace to a nextercism workspace, and I have to do it only once?

Yes, this would be a one-time thing to update an existing Exercism workspace to a nextercism workspace. We might actually get away with not having this command, I need to think through some of the details.

I really like the suggestion of having the `--remote` call look at the API instead of the local filesystem, and give you the equivalent choices. I'll think about that one, too!

> I'm doing quite a dance do keep my three workstations in sync and fail often.

Oh dear! Yes, I can imagine. I'm putting this on the list of things that we absolutely need to solve.

cmd/open.go
kytrinyx: Indeed. Sheesh :) Thanks for catching that.

kytrinyx:
I'm going to merge this as a first pass, and will do some thinking about the `--remote` option.

PR #430, opened 2017-08-08
[nextercism] Fix the incorrect instructions for testing on Windows

kytrinyx:
The comment that I was directed to about how to test without vendor on
Windows turned out to not work.

Dave Cheney pointed me to something that does.

@robphoenix would you mind testing this on your Windows machine before we merge this?


robphoenix:
Sure, I'll try it today.

kytrinyx:
Ah, great. Thank you! So I'll kill this.

Issue #406, opened 2017-07-29
"\" or "\\" for "dir" in Windows CLI?


rpottsoh:
The Windows Installer for the CLI passes the install `dir` with double `\` characters in path when invoking configure.  
```
exercism configure --dir=C:\\Exercism
```

Will paths with with `\` for `dir` break CLI?
```
exercism configure --dir=c:\Exercism
```

kytrinyx:
They shouldn't. We should be using the Go `filepath` package which handles all paths in a cross-platform way. If we're not, then that would explain it.

rpottsoh:
Thanks @kytrinyx.  That is what I assumed.  It isn't idiomatic to type:
```
exercism configure --dir=c:\\exercism
```
at the command prompt in windows.  I re-checked my `.exercism.json` file.  Yup, if I enter just single `\` the CLI writes 2 to the json file.

I tried a more boneheaded thing:
```
exercism configure --dir=c:\\\exercism
```
The json file ended up with `"dir": "c:\\\\\\exercism",`

At this point `configure --dir` crashes if I try to fix the problem.  `exercism st ruby` is also broken if `dir` is bogus.  The only solution is to modify the json file by hand.  Not a big deal for some, others might have a hard time locating the file.

Could `configure --dir` test what is coming in before storing the change permanently?  Should I open a new issue for this and close this one?

kytrinyx:
Go ahead and leave this open. I think it would make a ton of sense to test what is coming in before writing it.

kytrinyx:
(also `"c:\\\\\\exercism"` is pretty hilariously bad)

NobbZ:
>  I re-checked my .exercism.json file. Yup, if I enter just single \ the CLI writes 2 to the json file.

Which is correct. On the MS Terminal and Powershell `\` can be used directly, in JSON strings it is a special char which forms a meaning with the character comming next. To actually surpress this meaning and get a single `\` after reading the JSON file, you have to escape it by prepending a `\`.

kytrinyx:
Yes, it's likely that we're getting this wrong when we're reading it back out.

kytrinyx:
@rpottsoh would you check the nextercism CLI to see if you can reproduce this behavior?

rpottsoh:
>@rpottsoh would you check the nextercism CLI to see if you can reproduce this behavior?

I'll try to take some time this evening and look into this.

rpottsoh:
@kytrinyx I tried the following:
```example
[...]

PR #431, opened 2017-08-08
[nextercism] Add user config tests for Windows

robphoenix:
I've ported the darwin tests to Windows, which led to a couple of changes in the `user_config.go` file due to different assumptions on Windows, and I've moved the test in `user_config_test.go` file to the darwin and windows tests suites because of these different assumptions.


config/user_config.go
```diff
@@ -84,9 +84,18 @@ func (cfg *UserConfig) resolve(path string) string {
 	if strings.HasPrefix(path, "~"+string(os.PathSeparator)) {
 		return strings.Replace(path, "~", cfg.Home, 1)
 	}
+	// if using "~/" on Windows
```
robphoenix: On Windows it's entirely plausible to write `~/dir` to indicate `C:\\Users\\alice\\dir`, yet `os.PathSeperator` on Windows is `\` so the conditional above doesn't work, but it does if someone writes `~\\dir`

config/user_config.go
```diff
@@ -84,9 +84,18 @@ func (cfg *UserConfig) resolve(path string) string {
...
+	if strings.HasPrefix(path, "~/") {
+		path = strings.Replace(path, "~/", "", 1)
+		return filepath.Join(cfg.Home, path)
+	}
 	if filepath.IsAbs(path) {
 		return filepath.Clean(path)
 	}
+	// if using "/dir" on Windows
```
robphoenix: On linux/darwin `/dir` represents a directory in the root, which on windows would be `C:\\`, yet my assumption would be that if a Windows user meant to use the C drive they would write `C:\\dir` and that writing `/dir` would indicate a relative dir, relative to `cfg.Home`. I may be wrong in this assumption, so happy to hear if there could be a better way.

config/user_config_test.go
```diff
@@ -1,35 +0,0 @@
...
-	assert.NoError(t, err)
-
-	// reload it
-	cfg = &UserConfig{
-		Config: New(dir, "user"),
-	}
-	err = cfg.Load(viper.New())
-	assert.NoError(t, err)
-	assert.Equal(t, "a", cfg.Token)
-	assert.Equal(t, "/a", cfg.Workspace)
```
robphoenix: I couldn't find a way to make this work across *nix & Windows, based on the different assumption of what `/a` means.

config/user_config_windows_test.go
```diff
@@ -1,9 +1,62 @@
...
+	assert.NoError(t, err)
+
+	// reload it
+	cfg = &UserConfig{
+		Config: New(dir, "user"),
+	}
+	err = cfg.Load(viper.New())
+	assert.NoError(t, err)
+	assert.Equal(t, "a", cfg.Token)
+	assert.Equal(t, filepath.Join(cfg.Home, "a"), cfg.Workspace)
```
robphoenix: Here, again, the assumption that `/a` would mean `C:\\Users\\alice\\a`

config/user_config_windows_test.go
```diff
@@ -1,9 +1,62 @@
...
+	cwd, err := os.Getwd()
+	assert.NoError(t, err)
+
+	cfg := &UserConfig{Home: "C:\\Users\\alice"}
+	tests := []struct {
+		in, out string
+	}{
+		{"", ""}, // don't make wild guesses
+		{"C:\\Users\\alice\\\\\\foobar", "C:\\Users\\alice\\foobar"},
+		{"/foobar", "C:\\Users\\alice\\foobar"},
```
robphoenix: An extra test from the Darwin ones for assuming that `/foobar` is in the home directory.

config/user_config_test.go
kytrinyx: Ah, yeah. We don't even need to have a `/a` just some jumble that comes back out the same.

config/user_config_windows_test.go
kytrinyx: This works for me.

config/user_config.go
```diff
@@ -84,9 +84,18 @@ func (cfg *UserConfig) resolve(path string) string {
 	if strings.HasPrefix(path, "~"+string(os.PathSeparator)) {
 		return strings.Replace(path, "~", cfg.Home, 1)
 	}
+	// if using "~/" on Windows
+	if strings.HasPrefix(path, "~/") {
+		path = strings.Replace(path, "~/", "", 1)
+		return filepath.Join(cfg.Home, path)
+	}
```
kytrinyx: I think if we have this, then we can delete lines 84-86, as your version should work on all platforms, right?

config/user_config_test.go
robphoenix: Ok, shall I move the test back and fix it?

config/user_config.go
robphoenix: The only case that would fail is a Windows user typing `~\\dir`

config/user_config_test.go
robphoenix: hang on, it's broken, I need my Linux.

robphoenix:
So I've refactored the test suite a bit, I may be way off, so do say if it's off the mark, and don't feel this needs to be anywhere near the top of your todo list, I imagine it's fairly substantial!

Initially I found a way to make the main user config test pass on Windows and *nix, then I copied the darwin tests to a linux test file and I didn't like having two test files that were exactly the same, so I thought we might be better off having the main `user_config_test.go` file that builds for linux and darwin, and then because it deals with filepath's so much, a separate Windows test file that can veer off and deal with Windows idiosyncrasies, without trying to make them play nice together. Does that make sense? Sorry, rambling.

I wonder if we need to set up appveyor, for these Windows tests.

config/user_config_test.go
kytrinyx: > I need my Linux.

Don't we all 😂

kytrinyx:
> Does that make sense?

Yeah that makes a lot of sense. Let me take a look through the whole thing, but your reasoning certainly sounds good to me.

> I wonder if we need to set up appveyor, for these Windows tests.

Yes! It's time. I'll get that set up. (Should be super straight forward, I already have an account).

For the build constraint... I suddenly wondered if we need a space between the constraint and the package declaration. Do you have your windows laptop handy to check if that would work? (Maybe AppVeyor will find out in a few minutes).

config/user_config_windows_test.go
```diff
@@ -1,9 +1,61 @@
...
-	t.Fatal("Need to implement normalization tests for windows.")
+	cwd, err := os.Getwd()
+	assert.NoError(t, err)
+
+	cfg := &UserConfig{Home: "C:\\Users\\alice"}
+	tests := []struct {
+		in, out string
+	}{
+		{"", ""}, // don't make wild guesses
+		{"C:\\Users\\alice\\\\\\foobar", "C:\\Users\\alice\\foobar"},
```
kytrinyx: Thanks for cleaning this up. Did you see #406? Do you know if this normalization will fix that?

robphoenix:
> I suddenly wondered if we need a space between the constraint and the package declaration.

:bulb: Yes! that's it, I'll open a PR now.

config/user_config_windows_test.go
robphoenix: I'll check and add some tests to cover those situations

config/user_config_test.go
robphoenix: I blame @jessfraz

kytrinyx: LOL. As you should :-)

config/user_config_windows_test.go
robphoenix: As far as I can tell it should. In the tests we have to have `\\` to signify `\`, the first one to escape the second, as I'm sure you realise. So here `C:\\Users\\alice` would be the same as typing `C:\Users\alice` at the command line.

Just to be sure I'll build this and give it a whirl tomorrow, doing what's mentioned in #406

kytrinyx:
@robphoenix OK if I add you to the committers on this repo?

@nywilken same goes for you.

I have you both as a generic "contributor" so that stuff can be assigned, but I think both of you have been doing great work both here and in Configlet, and I'd love to make it official here as well.

robphoenix:
> Oh. Wait.

Yeah, sorry, jumping from one machine to the other, not doing my git properly. Should be all good to merge now, phew.

> OK if I add you to the committers on this repo?

Absolutely! Thankyou :pray: :heart:

robphoenix:
> When you're happy with it go ahead and merge.

I don't have the permissions to merge yet.

nywilken:
>@robphoenix OK if I add you to the committers on this repo?
> @nywilken same goes for you.

Absolutely! Thank you.

config/user_config_windows_test.go
kytrinyx: Yeah, I'm just not sure how we ended up adding slashes in the current one /shrug

kytrinyx:
I've added you both—I'll let one of you do the honors :)

Issue #412, opened 2017-08-04
nextercism: implement version command 


kytrinyx:
_To be merged into the nextercism branch in https://github.com/exercism/cli/pull/410_

Command name: `version` (alias `v`).
`cmd/version.go`.

## Current Version

The default behavior of this command is to output the current version.

## Latest Version

When passed the `--latest` (`-l`) flag, it checks whether or not there's a newer version available.

We have logic for this in the `cli` package: https://github.com/exercism/cli/blob/master/cli/cli.go#L66

kytrinyx:
@nywilken You mentioned being interested in getting involved with the new CLI implementation. This issue should be a good first issue to get started with, to help you get your head wrapped around the project.

To work on it, pull down and branch off of the `nextercism` branch on this repository. That's where you would submit the PR back to (you can choose the "base" branch when creating the PR. If you get it wrong—I almost always do—then you can edit the PR to make it go against the correct branch after).

kytrinyx:
Oh, and to set the flag on a command you would stick this in the `init()` for the command:

```
versionCmd.Flags().BoolP("latest", "l", false, "something that describes the flag")
```

nywilken:
Sweet! Thanks for throwing this my way. I will dive in this weekend and reach out if I run into any issues or have questions.

kytrinyx:
Awesome! I'm sticking `good-first-patch` labels on a few other things as well, and then as we dig into it we'll figure out what other pieces would be good to carve off.

kytrinyx:
Did you get a chance to dive in at all? (No pressure, just curious.)

nywilken:
Not yet, I spent some time looking through the PRs you asked for my review to get a sense of how things are currently working.  I planned to attack this tonight and tomorrow.

kytrinyx:
✨ Excellent. Do holler if you run into anything, I plan to be around and working on this both tonight and tomorrow.

nywilken:
Sounds good, thanks.

nywilken:
@kytrinyx any particular reason for using `BoolP` over `BoolVarP`?

```Go
versionCmd.Flags().BoolP("latest", "l", false, "something that describes the flag")
```

Versus 

```Go
versionCmd.Flags().BooVarlP(&checkLatest, "latest", "l", false, "something that describes the flag")
```

kytrinyx:
[...]

PR #432, opened 2017-08-08
[nextercism] Implement version command

nywilken:
First stab at the version command. I will most likely split the `func showVersion` up so that the command output can be tested. Also need to add mock test for checking latest against current version.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -3,12 +3,16 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/exercism/cli/cli"
 	"github.com/spf13/cobra"
 )
 
 // Version is the version of the current build.
 // It follows semantic versioning.
 const Version = "3.0.0-alpha.1"
+
+// checkLatest flag for version command.
+var checkLatest bool
 
 // versionCmd outputs the version of the CLI.
 var versionCmd = &cobra.Command{
@@ -17,14 +21,50 @@ var versionCmd = &cobra.Command{
 	Short:   "Version outputs the version of CLI.",
 	Long: `Version outputs the version of the exercism binary that is in use.
 
-To see if there is a more recent version available, call the command with the
+To check for the latest available version, call the command with the
 --latest flag.
 	`,
+
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(Version)
+		fmt.Println(currentVersion())
+
+		if checkLatest {
+			c := cli.New(Version)
+			l, err := checkForUpdate(c)
+			BailOnError(err)
+
+			fmt.Println(l)
+		}
+
+		return
 	},
+}
+
+// currentVersion returns a formatted version string for the Exercism CLI.
+func currentVersion() string {
+	return fmt.Sprintf("exercism version %s", Version)
+}
+
+// checkForUpdate verifies if the CLI is running the latest version.
+// If the client is out of date, the function returns upgrade instructions.
+func checkForUpdate(c *cli.CLI) (string, error) {
+
+	ok, err := c.IsUpToDate()
+	if err != nil {
+		return "", err
+	}
+
+	if ok {
+		return "Your CLI version is up to date.", nil
+	}
+
+	// Anything but ok is out of date.
+	msg := fmt.Sprintf("A new CLI version is available. Run `exercism upgrade` to update to %s", c.LatestRelease.Version())
+	return msg, nil
+
 }
 
 func init() {
 	RootCmd.AddCommand(versionCmd)
+	versionCmd.Flags().BoolVarP(&checkLatest, "latest", "l", false, "check latest available version")
 }
--- a/cmd/version_test.go
+++ b/cmd/version_test.go
@@ -1,0 +1,60 @@
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/exercism/cli/cli"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCurrentVersion(t *testing.T) {
+	expected := fmt.Sprintf("exercism version %s", Version)
+
+	actual := currentVersion()
+	assert.Equal(t, expected, actual)
+}
+
+func TestVersionUpdateCheck(t *testing.T) {
+	fakeEndpoint := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, `{"tag_name": "v2.0.0"}`)
+	})
+	ts := httptest.NewServer(fakeEndpoint)
+	defer ts.Close()
+	cli.LatestReleaseURL = ts.URL
+
+	tests := []struct {
+		version  string
+		expected string
+	}{
+		{
+			// It returns new version available for versions older than latest.
+			version:  "1.0.0",
+			expected: "A new CLI version is available. Run `exercism upgrade` to update to 2.0.0",
+		},
+		{
+			// It returns up to date for versions matching latest.
+			version:  "2.0.0",
+			expected: "Your CLI version is up to date.",
+		},
+		{
+			// It returns up to date for versions newer than latest.
+			version:  "2.0.1",
+			expected: "Your CLI version is up to date.",
+		},
+	}
+
+	for _, test := range tests {
+		c := &cli.CLI{
+			Version: test.version,
+		}
+
+		actual, err := checkForUpdate(c)
+
+		assert.NoError(t, err)
+		assert.NotEmpty(t, actual)
+		assert.Equal(t, test.expected, actual)
+	}
+}
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -79,7 +79,7 @@ func (c *CLI) IsUpToDate() (bool, error) {
 		return false, fmt.Errorf("unable to parse current version (%s): %s", c.Version, err)
 	}
 
-	return rv.EQ(cv), nil
+	return cv.GTE(rv), nil
 }
 
 // Upgrade allows the user to upgrade to the latest version of the CLI.
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -15,8 +15,30 @@ func TestIsUpToDate(t *testing.T) {
 		releaseTag string
 		ok         bool
 	}{
-		{"1.0.0", "v1.0.1", false},
-		{"2.0.1", "v2.0.1", true},
+		{
+			// It returns false for versions less than release.
+			cliVersion: "1.0.0",
+			releaseTag: "v1.0.1",
+			ok:         false,
+		},
+		{
+			// It returns false for pre-release versions of release.
+			cliVersion: "1.0.1-alpha.1",
+			releaseTag: "v1.0.1",
+			ok:         false,
+		},
+		{
+			// It returns true for versions equal to release.
+			cliVersion: "2.0.1",
+			releaseTag: "v2.0.1",
+			ok:         true,
+		},
+		{
+			// It returns true for versions greater than release.
+			cliVersion: "2.0.2",
+			releaseTag: "v2.0.1",
+			ok:         true,
+		},
 	}
 
 	for _, test := range tests {


cmd/version.go
```diff
@@ -20,11 +26,45 @@ var versionCmd = &cobra.Command{
...
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(Version)
-	},
+	Run: showVersion,
+}
+
+// showVersion outputs the version of the exercism binary that is in us.
+func showVersion(cmb *cobra.Command, args []string) {
+	cmdName := RootCmd.Use
+	fmt.Printf("%s version %s", cmdName, Version)
```
kytrinyx: Good call on getting the command name dynamically.

cmd/version.go
```diff
@@ -20,11 +26,45 @@ var versionCmd = &cobra.Command{
...
+		cv, _ := semver.Make(c.Version)
+		rv, _ := semver.Make(c.LatestRelease.Version())
+
+		switch {
+		case cv.LT(rv):
+			fmt.Println(", is out of date.")
+			fmt.Printf("To upgrade to %s, see `%s upgrade`.\n", rv, cmdName)
+		case cv.GT(rv):
+			fmt.Printf(", is newer than the latest version %s\n", rv)
+		}
```
kytrinyx: I would avoid trying to DRY up the strings here. Let's make one message for "up to date" and one for "please upgrade". I'd much rather have some duplication and be able to read/edit the entire message at once.

We're eventually going to move these into configuration so that the API has control over them.

cmd/version.go
```diff
@@ -20,11 +26,45 @@ var versionCmd = &cobra.Command{
...
+	}
+
+	// Give a little extra line love.
+	fmt.Println()
```
kytrinyx: 👍

cmd/version.go
```diff
@@ -20,11 +26,45 @@ var versionCmd = &cobra.Command{
...
+
+		cv, _ := semver.Make(c.Version)
+		rv, _ := semver.Make(c.LatestRelease.Version())
+
+		switch {
+		case cv.LT(rv):
+			fmt.Println(", is out of date.")
+			fmt.Printf("To upgrade to %s, see `%s upgrade`.\n", rv, cmdName)
+		case cv.GT(rv):
+			fmt.Printf(", is newer than the latest version %s\n", rv)
```
kytrinyx: Let's assume that the only people with newer versions are people who have edited the CLI themselves... and it's OK if we tell them they're up to date, because the newest one hasn't been released yet.

Gopkg.lock
```diff
@@ -136,6 +136,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "6ade71befc0ce2335f0dc7f11d68930ccaffee771dfdc49f92673dea0b564cc8"
+  inputs-digest = "158b0077955d6d252f0b332b791bc8fb744b7620871ee86018e115c41e111dc6"
```
kytrinyx: Do we need this specifically for this change? Or is this something that we should fix on the `nextercism` branch and then rebase this PR onto that?

cmd/version.go
```diff
@@ -20,11 +26,45 @@ var versionCmd = &cobra.Command{
...
 --latest flag.
 	`,
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println(Version)
-	},
+	Run: showVersion,
+}
+
+// showVersion outputs the version of the exercism binary that is in us.
+func showVersion(cmb *cobra.Command, args []string) {
```
robphoenix: Is `cmb` a typo?

cmd/version.go
nywilken: Sounds good, thanks for clarifying.

cmd/version.go
nywilken: 👍

Gopkg.lock
nywilken: I think this may have been an issue on my end. I will verify and drop on the rebase if not needed.

cmd/version.go
nywilken: 🤦‍♂️ - good catch!

nywilken:
@kytrinyx I saw your test mock, thanks for calling that out. I have working tests and should have an update later tonight.

Gopkg.lock
kytrinyx: I ran `dep ensure` on the `nextercism` branch, and it updated this. Go ahead and rebase onto the `nextercism` branch, and it should remove this from the diff.

nywilken:
@kytrinyx @robphoenix I refactored and added tests. When you have a moment can you please take another look. Thanks a million.

cmd/version.go
```diff
@@ -20,11 +26,60 @@ var versionCmd = &cobra.Command{
 To see if there is a more recent version available, call the command with the
 --latest flag.
 	`,
+
 	Run: func(cmd *cobra.Command, args []string) {
```
robphoenix: Maybe it's just me but I think it reads easier if this is a named function, like [here](https://github.com/exercism/configlet/blob/master/cmd/lint.go#L22)

cmd/version.go
```diff
@@ -20,11 +26,60 @@ var versionCmd = &cobra.Command{
...
-		fmt.Println(Version)
+		o := currentVersion()
```
robphoenix: Do we need to create this variable `o` here, could this be `fmt.Println(currentVersion())`?
Assigning `currentVersion()` to `o` makes me think it's used later on.

cmd/version.go
```diff
@@ -20,11 +26,60 @@ var versionCmd = &cobra.Command{
...
+		fmt.Println(o)
+
+		if checkLatest {
+			c := cli.New(Version)
+			l, err := checkForUpdate(c)
+			if err != nil {
+				Bail(err)
```
robphoenix: this may change https://github.com/exercism/cli/pull/433

cmd/version.go
```diff
@@ -20,11 +26,60 @@ var versionCmd = &cobra.Command{
...
 }
 
+// currentVersion returns a formatted version string for the Exercism CLI.
+func currentVersion() string {
+	cmdName := RootCmd.Use
+	return fmt.Sprintf("%s version %s", cmdName, Version)
+}
+
+// checkForUpdate returns a string indicating if the current running version of the Exercism CLI is at the latest release.
+// For CLI versions found to be out of date a string with upgrade instructions in included in the return.
```
robphoenix: s/in included/is included/

cmd/version.go
```diff
@@ -20,11 +26,60 @@ var versionCmd = &cobra.Command{
...
+func checkForUpdate(c *cli.CLI) (o string, err error) {
+
+	ok, err := c.IsUpToDate()
+	if err != nil {
+		return "", err
+	}
+
+	o = "Your CLI version is up to date."
```
robphoenix: What do you think of removing the named returns, and changing this to `o := `. I feel like named returns should be there to add a descriptive context for what is being returned.

cmd/version.go
nywilken: 👍

cmd/version.go
nywilken: Remnants of a previous refactor I will update. 

I agree with you on the named returns. However, the following post https://blog.minio.io/golang-internals-part-2-nice-benefits-of-named-return-values-1e95305c8687 did give a different perspective on named returns.

cmd/version.go
nywilken: Cool - I'll take a look at the PR.

cmd/version.go
nywilken: Not at all. I will update to `fmt.Println(currentVersion())`

cmd/version.go
nywilken: I agree named functions read better. But my reason for changing from a named function into an anonymous function was to prevent the extra lookup (someone reading the code) and to prevent it from being called on its own. The function itself is nothing more than a wrapper around the two functions `currentVersion and checkForUpdate` so I thought it was best to define within versionCmd. 

But if it is more readable as a named function I can easily break it out.

nywilken:
@robphoenix I pushed out an update to account for your feedback. I left a comment on your comment regarding the named function. I would like to get your thoughts on my reasoning before making any changes. Thanks for the review.

cmd/version.go
robphoenix: I'm easy, I think consistency across the codebase is more important, and so far how you have it here has been the convention.

cmd/version.go
kytrinyx: This... has changed :-)

cmd/version.go
kytrinyx: Let's leave it for now, but let's also reconsider what we think the convention for the codebase should be.

cmd/version.go
```diff
@@ -20,11 +26,59 @@ var versionCmd = &cobra.Command{
...
+	}
+
+	o := "Your CLI version is up to date."
+	if ok {
+		return o, nil
+	}
+
+	// If not an exact match let's check if we are over or under
+	cv, _ := semver.Make(c.Version)
+	rv, _ := semver.Make(c.LatestRelease.Version())
```
kytrinyx: If `IsUpToDate` says "no" if you're over, then lets change `IsUpToDate` so it only complains if you're under.

cmd/version_test.go
```diff
@@ -0,0 +1,65 @@
...
+		{
+			// It returns the out of date message for versions older than latest.
+			version:  "1.0.0",
+			expected: "Your CLI version is out of date.\nTo upgrade to 2.0.0, see `exercism upgrade`.",
+		},
+		{
+			// It returns the out of date message for pre-release versions.
+			version:  "2.0.0-alpha",
+			expected: "Your CLI version is out of date.\nTo upgrade to 2.0.0, see `exercism upgrade`.",
+		},
```
kytrinyx: If we want to check that the semver library knows how to deal with prereleases, let's move that test into the tests for UpToDate.

cmd/version.go
nywilken: 👌

cmd/version_test.go
nywilken: Good idea. I'll move this test over.

nywilken:
@kytrinyx @robphoenix I pushed up what I think just might be good to go. Please give it a look over when you get a second. I should be able to pick this back up later this evening.

cmd/version.go
```diff
@@ -3,28 +3,68 @@ package cmd
...
 	},
 }
 
+// currentVersion returns a formatted version string for the Exercism CLI.
+func currentVersion() string {
+	return fmt.Sprintf("exercism version %s", Version)
+}
+
+// checkForUpdate returns a string indicating if the current running version of the Exercism CLI is at the latest release.
+// For CLI versions found to be out of date a string with upgrade instructions is included in the return.
```
kytrinyx: This is a total nitpick, but it's something I think about a lot, so I wanted to bring it up.

I try to avoid talking about data types in the comments, unless it's information that is hard to pick up from reading the code. Here my suggestion would be:

> checkForUpdate verifies if the CLI is running the latest release.
> If the client is out of date, the function returns upgrade instructions.


I wrote a bit about a related topic a while back: http://whipperstacker.com/2015/10/14/idiomatic-doc-comments-document-your-function-not-your-function-signature/

cmd/version.go
```diff
@@ -3,28 +3,68 @@ package cmd
...
+		return "", err
+	}
+
+	if ok {
+		return "Your CLI version is up to date.", nil
+	}
+
+	// Anything but ok is out of date.
+	o := fmt.Sprintf("A new CLI version is available. Run `exercism upgrade` to update to %s", c.LatestRelease.Version())
+	return o, nil
```
kytrinyx: I don't know what `o` means. For a `string` the single-letter variable name would typically be `s`.

Here I would probably use `msg`, since that's what it feels like. I don't feel strongly about `s` vs `msg`, but `o` makes me do a double-take and try to figure out what it stands for.

cmd/version.go
nywilken: 👍  I'm definitely trying to lose the language baggage - thanks for the link.

cmd/version.go
nywilken: Funny, I went back and fourth on this for a bit and thought maybe my minds playing tricks on me - guess not. I like msg.

nywilken:
Merging to close #412

nywilken:
hmm... looks like the merge commit option added a new commit, I was expecting it to act like a rebase commit. Is this a problem? 

Moving forward the option to select is rebase and merge - even if I rebase the PR?

kytrinyx:
Yeah, the default is to add a merge commit. 👍

robphoenix:
Ah, this is interesting, I did a rebase, sorry. Now I know for the future.

kytrinyx:
I don't think we have an explicit policy. Maybe we should chat about it at some point. (Totally no biggie as long as the commits make sense)

nywilken:
👍 Let's discuss in an issue and work to update the contributing doc for maintainers with push access. I would like to touch on issue automatic resolution and closing steps for anything that is pending review. I'll open an issue in docs for this later tonight.

kytrinyx:
@nywilken That would be great, thank you!

PR #410, opened 2017-08-04
[WIP] New CLI for v2 redesign of Exercism

kytrinyx:
🚧 🚧 🚧 **PLEASE DO NOT MERGE** 🚧 🚧 🚧 

This forms the basis for a complete redesign of the CLI. I'm working off of the proposal in https://github.com/exercism/discussions/issues/129#issuecomment-318867116, though there will likely be small differences as I get into the nitty-gritty details and discover that I hadn't thought certain edge cases through.

I will open issues for each of the sub-tasks, each of which will be closed by submitting a pull request **to the nextercism branch** rather than to master. After all the functionality for launch is ready, we will cut an alpha release, which will not affect the `upgrade` command for people using the current client.

I've left in a lot of the logic that we already have, but need to rewrite pretty much everything to do with downloading exercises and submitting solutions, as the new site deals with this very differently. I will also be moving all the strings that we use to communicate with users into a settings configuration file, which the CLI will update on a regular basis by calling an endpoint on the website.

This pull request does a few things:

- updates the [README][] and moves development instructions into a [Contributing guide][contributing]
- creates the basic structure of the app using the github.com/spf13/cobra library
- stubs in empty commands for all the commands that I want in the initial release
- adds basic user and api configuration
- updates the api and cli packages to work with the new configuration objects
- updates all the godeps

It leaves in all the basic api client implementation stuff, as well as the functionality that is driving the debug, upgrade, and open commands. The commands themselves will have to be reimplemented within Cobra, but that should be quite straight forward.

## New configuration implementation

The new config package in this PR is worth looking at. I've used github.com/spf13/viper as the basis of the functionality, which means that we can bind the existing configs to command-line flags where that makes sense.

We've had a number of discussions about where configuration files should go, most recently in https://github.com/exercism/cli/issues/400

In this new implementation of the CLI, we're operating with the concept of the **config home directory**, which can be accessed via `config.Dir()`. This is implemented in [dir.go][]. This directory will contain several configuration files: `user.json`, `cli.json`, and `api.json`.

The config home directory can be configured via the `EXERCISM_CONFIG_HOME` environment variable. If that is not set, the CLI will use the `XDG_CONFIG_HOME` variable. If that isn't set either, it will use `~/.config/exercism` on Linux and MacOS, and `%APPDATA%` on Windows.

I'm going to need help from the @exercism/windows folks testing the Windows part, as I don't have access to a Windows machine.

[...]

PR #433, opened 2017-08-09
[nextercism] Refactor Bail function to BailOnError

robphoenix:
Ref https://github.com/exercism/cli/pull/410#discussion_r132183548

I briefly considered
```
BailOnError(usrCfg.Load(viperUserConfig))
```
but decided against for readability.

Not as much work as I thought it might be.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -5,9 +5,11 @@ import (
 	"os"
 )
 
-// Bail handles exitable errors in commands.
+// BailOnError handles exitable errors in commands.
 // TODO: figure out what goes here.
-func Bail(err error) {
-	log.Println(err)
-	os.Exit(1)
+func BailOnError(err error) {
+	if err != nil {
+		log.Println(err)
+		os.Exit(1)
+	}
 }
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -26,20 +26,20 @@ You can also override certain default settings to suit your preferences.
 `,
 	Run: func(cmd *cobra.Command, args []string) {
 		usrCfg := config.NewEmptyUserConfig()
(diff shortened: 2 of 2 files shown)


cmd/cmd.go
```diff
@@ -5,9 +5,11 @@ import (
 	"os"
 )
 
-// Bail handles exitable errors in commands.
+// BailOnError handles exitable errors in commands.
 // TODO: figure out what goes here.
-func Bail(err error) {
-	log.Println(err)
-	os.Exit(1)
+func BailOnError(err error) {
```
nywilken: Any particular reason why we are using the log package here over a fmt?

cmd/cmd.go
```diff
@@ -5,9 +5,11 @@ import (
...
+	if err != nil {
+		log.Println(err)
```
nywilken: Since we are using the log pkg we can replace lines 12 and 13 with `log.Fatalln(err)`

cmd/cmd.go
kytrinyx: Nah, not really. I haven't figured out what the behavior on bail should be yet, so I just stuck something in there as a placeholder.

Before we release the CLI we should definitely talk about this!

cmd/cmd.go
kytrinyx: Yeah, I don't feel strongly yet about what we put here, since it's just a placeholder. If you have some thoughts about how we should deal with errors, would you go ahead and submit a PR with them so we can discuss? (Or open an issue).

For this I just want to bring it in, as it's mostly about tucking the `if err != nil {}` bit into this method than getting the actual behavior right on exit.

PR #436, opened 2017-08-09
[nextercism] Clean up the command tests

kytrinyx:
There are two commits in this one.

The first introduces an abstraction to hide the boilerplate involved in writing tests for cobra commands.

The second uses the abstraction for the configure command.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -1,0 +1,51 @@
+package cmd
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+)
+
+const cfgHomeKey = "EXERCISM_CONFIG_HOME"
+
+type CommandTest struct {
+	App            *cobra.Command
+	Cmd            *cobra.Command
+	InitFn         func()
+	TmpDir         string
+	Args           []string
+	OriginalValues struct {
+		ConfigHome string
+		Args       []string
+	}
+}
+
+func (test *CommandTest) Setup(t *testing.T) {
+	dir, err := ioutil.TempDir("", "command-test")
+	assert.NoError(t, err)
+
+	test.TmpDir = dir
+	test.OriginalValues.ConfigHome = os.Getenv(cfgHomeKey)
+	test.OriginalValues.Args = os.Args
+
+	os.Setenv(cfgHomeKey, test.TmpDir)
+
+	os.Args = test.Args
+
+	test.Cmd.ResetFlags()
+	test.InitFn()
+
+	test.App = &cobra.Command{}
+	test.App.AddCommand(test.Cmd)
+}
+
+func (test *CommandTest) Teardown(t *testing.T) {
+	os.Setenv(cfgHomeKey, test.OriginalValues.ConfigHome)
+	os.Args = test.OriginalValues.Args
+	if err := os.RemoveAll(test.TmpDir); err != nil {
+		t.Fatal(err)
+	}
+}
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -11,6 +11,27 @@ import (
 
 const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 
+// CommandTest makes it easier to write tests for Cobra commands.
+//
+// To initialize, give it the three fields Cmd, InitFn, and Args.
+// Then call Setup, and defer the Teardown.
+// The args are the faked out os.Args. The first two arguments
+// in the Args will be ignored. These represent the command (e.g. exercism)
+// and the subcommand (e.g. download).
+//
+// Finally, when you have done whatever other setup you need in your
+// test, call the command by calling Execute on the App.
+//
+// Example:
+// cmdTest := &CommandTest{
+// 	Cmd:    myCmd,
+// 	InitFn: initMyCmd,
+// 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
+// }
+// cmdTest.Setup(t)
+// defer cmdTest.Teardown(t)
+// ...
+// cmdTest.App.Execute()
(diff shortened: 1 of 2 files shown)


cmd/cmd_test.go
```diff
@@ -0,0 +1,51 @@
...
+	Cmd            *cobra.Command
+	InitFn         func()
+	TmpDir         string
+	Args           []string
+	OriginalValues struct {
+		ConfigHome string
+		Args       []string
+	}
+}
+
```
kytrinyx: I didn't create a `NewCommandTest` method, because the arguments would end up being long and messy, and I prefer to have them all on separate lines.

To initialize one of these it needs 3 of the fields: `Cmd` (e.g. `downloadCmd`), `InitFn` (e.g. `initDownloadCmd`), and `Args` (the faked out os.Args).

cmd/cmd_test.go
```diff
@@ -0,0 +1,51 @@
...
+func (test *CommandTest) Setup(t *testing.T) {
```
kytrinyx: Both `Setup` and `Teardown` take a `testing.T`, which means that they can `t.Fatal` if things go terribly wrong, or `assert.NoError` for things that might go a little bit wrong.

cmd/cmd_test.go
```diff
@@ -0,0 +1,51 @@
...
+	os.Args = test.Args
+
+	test.Cmd.ResetFlags()
+	test.InitFn()
+
+	test.App = &cobra.Command{}
+	test.App.AddCommand(test.Cmd)
+}
+
+func (test *CommandTest) Teardown(t *testing.T) {
```
kytrinyx: This is so we can `defer test.Teardown(t)` and have all the cleanup happen all in one go.

cmd/configure_test.go
```diff
@@ -1,106 +1,82 @@
...
-		os.Args = osArgs
-	}()
-
-	// Set up a bogus root command.
-	fakeCmd := &cobra.Command{}
-	// Add the real configureCmd to it.
-	fakeCmd.AddCommand(configureCmd)
-
 	tests := []struct {
+		desc           string
```
kytrinyx: This is so that we're not confused about which test is failing.

cmd/configure_test.go
```diff
@@ -1,106 +1,82 @@
...
-			args: []string{"fakeapp", "configure", "--token", "c"},
-			// It overwrites the flags that are passed without losing the ones that are not.
+			desc:           "It overwrites the flags that are passed, without losing the ones that are not.",
+			args:           []string{"fakeapp", "configure", "--token", "c"},
 			existingUsrCfg: &config.UserConfig{Token: "token-c", Workspace: "/workspace-c"},
 			expectedUsrCfg: &config.UserConfig{Token: "c", Workspace: "/workspace-c"},
-			// It gets the default API base URL.
+		},
+		{
+			desc:           "It gets the default API base URL.",
```
kytrinyx: I split apart the last test case into two separate ones. I realized that they were completely unrelated.

cmd/configure_test.go
```diff
@@ -1,106 +1,82 @@
...
-		// Override the environment to use the fake config dir.
-		os.Setenv(cfgHomeKey, dir)
+	for _, test := range tests {
+		cmdTest := &CommandTest{
+			Cmd:    configureCmd,
+			InitFn: initConfigureCmd,
+			Args:   test.args,
+		}
+		cmdTest.Setup(t)
+		defer cmdTest.Teardown(t)
```
kytrinyx: This block is what the initialization looks like for the `CommandTest` value.

cmd/configure_test.go
```diff
@@ -1,106 +1,82 @@
...
 
 		if test.existingUsrCfg != nil {
 			// Write a fake config.
 			cfg := config.NewEmptyUserConfig()
 			cfg.Token = test.existingUsrCfg.Token
 			cfg.Workspace = test.existingUsrCfg.Workspace
-			err = cfg.Write()
-			assert.NoError(t, err)
+			err := cfg.Write()
+			assert.NoError(t, err, test.desc)
```
kytrinyx: The `test.desc` is used as the last argument to all the `assert` package functions.

cmd/configure_test.go
```diff
@@ -1,106 +1,82 @@
...
-		// Now let's get a new config and see that it got written properly.
-		usrCfg, err := config.NewUserConfig()
-		assert.NoError(t, err)
-
-		assert.Equal(t, test.expectedUsrCfg.Token, usrCfg.Token)
-		assert.Equal(t, test.expectedUsrCfg.Workspace, usrCfg.Workspace)
-
-		apiCfg, err := config.NewAPIConfig()
-		assert.NoError(t, err)
+		if test.expectedUsrCfg != nil {
```
kytrinyx: I had to wrap the sets of assertions in an if statement so that I could split apart the unrelated test cases.

kytrinyx:
Dang it. I have the configure cleanup in a separate PR. I'm going to put a WIP on this—don't review it yet. I'll rebase and make it smaller after the other PR gets merged.

kytrinyx:
Alrighty, this is good to go.

cmd/cmd_test.go
robphoenix: Is it worth adding this to a doc string?
```
// CommandTest ...
//
// To initialize  it needs 3 of the fields: 
// Cmd, InitFn, and Args (the faked out os.Args)
// 
// For example:
// cmdTest := &CommandTest{
//			Cmd:    configureCmd,
//			InitFn: initConfigureCmd,
//			Args:   test.args,
//		}
```
It'd be good to have this explanation with the struct for future contributors.

kytrinyx: Ah, yes! That's a great idea.

kytrinyx:
Thanks @robphoenix. I added comments, and will merge as soon as the build is green.

PR #437, opened 2017-08-09
[nextercism] Tidy up the tests for the configure command

kytrinyx:
Make sure we have a description so that test failures are easier to track down.

Don't combine two unrelated test cases into one.


kytrinyx:
Minor cleanup. Merging.

PR #439, opened 2017-08-10
[nextercism] Figure out where to download an exercise to

kytrinyx:
Nextercism is going to let people have separate solutions exploring the same problem.
Each solution has its own ID, and can contain any number of iterations. So someone might be exploring an object oriented approach and a functional approach to the same exercise, at the same time.

They will be able to do this in directories that are numbered serially, so the first one
would be 'thingy', then the second would be 'thingy-2'.

When downloading an exercise, this means that we need to figure out which of the existing
directories for the exercise of the same name is the one that we want to write to.

If it's none of them, then we need to make a new directory.

I really wasn't sure about the naming (which is why this doesn't pass golint right now).


workspace/workspace.go
```diff
@@ -102,3 +103,53 @@ func (ws Workspace) Locate(exercise string) ([]string, error) {
...
 }
+
+func (ws Workspace) SolutionPath(exercise, solutionID string) (string, error) {
+	paths, err := ws.Locate(exercise)
+	if !IsNotExist(err) && err != nil {
+		return "", err
+	}
+
+	return ws.DetermineSolutionPath(paths, exercise, solutionID, PathContainsSolution)
+}
```
kytrinyx: This is the method that will be used in the download command. It will create a new workspace that is based on the track rather than on the entire Exercism workspace. That way it only finds the directories that contain the exercise in question (e.g. `thingy`, `thingy-2`, `thingy-3`).

The `Locate` method is the one that walks the workspace to find all the directories that match the name plus an optional numeric suffix. We're using it in the open command to find the list of potential solutions to open in the browser.

The last argument to `DetermineSolutionPath` that lets us figure out for a given path whether or not it contains the solution with the given ID. I inject the function so that I can test it without having to create fixtures for all the combinations of paths that I have.

workspace/workspace.go
```diff
@@ -102,3 +103,53 @@ func (ws Workspace) Locate(exercise string) ([]string, error) {
...
+
+// TODO: PathContainsSolution might need a different name.
```
kytrinyx: I would love some help thinking about the naming (this will need to be updated even if the names stay the same, as this is not passing golint at the moment).

workspace/workspace.go
```diff
@@ -102,3 +103,53 @@ func (ws Workspace) Locate(exercise string) ([]string, error) {
...
+			return "", err
+		}
+		if ok {
+			return path, nil
+		}
+	}
+	m := map[string]bool{}
+	for _, path := range paths {
+		m[path] = true
+	}
```
kytrinyx: If we haven't found a solution, then I want to be able to do a lookup to see whether or not we have a particular path. This creates a lookup table of the paths that were passed in.

workspace/workspace_test.go
```diff
@@ -153,3 +153,117 @@ func TestLocate(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.True(t, ok)
+
+	ok, err = PathContainsSolution(filepath.Join(root, "nope"), "abc")
+	assert.NoError(t, err)
+	assert.False(t, ok)
+
+	_, err = PathContainsSolution(filepath.Join(root, "broken"), "abc")
+	assert.Error(t, err)
+}
```
kytrinyx: I added a test for this because I wanted to be sure that I was handling the difference between a NotExists error (which is not an error in this context) and a broken solution metadata file.

workspace/workspace_test.go
```diff
@@ -153,3 +153,117 @@ func TestLocate(t *testing.T) {
...
+			filepath.Join(ws.Dir, "bat"):      "aaa",
+			filepath.Join(ws.Dir, "dog"):      "aaa",
+			filepath.Join(ws.Dir, "dog-2"):    "bbb",
+			filepath.Join(ws.Dir, "dog-3"):    "ccc",
+			filepath.Join(ws.Dir, "rabbit"):   "aaa",
+			filepath.Join(ws.Dir, "rabbit-2"): "bbb",
+			filepath.Join(ws.Dir, "rabbit-4"): "ccc",
+		}
+		return pathToSolutionID[path] == solutionID, nil
+	}
```
kytrinyx: This is the mock function that lets us get away with not having fixture directories and fixture solution metadata files for all of these tests.

workspace/workspace.go
robphoenix: `IsSolutionPath`
`PathHasSolution`
`SolutionInPath`
`IsSolutionDir`
`SolutionPathExists`
`PathExists`
`SolutionExists`
`IsSolutionExist`
`SolutionIsExist`

workspace/workspace.go
```diff
@@ -102,3 +103,57 @@ func (ws Workspace) Locate(exercise string) ([]string, error) {
...
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return s.ID == solutionID, nil
+}
+
+// TODO: DetermineSolutionPath might need a different name.
```
robphoenix: `ResolveSolutionPath`
`SolutionPath`

workspace/workspace.go
robphoenix: Nice.

workspace/workspace.go
kytrinyx: I like `IsSolutionDir` and `IsSolutionPath`. I'm thinking _path_ rather than _dir_ since we're passing in an absolute path.

If we go with one of those then we'll need to swap the order of the arguments. My thinking is:

- **path** contains **solution** -> path, solution
- is **solution** **path** -> solution, path

workspace/workspace.go
kytrinyx: I'm already using `SolutionPath` at the top level, which is why I suspect that my abstraction is a little bit forced. When it's hard to name things that tends to be a sign that my thinking isn't completely clear yet.

I like `ResolveSolutionPath` better than `DetermineSolutionPath`. I think I'll go with that.

PR #443, opened 2017-08-15
[nextercism] Implement troubleshoot command.

### LINKED ISSUE #415: nextercism: implement troubleshoot command

Command name: `troubleshoot` (alias `t`).
`cmd/troubleshoot.go`.

I chose the name `troubleshoot` rather than `debug` so that we can have a single-letter alias without crashing into the `download` command.

The current implementation (https://github.com/exercism/cli/blob/master/cmd/debug.go) is probably almost directly portable. The only difference would be that you can get a user configuration with `config.NewUserConfig()` instead of making one from scratch and populating it.

**UPDATE** There is a second difference. We'll have an explicit `/ping` endpoint to call for the Exercism API.

robphoenix:
Closes #415

### DIFF
--- a/cli/status.go
+++ b/cli/status.go
@@ -70,7 +70,11 @@ func (status *Status) Check() (string, error) {
 	status.Version = newVersionStatus(status.cli)
 	status.System = newSystemStatus()
 	status.Configuration = newConfigurationStatus(status)
-	status.APIReachability = newAPIReachabilityStatus()
+	ar, err := newAPIReachabilityStatus()
+	if err != nil {
+		return "", err
+	}
+	status.APIReachability = ar
 
 	return status.compile()
 }
@@ -85,12 +89,15 @@ func (status *Status) compile() (string, error) {
 	return bb.String(), nil
 }
 
-func newAPIReachabilityStatus() apiReachabilityStatus {
+func newAPIReachabilityStatus() (apiReachabilityStatus, error) {
+	apiCfg, err := config.NewAPIConfig()
+	if err != nil {
+		return apiReachabilityStatus{}, nil
+	}
 	ar := apiReachabilityStatus{
 		Services: []*apiPing{
 			{Service: "GitHub", URL: "https://api.github.com"},
-			{Service: "Exercism", URL: "http://exercism.io/api/v1"},
-			{Service: "X-API", URL: "http://x.exercism.io"},
+			{Service: "Exercism", URL: apiCfg.URL("ping")},
 		},
 	}
 	var wg sync.WaitGroup
@@ -99,7 +106,7 @@ func newAPIReachabilityStatus() apiReachabilityStatus {
 		go service.Call(&wg)
 	}
 	wg.Wait()
-	return ar
+	return ar, nil
 }
 
 func newVersionStatus(cli *CLI) versionStatus {
@@ -166,9 +173,8 @@ func redactToken(token string) string {
 }
 
 const tmplSelfTest = `
-
-Debug Information
-=================
+Troubleshooting Information
+===========================
 
 Version
 ----------------
@@ -212,5 +218,4 @@ https://github.com/exercism/exercism.io/issues and include
 this information.
 {{ if not .Censor }}
 Don't share your API key. Keep that private.
-{{ end }}
-`
+{{ end }}`
--- a/config/api_config.go
+++ b/config/api_config.go
@@ -13,6 +13,7 @@ var (
 		"download":      "/solutions/%s",
 		"submit":        "/solutions/%s",
 		"prepare-track": "/tracks/%s/settings",
+		"ping":          "/ping",
 	}
 )
 
@@ -41,7 +42,6 @@ func (cfg *APIConfig) SetDefaults() {
 	if cfg.BaseURL == "" {
 		cfg.BaseURL = defaultBaseURL
 	}
-
 	if cfg.Endpoints == nil {
 		cfg.Endpoints = defaultEndpoints
 		return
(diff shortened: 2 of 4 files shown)


robphoenix:
I've not included tests. I wondered if any tests should be against `status.go` rather than parsing the output from the troubleshoot command. If there are tests that should be here let me know and I'll write them up.

Example output for reference:

```shell
gopath/src/github.com/exercism/cli [troubleshoot ≡] ◯ 
> go run exercism/main.go troubleshoot     


Troubleshooting Information
===========================

Version
----------------
Current: 3.0.0-alpha.1
Latest:  2.4.1


Operating System
----------------
OS:           linux
Architecture: amd64


Configuration
----------------
Home:      /home/rob
Workspace: /home/rob/Dropbox/code/exercism
Config:    /home/rob/.config/exercism/user.json
API key:   6a99*************************bb0

API Reachability
----------------

GitHub:
    * https://api.github.com
    * [connected]
    * 159.922897ms

Exercism:
    * http://exercism.io/api/v1
    * [connected]
    * 495.754551ms

X-API:
    * http://x.exercism.io/ping
    * [connected]
    * 398.858616ms


If you are having trouble please file a GitHub issue at
https://github.com/exercism/exercism.io/issues and include
this information.



```

kytrinyx:
> I wondered if any tests should be against status.go rather than parsing the output from the troubleshoot command.

Yeah, I completely agree. Doing an example test of the troubleshoot output or whatever is messy and gross. I actually tried doing it when I extracted the debug stuff into the `cli` package and... gave up.

cli/status.go
```diff
@@ -90,7 +90,7 @@ func newAPIReachabilityStatus() apiReachabilityStatus {
 		Services: []*apiPing{
 			{Service: "GitHub", URL: "https://api.github.com"},
 			{Service: "Exercism", URL: "http://exercism.io/api/v1"},
-			{Service: "X-API", URL: "http://x.exercism.io"},
+			{Service: "X-API", URL: "http://x.exercism.io/ping"},
```
kytrinyx: I just realized—we're going to be killing off the x-api (yay).

So we will only have one exercism API to test against. The URL is `https://api.exercism.io/v1/ping` right now. I don't know how much of the `config.APIConfig` is present in the codebase at the moment, but we can rework this to use load up that config, which has the `BaseURL` (which I think is https://api.exercism.io/v1 right now), and we can add a `ping` endpoint to the default endpoints.

config/api_config.go
```diff
@@ -8,10 +8,11 @@ import (
 )
 
 var (
-	defaultBaseURL   = "https://api.exercism.com/v1"
```
robphoenix: this is a mistake, right?

config/api_config.go
```diff
@@ -78,3 +78,8 @@ func (cfg *APIConfig) Load(v *viper.Viper) error {
 	cfg.readIn(v)
 	return v.Unmarshal(&cfg)
 }
+
+// APIEndpoint returns the full URL for a given endpoint
+func APIEndpoint(endpoint string) string {
```
robphoenix: I wanted to be able to use the "ping" default endpoint without having to create a new API config and dealt with the error. Let me know if this is the wrong way to go about this.

robphoenix:
> maybe a test for Censor to start since its a flag in troubleshoot?

good idea, added.

config/api_config.go
kytrinyx: Indeed. That should be fixed in the nextercism branch.

config/api_config.go
kytrinyx: This will work as long as people are not trying to test something locally. If they've overridden the base url in their local config to work against a development server, this won't pick it up.

robphoenix: You're right, I've taken this out, and used `NewAPIConfig()`. Is the error propagation alright?

kytrinyx: >  Is the error propagation alright?

I'm not following. Would you elaborate?

kytrinyx: Oh, wait. You mean the err from the NewAPIConfig()? I think that looks fine. It really shouldn't error out under normal circumstances (I guess that goes for most things, but still. Some errors are fairly expected, whereas this is going to be a very rare occurrence).

robphoenix: Yeah I figured that would be the case, but I thought I'd best be safe than sorry.

cli/status.go
```diff
@@ -85,12 +89,15 @@ func (status *Status) compile() (string, error) {
...
+	apiCfg, err := config.NewAPIConfig()
+	if err != nil {
+		return apiReachabilityStatus{}, nil
+	}
 	ar := apiReachabilityStatus{
 		Services: []*apiPing{
 			{Service: "GitHub", URL: "https://api.github.com"},
-			{Service: "Exercism", URL: "http://exercism.io/api/v1"},
-			{Service: "X-API", URL: "http://x.exercism.io"},
+			{Service: "Exercism", URL: apiCfg.BaseURL + apiCfg.Endpoints["ping"]},
```
kytrinyx: I think this could be `apiCfg.URL("ping")`. It basically does exactly this, so I'm not sure it's an improvement :)

kytrinyx: I improved the URL() function in #451

robphoenix: I think I actually used `URL("ping")` initially, but removed it to avoid creating the error from `NewAPIConfig()` and then completely forgot it existed when I changed it to what it is now. Brain like a sieve. Or as my son keeps saying *My brain forgot to tell me*.

kytrinyx: > My brain forgot to tell me.

That is lovely. My brain does that all the time.

PR #444, opened 2017-08-16
[nextercism] Detect the type of a path

kytrinyx:
We are pretty flexible in the type of argument that we receive. It might be
an actual path, or it might be a conceptual path: the name of an exercise that
we need to walk the workspace to find. If it's a conceptual path we might find
more than one, in which case we'll do an interactive thing to disambiguate. If
it's a real path, then we can do a lot less guessing.

This is in preparation for the submit command.
We need to figure out what the argument to the submit command is. Or if multiple, what each of them are.


workspace/path_type.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// DetectPathType determines whether the given path is a directory, a file, or the name of an exercise.
+func DetectPathType(path string) (PathType, error) {
+	// If it's not an absolute path, make it one.
+	if !filepath.IsAbs(path) {
+		var err error
+		path, err = filepath.Abs(path)
+		if err != nil {
+			return 0, err
```
nywilken: This is a little confusing seeing as TypeExerciseID is also 0. What do you think about starting the constants at 1 or maybe returning a nagative number.

Tonkpils: > returning a nagative number

👍

kytrinyx: I like the idea of returning a negative number. Very clean!

kytrinyx:
This should be good to go now.

PR #446, opened 2017-08-16
[nextercism] Implement prepare command

### LINKED ISSUE #418: nextercism: implement prepare command

Command name: `prepare` (alias `p`).
`cmd/prepare.go`.

I was going to call this `setup`, but then it crashes with the alias for `submit`.

## Prepare the Exercism CLI

```
$ exercism prepare
```

The CLI ships with default messaging, but will call the `/cli/settings` endpoint on the API to get updated copy.

This will let us update copy without releasing a new version of the CLI, perform A/B tests, and translate the CLI to other (human) languages.

Store the timestamp/version of the latest download in the config. The API will provide a header on all responses containing the version of the latest settings, which will allow us to auto-update these when they're outdated.

We need to implement a user setting that lets people explicitly turn off auto-update.

## Prepare a language track

```
$ exercism prepare haskell
```

Call the API to get a list of files to download and settings to store for a language track.

The files live in the `global` directory in the track repository. This is useful in the case where we need a harness or some shims in order to do Exercism exercises in that language, or where there is no official testing library or package management system to deliver them.

The settings are things like the `test_pattern` from the track's `config.json` so that we don't auto-upload the tests when someone submits a whole directory.

Store the timestamp of the latest download. The API will provide a header on all responses containing the version of the latest settings, which will allow us to auto-update these when they're outdated.

kytrinyx:
The prepare command will eventually need other things, but I don't know what they are yet. For now, we're just getting the test pattern from the track config so that we can filter files out when we submit them.

The prepare logic will also need to be called from the submit command in the case where we don't have the settings for the track yet. We need this when someone submits a solution by indicating a directory rather than a list of files, so that we can filter out files that shouldn't be submitted.

Closes #418 though I will be opening a new issue to deal with the auto-checking of last version etc, which still needs some hashing out.

We are also hashing out the details of the files to download during track setup (https://github.com/exercism/gnu-apl/pull/6). Once we've got that sorted, I'll open an issue describing what needs to be done.

### DIFF
--- a/cmd/prepare.go
+++ b/cmd/prepare.go
@@ -1,8 +1,12 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"net/http"
 
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
 	"github.com/spf13/cobra"
 )
 
@@ -10,7 +14,7 @@ import (
 var prepareCmd = &cobra.Command{
 	Use:     "prepare",
 	Aliases: []string{"p"},
-	Short:   "Prepare does generic setup for Exercism and its tracks.",
+	Short:   "Prepare does setup for Exercism and its tracks.",
 	Long: `Prepare downloads settings and dependencies for Exercism and the language tracks.
 
 When called without any arguments, this downloads all the copy for the CLI so we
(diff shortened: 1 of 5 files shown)


kytrinyx:
This is now ready for review.

PR #448, opened 2017-08-16
[nextercism] Implement workspace command

### LINKED ISSUE #413: nextercism: implement workspace command

Command name: `workspace` (alias `w`).
`cmd/workspace.go`.

This outputs the path to the user's workspace directory.

To get this value, load up a `config.NewUserConfig()`, which has a `Workspace` field.

kytrinyx:
This adds a simple `workspace` command that outputs the absolute path to the Exercism workspace. This can be composed with the `cd` command on Mac, Linux, or in Windows Powershell in order to take the user straight there.

```
cd $(exercism workspace)
```

Closes #413

### DIFF
--- a/cmd/workspace.go
+++ b/cmd/workspace.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/exercism/cli/config"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/workspace.go
```diff
@@ -15,16 +16,18 @@ var workspaceCmd = &cobra.Command{
 
 This command can be combined with shell commands to take you there.
 
-For example, on Linux or MacOS you can run:
+For example you can run:
```
nywilken: I think the descriptions  could be reworded. I recommend something like

> Short: Workspace outputs the path to your configured Exercism workspace.  
> Long: ` Workspace outputs a normalized version of the configured Exercism workspace path that can be used for scripting. `

robphoenix: I don't think the repetition of `workspace` sounds right, and I'd change `outputs` to `prints out`, so maybe `Print out the path to your Exercism workspace.`?

nywilken: @robphoenix your version reads better.

kytrinyx:
I've given some new wording a go. Would you take one last look at this to see if it reads nicely?

cmd/workspace.go
```diff
@@ -3,28 +3,32 @@ package cmd
...
 
     cd $(exercism workspace)
 
-On Windows, the equivalent command is:
-
-    TODO ask @exercism/windows for help
+On Windows, this command works in Powershell, however you would need to
```
robphoenix: This reference to `this command` may be confusing as to whether it is referring to the workspace command or the `cd ...` command above. Maybe `On Windows, this will work only with Powershell, however ...`

kytrinyx: Nice, yeah that's better.

kytrinyx:
I'll merge this when it goes green. I imagine we'll want to nitpick on all the wording for the entire CLI once we see it in action.

PR #451, opened 2017-08-18
[nextercism] Make the API URL function nicer and more useful

kytrinyx:
This is used in some of the commands (other PRs and branches, some of which I've not submitted).

I had things like this:

```
url := fmt.Sprintf(apiCfg.URL("prepare-track"), id) 
```

Which I think is ugly and annoying. This lets us swap it out with 

```
url := apiCfg.URL("prepare-track", id) 
```

It's a variadic function, and we'll need to pay attention to what types of arguments each url placeholder takes, but we would have had to do that anyway.

### DIFF
--- a/config/api_config.go
+++ b/config/api_config.go
@@ -55,8 +55,9 @@ func (cfg *APIConfig) SetDefaults() {
 }
 
 // URL provides the API URL for a given endpoint key.
-func (cfg *APIConfig) URL(key string) string {
-	return fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key])
+func (cfg *APIConfig) URL(key string, args ...interface{}) string {
+	pattern := fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key])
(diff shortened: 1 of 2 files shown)


config/api_config.go
```diff
@@ -55,8 +55,8 @@ func (cfg *APIConfig) SetDefaults() {
 }
 
 // URL provides the API URL for a given endpoint key.
-func (cfg *APIConfig) URL(key string) string {
-	return fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key])
+func (cfg *APIConfig) URL(key string, args ...interface{}) string {
+	return fmt.Sprintf(fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key]), args...)
```
nywilken: Nice refactor- but the double Sprintf threw me a bit. Maybe split this into two lines?

kytrinyx: Yeah, that works for me.

kytrinyx:
Tweaked. Thanks!

PR #452, opened 2017-08-18
[nextercism] Implement download command

### LINKED ISSUE #411: nextercism: implement download command

Command name: `download` (alias `d`).
`cmd/download.go`.

## Track is inferred

By default, you download by specifying the slug of the exercise. E.g.:

```
exercism download clock
```

The track can almost always be inferred based on context. Most people only work on one track. At other times we can determine the track based on the current location on the file system. If we are unable to identify a track, we will submit the request without one, and the API will respond with a list of possible tracks where this exercise is available to the user.

People can specify the `--track` flag (shortflag: `t`) to explicitly disambiguate.

## Selective overwrite

If download an exercise and there are already files there, we will interactively ask whether or not the CLI should overwrite the file (each file individually). We'll show a diff to help make the decision.

If you run the download command with `--force` (or `-f`), it will automatically overwrite any existing files. There will also be a user setting so you can set force as default.

## Downloading other people's solutions

You can download someone else's solution to something given a valid UUID, or the URL to their solution.

```
$ exercism download --uuid $UUID
$ exercism download --url $URL
```

This downloads it to a separate directory tree (TBD). E.g.

```
./$EXERCISM_DIR/users/:username/:track/:exercise
```

The separate directory tree is important, because we need to make it very easy for people to submit their exercises to version control, without accidentally storing (and exposing) other people's solutions.

We will be storing metadata about the solution in a hidden file locally, and the `open` command can be used with the path to an exercise (yours or someone else's) to make it easy to open it in the browser.

If you download multiple iterations of the same exercise, then they will be added to separate directories, with a serial, numeric suffix. This number is only for the user to disambiguate, and does not have anything to do with the actual iteration number on the site.

```
$EXERCISM_WORKSPACE/users/alice/go/clock
$EXERCISM_WORKSPACE/users/alice/go/clock-2
$EXERCISM_WORKSPACE/users/alice/go/clock-3
```

## API Payload

```
{
      solution: {
        id: solution.uuid,
        url: solution_url,
        user: {
          handle: user_handle,
          is_requester: solution.user_id == requester.id
        },
        exercise: {
          id: solution.exercise.slug,
          instructions_url: instructions_url,
          track: {
            id: solution.exercise.track.slug,
            language: solution.exercise.track.title
          }
        },
        file_download_base_url: "https://api.exercism.io/v1/solutions/#{solution.uuid}/files/",
        files: files,
        iteration: iteration_hash
      }
    }
```

## [maybe] Download all available

For the beta launch we will not be implementing a flag that lets you download all available exercises, but we will consider it for a future version.

If there is a huge need for downloading all exercises in a track (e.g. so that you can do Exercism exercises while on a long trip), then we will consider how to make that possible. It probably won't be built into the CLI, though.

## [maybe] Download all solved

For the beta launch we will not be implementing behavior to let you download everything you've done on a track all at once, but we will consider it for a future version.

### LINKED ISSUE #464: nextercism: (minor) don't try to download if files are empty
When someone calls `download` there is a slight chance that they'll get a valid response with an empty list of files to download.

We should bail with an appropriate error message. Don't worry too much about the details of what that message should say, as we're going to overhaul all the error messages before we launch publicly.

kytrinyx:
There are still rough edges here, especially when it comes to error handling, but the basics are in place.

Also this still needs to do the diff with interactive selective overwrite when downloading to an existing directory. And it needs the `--force` flag to automatically overwrite (which is what it's doing now).

Closes #411 
Closes #464

I've opened a new issue for the diff.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -2,31 +2,179 @@ package cmd
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
 
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
 )
 
-// downloadCmd lets people download exercises and associated solutions.
+// downloadCmd represents the download command
 var downloadCmd = &cobra.Command{
 	Use:     "download",
 	Aliases: []string{"d"},
 	Short:   "Download an exercise.",
-	Long: `Download an Exercism exercise to work on.
+	Long: `Download an exercise.
 
-If you've already started working on the exercise, the command will
-also download your most recent solution.
+You may download an exercise to work on. If you've already
+started working on it, the command will also download your
+latest solution.
 
-This does not automatically overwrite identically named files.
-If it finds a file with the same name, it will ask if you want to overwrite it.
-Or you can overwrite files automatically by passing the --force flag.
-
-Download other people's solution by providing the UUID.
+Download other people's solutions by providing the UUID.
 `,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("download called")
+		uuid, err := cmd.Flags().GetString("uuid")
+		BailOnError(err)
+		if uuid == "" && len(args) == 0 {
+			// TODO: usage
+			fmt.Fprintf(os.Stderr, "need an exercise name or a solution --uuid")
+			return
+		}
+		apiCfg, err := config.NewAPIConfig()
+		BailOnError(err)
+
+		var slug string
+		if uuid == "" {
+			slug = "latest"
+		} else {
+			slug = uuid
+		}
+		url := apiCfg.URL("download", slug)
+
+		client, err := api.NewClient()
+		BailOnError(err)
+
+		req, err := client.NewRequest("GET", url, nil)
+		BailOnError(err)
+
+		track, err := cmd.Flags().GetString("track")
+		BailOnError(err)
+		var exercise string
+		if len(args) > 0 {
+			exercise = args[0]
+		}
+
+		if uuid == "" {
+			q := req.URL.Query()
+			q.Add("exercise_id", exercise)
+			if track != "" {
+				q.Add("track_id", track)
+			}
+			req.URL.RawQuery = q.Encode()
+		}
+
+		payload := &downloadPayload{}
+		res, err := client.Do(req, payload)
+		BailOnError(err)
+
+		if res.StatusCode != http.StatusOK {
+			switch payload.Error.Type {
+			case "track_ambiguous":
+			default:
+				fmt.Println(payload.Error.Message)
+				os.Exit(1)
+			}
+		}
+
+		solution := workspace.Solution{
+			Track:       payload.Solution.Exercise.Track.ID,
+			Exercise:    payload.Solution.Exercise.ID,
+			ID:          payload.Solution.ID,
+			URL:         payload.Solution.URL,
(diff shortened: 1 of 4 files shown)


kytrinyx:
This is ready for review.

kytrinyx:
I ended up refactoring this... I think I might not be done refactoring.

kytrinyx:
Ok, yeah. Done for now.

I'm not really happy with this—it feels like I haven't found the right abstraction yet.

cmd/download.go
```diff
@@ -11,22 +19,200 @@ var downloadCmd = &cobra.Command{
...
+		BailOnError(err)
+
+		if res.StatusCode != http.StatusOK {
+			switch dr.payload.Error.Type {
+			case "track_ambiguous":
+				// TODO: interactive selection
+				fmt.Fprintf(os.Stderr, "You have multiple %s exercises available to you.\n", solution.Exercise)
+				fmt.Fprintln(os.Stderr, "Specify the the --track flag:")
+				for _, id := range dr.payload.Error.PossibleTrackIDs {
+					fmt.Fprintf(os.Stderr, "%s download %s --track=%s", BinName, solution.Exercise, id)
```
robphoenix: s/BinName/BinaryName ?

cmd/download.go
```diff
@@ -11,22 +19,200 @@ var downloadCmd = &cobra.Command{
...
+				}
+				os.Exit(1)
+			default:
+				BailOrError(errors.New(dr.payload.Error.Message))
+			}
+		}
+
+		solution := dr.newSolution()
```
robphoenix: Does this need to move up before lines 57 & 60?

cmd/download.go
```diff
@@ -11,22 +19,200 @@ var downloadCmd = &cobra.Command{
...
+			case "track_ambiguous":
+				// TODO: interactive selection
+				fmt.Fprintf(os.Stderr, "You have multiple %s exercises available to you.\n", solution.Exercise)
+				fmt.Fprintln(os.Stderr, "Specify the the --track flag:")
+				for _, id := range dr.payload.Error.PossibleTrackIDs {
+					fmt.Fprintf(os.Stderr, "%s download %s --track=%s", BinName, solution.Exercise, id)
+				}
+				os.Exit(1)
+			default:
+				BailOrError(errors.New(dr.payload.Error.Message))
```
robphoenix: s/BailOrError/BailOnError

cmd/download.go
kytrinyx: Gah. I must not have run the tests the final time after resolving some conflicts. That's terrible.

kytrinyx:
I backed out the broken changes.

This is essentially what is released in the alpha, with the addition of not writing empty files listed in the payload from the API.

I think this is "good enough" right now that I'm going to merge it in if it passes the build, and then I'm going to make some iterative smaller changes to improve it.

kytrinyx:
The build passed on 1.8 and tip, but failed for 1.7 due to a failure with `dep`.
I've retriggered the build.

kytrinyx:
Yeah. Dep is not working on 1.7 - what the heck? This had to have worked before.

I suspect that Dep has removed some backwards-compatible thing after 1.9 got released.

Maybe it's time to remove 1.7 from the build and add 1.9

kytrinyx:
I suppose there's some amount of irony in the fact that we can't easily pin `dep`.

kytrinyx:
I rebased onto nextercism to get the passing build, and have merged this to consolidate everything that is working in the nextercism branch.

PR #454, opened 2017-08-18
[WIP] [nextercism] Transmute submit arguments into file paths

kytrinyx:
This is a little bit of dirty alchemy, and I truly do not know if it's the right approach.

It builds on top of #453, so if you've seen that, you only have to look at the second commit in this PR.

When someone submits, they might specify a directory. Or files. Or just the name of an exercise. Depending on what the arguments are, there are numerous ways in which everything can go wrong.

So we need to figure out a few things:

- what is the root solution directory that they're trying to submit
- are they trying to specify specific files within that directory

If they're specifying specific files, then we don't want to filter out the README, test suite, and metadata automatically. If they're saying it should be submitted, then it should be submitted.

But if they're specifying a directory, then we are going to need to walk that directory, find the files, and filter out stuff we don't want to automatically push up to the site.

This function munges the arguments that are passed to the submit command, to figure out how it's going to have to handle things.

### DIFF
--- a/workspace/transmission.go
+++ b/workspace/transmission.go
@@ -1,0 +1,59 @@
+package workspace
+
+import (
+	"errors"
+	"path/filepath"
+)
+
+// Transmission is the data necessary to submit a solution.
+type Transmission struct {
+	Files   []string
+	Dir     string
+	argDirs []string
+}
+
+// NewTransmission processes the arguments to the submit command to prep a submission.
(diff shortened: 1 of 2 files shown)


kytrinyx:
> Kinda feels like an exercism exercise

Haha yes, it does. It has that smallish scope. :-)

I've pushed up the typo fix, and will merge this when the build goes green.

PR #468, opened 2017-08-20
[nextercism] Implement submit command

kytrinyx:
There are some unresolved corners in this, but I'd love to get a review of it, to see where I should go next.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -1,9 +1,21 @@
 package cmd
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"io"
+	"mime/multipart"
+	"os"
+	"path/filepath"
+	"strings"
 
+	"github.com/exercism/cli/api"
+	"github.com/exercism/cli/comms"
+	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/submit.go
```diff
@@ -15,20 +26,143 @@ var submitCmd = &cobra.Command{
...
+		BailOnError(err)
+
+		dirs, err := ws.Locate(tx.Dir)
+		BailOnError(err)
+
+		sx, err := workspace.NewSolutions(dirs)
+		BailOnError(err)
+
+		var solution *workspace.Solution
+		for {
```
kytrinyx: We only show the prompt if we actually found more than one. If we only found one we just pick it and return. It's not super obvious, but it let me make the code a little bit more straight forward, so I went with it.

kytrinyx:
The build started failing, though it passed before. I'll figure out what changed.

kytrinyx:
OK, I've rebased against the most recent `nextercism` and tweaked to take the changes there into account.

kytrinyx:
I'm going to go ahead and merge this when the tests pass. I need to start getting the `nextercism` branch consolidated, as there's too much going on in too many places for me to figure out what to fix next.

kytrinyx:
Oh. Wait. This won't work until we've gotten the `dep` thing worked out in  #478

kytrinyx:
I rebased onto nextercism to get the passing build, and have merged this to consolidate everything that is working in the nextercism branch.

PR #470, opened 2017-08-21
[WIP] [nextercism] Implment the Upgrade command

### LINKED ISSUE #414: nextercism: implement upgrade command

Command name: `upgrade` (alias `u`).
`cmd/upgrade.go`

Reimplement the Upgrade command in the nextercism CLI. The current functionality (https://github.com/exercism/cli/blob/master/cmd/upgrade.go) should be directly portable.

nywilken:
### Todo
- [ ] Confirm messaging for upgrade command
- [ ] Test on different architectures and operating systems (Windows, Linux, MacOs)

Closes #414


cmd/upgrade.go
```diff
@@ -21,7 +23,22 @@ var upgradeCmd = &cobra.Command{
 	You can always delete this file.
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("upgrade called")
+		// Enable debug output for upgrade command
+		debug.Verbose = true
```
kytrinyx: This should probably be based on a `--verbose` flag.

cmd/upgrade.go
```diff
@@ -21,7 +23,22 @@ var upgradeCmd = &cobra.Command{
...
+		if ok {
+			fmt.Println("Your CLI version is up to date.")
+			return
+		}
+
+		fmt.Printf("Upgrading to version %s\n", c.LatestRelease.Version())
+		err = c.Upgrade()
+		BailOnError(err)
+
+		fmt.Println("Upgrade complete \u2714")
```
kytrinyx: While I think most terminals handle unicode properly, I'm not sure all of them do, so I'd rather be very conservative and stick with plain old ascii here.

robphoenix: I bet this would end up being an issue on Windows! :smile:

kytrinyx:
> Confirm messaging for upgrade command.

Just leave it as is, for now. We're going to do a full pass on messaging once we have all the basics in place.

cmd/upgrade.go
nywilken: Good idea.

cmd/upgrade.go
nywilken: Awww, I really like the check mark. J/k, this along with the "Upgrading to ..." messaging will be dropped in the next PR.

nywilken:
@kytrinyx @robphoenix thanks for the review. I didn't have much time to get to this yesterday but will do so today so keep your eyes open for an updated PR. 🙏

nywilken:
I took a slightly different approach for testing the upgrade command. When you have a moment please take a look over the latest PR and leave your thoughts, comments, and suggestions. Thanks.

cli/cli.go
```diff
@@ -49,6 +49,12 @@ var (
 	LatestReleaseURL = "https://api.github.com/repos/exercism/cli/releases/latest"
 )
 
+// Updater is a simple upgradable file interface.
+type Updater interface {
+	IsUpToDate() (bool, error)
+	Upgrade() error
+}
```
kytrinyx: I think this is a good tactic.

cmd/upgrade.go
```diff
@@ -2,10 +2,16 @@ package cmd
...
 	"fmt"
+	"io"
+	"os"
 
+	"github.com/exercism/cli/cli"
+	"github.com/exercism/cli/debug"
 	"github.com/spf13/cobra"
 )
 
+var out io.Writer
```
kytrinyx: I've done the same thing in the download PR :)

cmd/upgrade.go
```diff
@@ -21,10 +27,37 @@ var upgradeCmd = &cobra.Command{
...
+	if ok {
+		fmt.Fprint(out, "Your CLI version is up to date.")
+		return nil
+	}
+
+	if err := c.Upgrade(); err != nil {
+		return err
+	}
+
+	return nil
```
kytrinyx: Lines 52-56 can be simplified to:

```
return c.Upgrade()
```

cmd/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+		cv string
+		rv string
+	}{
+		{ // Calling upgrade on an outdated CLI initiates an upgrade.
+			cv: "1.0.0",
+			rv: "2.0.0",
+		},
+		{ // Calling upgrade on a CLI that's up-to-date  does nothing.
+			cv: "2.0.1",
+			rv: "2.0.1",
```
kytrinyx: Should this be 2.0.0 for the sake of clarity, since the endpoint returns 2.0.0.?

cmd/upgrade_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/exercism/cli/cli"
+	"github.com/stretchr/testify/assert"
+)
+
+type fakeCLI struct {
+	*cli.CLI
+}
```
kytrinyx: I woke up thinking about this today.

I'm wondering if we can go all the way with the mock—we are testing all the upgrade logic elsewhere, so all we really want to make sure of here is that if

1. If the `IsUpToDate()` function responds with ok, then `Upgrade()` doesn't get called.
2. If not ok, then `Upgrade()` doesn't get called.

So we could create a fake client that doesn't embed the real one. Fake out `IsUpToDate` to just check a field that we set while setting up the test fixture (we just decide if the test should say yes or no), and then the fake upgrade command we just say "UpgradeCalled = true" or something.

nywilken: How does that saying go? 

> "If you wake up in the mornin'and you can't think of anything but singin' first...then you're supposed to be a singer"  ❤️ 

Yeah, I agree 100%. It will simplify things and make the test clearer. FYI, we don't have an actual test for cli.Upgrade "yet", but that will be a separate PR. 

>Should this be 2.0.0 for the sake of clarity, since the endpoint returns 2.0.0.?

 I saw your comment above and thought maybe we could be more explicit in the test seeing as I am just using that the version as a check that Upgrade is not being called for updated clients. So I'm glad you slept on this for a bit and came back with the suggestion. I will update and push back for final review 👍

cmd/upgrade.go
nywilken: I might've borrowed it 😉 

Funny enough, I was reading up on testing in the GPL book and came across it for testing command lines, so good call.

cli/cli.go
nywilken: Thanks! My appreciation for interfaces and embedding went through the roof as I was writing this.

nywilken:
@kytrinyx @robphoenix this is up for final review. Please give it a look for any thoughts or suggestions; if good to go could someone merge it on my behalf. Many thanks!


@robphoenix I tested the command on Windows, but please feel free to give it a run.

cmd/upgrade_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type fakeCLI struct {
+	Version       string
+	UpgradeCalled bool
+}
+
+func (fc *fakeCLI) IsUpToDate() (bool, error) {
+	ok := fc.Version == "2.0.0"
```
kytrinyx: I'd suggest simplifying even further.

```
fc.UpToDate = true
```

We don't really need to deal in real version numbers—that's kind of a red herring.

cmd/upgrade_test.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (fc *fakeCLI) Upgrade() error {
+	fc.UpgradeCalled = true
+	return nil
+}
+
+func TestUpgrade(t *testing.T) {
+	//disable printing to Stdout
+	out = ioutil.Discard
```
kytrinyx: Should we put the old `out` back after?

```
oldOut := out
out = ioutil.Discard
defer func() {
  out = oldOut
}()
```

cmd/upgrade_test.go
nywilken: Good call - that is easier and so much better.

cmd/upgrade_test.go
nywilken: Yes, I will add this in.

nywilken:
@robphoenix thanks for confirming the command on Windows 👍  I'll saw your comment on the formatting, I will address it. Thanks again y'all for the feedback. I'll update and merge later today.

Issue #462, opened 2017-08-18
Interactively ask for token if missing


kytrinyx:
If we don't have a token during download (`cmd/download.go`) then

* print "Please paste your Exercism API token here: " to the terminal and wait for input
* use `browser.Open` (like in `cmd/open.go`) to open a browser to https://v2.exercism.io/my/settings
* On enter, read the token, store it in the user settings.

## Notes

We will eventually change this to https://v2.exercism.io/my/settings/token which will contain just the token, but that page doesn't exist yet.

The `token` lives on the `config.UserConfig`. Here's an example of accessing that config: https://github.com/exercism/cli/blob/51af16d8ff2ab1783c10860d4620c90d32fdb31d/cmd/open.go#L26-L29

We have a package to handle interactive stuff (`comms`). See it in action here: https://github.com/exercism/cli/blob/51af16d8ff2ab1783c10860d4620c90d32fdb31d/cmd/submit.go#L148-L160

We already have logic for storing the token in the configure command. It might be easiest to defer to that command, like we do here: https://github.com/exercism/cli/blob/51af16d8ff2ab1783c10860d4620c90d32fdb31d/cmd/download.go#L37-L40

0sc:
Hey @kytrinyx, is this still up for grabs? I'll like to try my hands with it

neil-hickey:
Hey @kytrinyx, @0sc , I was having a quick look at this also. What benefit does requiring the API token give, if the endpoint to download does not require it? 

e.g GET to http://exercism.io/api/v1/submissions/<submission_id> does not require the token to access the information.

kytrinyx:
The `nextercism` branch, which this refers to is for v2.exercism.io, which does require the token.

It's up for grabs still, and all contributions are much appreciated!

kytrinyx:
I'm reworking the specification for the CLI to lock down what the minimal adequate specification is.

As part of this process, I am going to close this issue, as it describes a more ambitious enhancement, and I'm not even sure we'll want to do this. After we launch v2 and get some experience with the new CLI, we will spend some time looking at what enhancements we'd like to implement.

Issue #461, opened 2017-08-18
Interactively ask for workspace if missing


kytrinyx:
**This issue is for the `nextercism` client. Branch off of the `nextercism` branch, and submit the pull request to `nextercism` as the base branch, rather than to `master`.**

If someone is downloading an exercise and they haven't yet configured a workspace, interactively ask where to put it.

We can default to `$HOME/$BINARYNAME` (so that people don't mix exercism and nextercism during the beta).

Show the following message

> By default we will install exercism exercises into a new directory in your user account (`$HOME/$BINARYNAME`).
> - If you would like to stick with the default just press return to continue.
> - If you would like to change this (advanced), type a different directory below:

So, once the person presses return or enters a new directory, then output: "You can always find out where things are downloaded to by running `$BINARYNAME workspace`."

Once we have this value we need to write it to the `UserConfig` (see `cmd/configure.go` for how we're doing that).

berryhill:
@kytrinyx Cool if I take this one?

kytrinyx:
That would be awesome, thank you 🎉

PR #473, opened 2017-08-21
[nextercism] Refactor interactive stuff into its own package

kytrinyx:
It was really bugging me that the workspace thing had user-facing / interactive stuff. I've sequestered this into a new package, `comms`, which I'll use to build out other interactive stuff (such as #461 and #462).

I also added tests for `Pick()` and `ReadSelection()` (now `Selection{}.Read()`) which were missing.

### DIFF
--- a/comms/selection.go
+++ b/comms/selection.go
@@ -1,0 +1,78 @@
+package comms
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// Selection wraps a list of items.
+// It is used for interactive communication.
+type Selection struct {
+	Items  []fmt.Stringer
+	Reader io.Reader
+	Writer io.Writer
+}
+
+// NewSelection prepares an empty collection for interactive input.
+func NewSelection() Selection {
+	return Selection{
+		Reader: os.Stdin,
+		Writer: os.Stdout,
+	}
+}
+
+// Pick lets a user interactively select an option from a list.
+func (sel Selection) Pick(prompt string) (fmt.Stringer, error) {
+	// If there's just one, then we're done here.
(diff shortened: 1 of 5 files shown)


comms/selection.go
```diff
@@ -0,0 +1,78 @@
...
+
+	o, err := sel.Get(n)
+	if err != nil {
+		return nil, err
+	}
+	return o, nil
+}
+
+// Display shows a numbered list of the solutions to choose from.
+// The list starts at 1, since that seems better in a user interface.
```
robphoenix: :+1:

comms/selection.go
```diff
@@ -0,0 +1,78 @@
...
+		return 0, err
+	}
+	return n, nil
+}
+
+// Get returns the solution corresponding to the number.
+// The list starts at 1, since that seems better in a user interface.
+func (sel Selection) Get(n int) (fmt.Stringer, error) {
+	if n <= 0 || n > len(sel.Items) {
+		return nil, errors.New("we don't have that one")
```
robphoenix: you weren't tempted to change this to the following?
```
return nil, errors.New("computer says no")
```

comms/selection_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+type thing struct {
+	name   string
+	rating int
+}
+
+func (t thing) String() string {
+	return fmt.Sprintf("%s (+%d)", t.name, t.rating)
```
robphoenix: Oh this is interesting. `Display` just uses the `String` method, whatever it is, to format. Clever. Little golang :bulb: moment.

comms/selection.go
kytrinyx: I am now :)

PR #474, opened 2017-08-22
Create small abstraction for interactive -uh- interactions

kytrinyx:
If someone tries downloading or submitting something before
they've configured their token or workspace, we're going to want
to ask them to paste it to us.

For the workspace we want to be able to provide a default suggestion.

This Question abstraction provides a very small wrapper around such
an interaction, and makes it testable by providing the Reader/Writer
which can be overwritten with pre-populated readers, or a writer that
discards everything.

I wonder if something like this would make the basis for a good Exercism
exercise. It took me a while to figure out how to make this both clean
and testable. E.g. how to not print random stuff to STDOUT while running
tests and how to stub in responses from a fake STDIN.

### DIFF
--- a/comms/question_test.go
+++ b/comms/question_test.go
@@ -1,0 +1,33 @@
+package comms
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestQuestion(t *testing.T) {
+	tests := []struct {
+		desc     string
+		given    string
+		fallback string
+		expected string
+	}{
+		{"records interactive response", "hello\n", "", "hello"},
+		{"responds with default if response is empty", "\n", "Fine.", "Fine."},
+	}
+	for _, test := range tests {
+		q := &Question{
+			Reader:       strings.NewReader(test.given),
+			Writer:       ioutil.Discard,
+			Prompt:       "Say something: ",
+			DefaultValue: test.fallback,
+		}
+
+		answer, err := q.Ask()
+		assert.NoError(t, err)
+		assert.Equal(t, answer, test.expected, test.desc)
+	}
+}
--- a/comms/question.go
+++ b/comms/question.go
@@ -1,0 +1,36 @@
+package comms
+
+import (
+	"bufio"
+	"fmt"
+	"io"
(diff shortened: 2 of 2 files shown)


Tonkpils:
Just as a side step for this, I haven't had too much time to look into what's going on but I noticed we're using a interaction package, this may be worth looking into: https://github.com/AlecAivazis/survey

kytrinyx:
AlecAivazis/survey looks very nice! I will take it for a spin and see what it feels like.

kytrinyx:
Another one worth checking out: https://github.com/c-bata/go-prompt

comms/question_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func TestQuestion(t *testing.T) {
+	tests := []struct {
+		desc     string
+		given    string
+		fallback string
+		expected string
+	}{
+		{"records response", "hello\n", "", "hello"},
```
nywilken: Wondering if the value for desc should be a little more descriptive and added as the message to assert.Equal to give a little context in case a test case fails, due to some future change. Thoughts?

kytrinyx: Oh. I meant to add it to the assertion for context. What is missing from the context in terms of detail?

kytrinyx:
My brain is in overload lately, so I'm going to go ahead and push forward with what I have, and then we can swap out the interaction part to make it more elegant once we've had a chance to test out some other libraries.

This should get us to "good enough to beta test" at least.

nywilken:
All good here; merging.

robphoenix:
> we can swap out the interaction part to make it more elegant once we've had a chance to test out some other libraries.

This looks good for now, and we can easily jazz it up later once the dust has settled a bit on a working version of the CLI. The `Survey` package looks quite good for this, whereas `go-prompt` looks interesting more for `REPL` type stuff. It's based off a Python package that's been used for some interesting projects like an improved Python REPL, rather than simpler Q&A prompts. Though at a future date it might be interesting to use for some kind of interactive stats command for users.

kytrinyx:
> do you want to include all the text?

Good point @nywilken, I'll clean that up and merge this.

PR #483, opened 2017-09-17
[nextercism] Replace references of os.Stdout with cmd.Out

nywilken:
This will consolidate command outputting to a package level io.Writer.


robphoenix:
:+1: :ship:

Issue #484, opened 2017-10-01
Accept --token flag in download command


kytrinyx:
**This issue is for the `nextercism` client. Branch off of the `nextercism` branch, and submit the pull request to `nextercism` as the base branch, rather than to `master`.**

In the `cmd/download.go` file, if someone passes the `--token`, then store that token to their user config and use it to call the download endpoint.

This gives us a single thing to paste when they do the interactive CLI walkthrough.

PR #488, opened 2017-10-07
Accept `--token` flag in download command

Ondama:
related to #484

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -29,6 +29,16 @@ Download other people's solutions by providing the UUID.
 `,
 	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		token, err := cmd.Flags().GetString("token")
+		if err != nil {
+			return err
+		}
+		if token != "" {
+			RootCmd.SetArgs([]string{"configure", "--token", token})
+			if err := RootCmd.Execute(); err != nil {
+				return err
+			}
+		}
 		uuid, err := cmd.Flags().GetString("uuid")
 		if err != nil {
 			return err
@@ -197,6 +207,7 @@ type downloadPayload struct {
 func initDownloadCmd() {
 	downloadCmd.Flags().StringP("uuid", "u", "", "the solution UUID")
 	downloadCmd.Flags().StringP("track", "t", "", "the track ID")
+	downloadCmd.Flags().StringP("token", "k", "", "authentication token used to connect to exercism.io")
 }
 
 func init() {


kytrinyx:
@Ondama Thank you so much for your contribution!

I completely forgot to specify in the "good first patch" issues that these features are for the next version of the client, which live in the `nextercism` branch, not in `master`. I'm so sorry! 🌷 

Would you be willing to redo this by branching off of the `nextercism` branch, and submitting the pull request back to it? It is a completely different setup, I'm afraid, using `dep` instead of `godeps`, and `spf13/cobra` instead of `urlfav/cli`.

kytrinyx:
(I've updated all the "good first patch" issues to specify that these are off of `nextercism` rather than master.)

Ondama:
I'll update the PR :)

tleen:
@kytrinyx one way we have handled with the dichotomy of the master/nextercism branches in the go track is to add a [pull request template](https://github.com/exercism/go/blob/master/.github/PULL_REQUEST_TEMPLATE.md). It's not as good a preemptively guiding users towards the `nextercism` branch but it will give people a consistent run-down of the situation before submitting a PR. Do you think that could be valuable in this repo till `nextercism` becomes the master? 

Perhaps we should mention that there may be `nextercism` branches in repos somewhere in the docs. @Ondama did you use any specific element of the documentation to guide you in creating this PR? A note may be helpful there.

kytrinyx:
> Do you think that could be valuable in this repo till nextercism becomes the master?

Yes!

> Perhaps we should mention that there may be nextercism branches in repos somewhere in the docs.

Also yes!

tleen:
Awesome I'll try to get to this soon.

Ondama:
@tleen @kytrinyx what do you think about updating the settings of the repo to make the branch under active development (`nextercism` for now) the default branch?

kytrinyx:
> what do you think about updating the settings of the repo to make the branch under active development (nextercism for now) the default branch?

That's an excellent idea. @exercism/cli if you all agree I'll go ahead and make the change.

cmd/download.go
```diff
@@ -197,6 +208,7 @@ type downloadPayload struct {
 func initDownloadCmd() {
 	downloadCmd.Flags().StringP("uuid", "u", "", "the solution UUID")
 	downloadCmd.Flags().StringP("track", "t", "", "the track ID")
+	downloadCmd.Flags().StringP("token", "k", "", "authentication token used to connect to exercism.io")
```
kytrinyx: Dang. I totally see why this is `k`, but now I'm wracking my brains trying to figure out if we should rename the whole `token` concept so we can have a consistent and unambiguous set of flags.

cmd/download.go
```diff
@@ -29,6 +29,17 @@ Download other people's solutions by providing the UUID.
 `,
 	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		token, err := cmd.Flags().GetString("token")
+		if err != nil {
+			return err
+		}
+		if token != "" {
+			RootCmd.SetArgs([]string{"configure", "--token", token})
```
kytrinyx: Oh, this is elegant! Very nice find.

cmd/download.go
```diff
@@ -29,6 +29,17 @@ Download other people's solutions by providing the UUID.
...
+			err := RootCmd.Execute()
+			if err != nil {
```
kytrinyx: I don't have very strong feelings about this, but I tend to put this on one line when there's a single return value and it is err:

```
if err := RootCmd.Execute(); err != nil {
```

Ondama: makes sense. i'll update.

cmd/download.go
Ondama: i liked when it was called `key`

kytrinyx: Yeah, we might need to go back to that. Meanwhile, we'll keep this and re-evaluate before we launch v2. Thank you!

PR #489, opened 2017-10-08
[nextercism] Implement submit success message pending AutoApprove

berryhill:
This will implement the success message upon submit, "Your solution has been submitted successfully and has been auto-approved. You can complete the exercise and unlock the next core exercise at $URL." in the case of `auto_approved: true`.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -202,7 +202,15 @@ figuring things out if necessary.
 			return err
 		}
 
-		fmt.Fprintf(Out, "Submitted. View at %s\n", solution.URL)
+		if solution.AutoApprove == true {
+			fmt.Fprintf(Out, "Your solution has been submitted " +
+				"successfully and has been auto-approved. You can complete " +
+				"the exercise and unlock the next core exercise at %s\n",
+				solution.URL)
+		} else {
+			//TODO
+		}
+
 		return nil
 	},
 }
--- a/config/api_config.go
+++ b/config/api_config.go
@@ -57,7 +57,11 @@ func (cfg *APIConfig) SetDefaults() {
 // URL provides the API URL for a given endpoint key.
 func (cfg *APIConfig) URL(key string, args ...interface{}) string {
 	pattern := fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key])
-	return fmt.Sprintf(pattern, args...)
+	if args != nil {
+		return fmt.Sprintf(pattern, args...)
+	}
+
+	return pattern
 }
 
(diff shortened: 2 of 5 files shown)


berryhill:
Ok awesome! And yes that makes sense to have it be dynamic when that time comes. Let me know if you're taking of this here out (merging.. ect) or if that's on me..

kytrinyx:
I want a second pair of eyes for sanity's sake, so I'll let one of the others merge it. Thanks!

cmd/submit.go
```diff
@@ -202,7 +202,15 @@ figuring things out if necessary.
 			return err
 		}
 
-		fmt.Fprintf(Out, "Submitted. View at %s\n", solution.URL)
```
robphoenix: Should this line go in the `else` block below? 
I may be missing something but otherwise will there be no feedback for a submitted solution that has not been auto-approved? @kytrinyx @nywilken

berryhill:
@kytrinyx Ok totally

cmd/submit.go
berryhill: @robphoenix Yes, I probably should have commented about this. I'll change it to whatever we conclude here. Thanks for the review!

nywilken:
@kytrinyx @berryhill I unlabeled it as stale and will give it a look over before merging. @berryhill please let me know if there are any pending changes you wish to make.

config/api_config.go
```diff
@@ -57,7 +57,11 @@ func (cfg *APIConfig) SetDefaults() {
 // URL provides the API URL for a given endpoint key.
 func (cfg *APIConfig) URL(key string, args ...interface{}) string {
 	pattern := fmt.Sprintf("%s%s", cfg.BaseURL, cfg.Endpoints[key])
-	return fmt.Sprintf(pattern, args...)
+	if args != nil {
```
nywilken: I see args being empty not being the standard use case. What do you think about

```go
...
If args == nil {
  return pattern
}

return fmt.Sprintf(...)
}
```

Basically, trying to keep the normal code path at a minimal indentation.

kytrinyx:
It looks like we dropped the ball on this one. I'm going to go ahead and tweak this per your suggestions @nywilken, and then go ahead and merge it.

nywilken:
@kytrinyx thanks for pushing this one forward. The end of the year has been a bit crazy. I'm starting to get back on things as of Today.

kytrinyx:
> The end of the year has been a bit crazy

Mine too! I've gotten a fair amount of stuff done now that there's a bit of a holiday break, but it never quite ends :-)

Issue #472, opened 2017-08-21
Prevent people from submitting empty files


kytrinyx:
**This issue is for the `nextercism` client. Branch off of the `nextercism` branch, and submit the pull request to `nextercism` as the base branch, rather than to `master`.**

We haven't worked out the messaging, but we should probably bail from the entire submit with a complaint that the file is empty, rather than skipping the file silently.

We might run into things like `.keep` files which would need to be filtered out.

AdhirRamjiawan:
Is this check to validate that the method/function to be implemented has any logic in the body or is it to ensure that at least some file with some content is submitted? I would imagine the former being a real effort to implement as it would probably need to have some sort of logic per language (reflection or something). Maybe the results of the test cases could provide some clue as to whether or not code has been written?

kytrinyx:
> Is this check to validate that the method/function to be implemented has any logic in the body or is it to ensure that at least some file with some content is submitted?

No, we only care that they're not trying to submit a completely empty file. If there's any content whatsoever, we'll submit it.

PR #493, opened 2017-10-21
[nextercism] skip empty files on submit

aldraco:
Work in progress, still haven't been able to test it out. Looking to address #472.

While trying to test, I'm running into issues trying to fetch exercises with my built `testercism` binary. When I run `./testercism download hello-world` I get an empty Error. I then tried running `./testercism prepare -t python` which looked to have a string escape error, which may be happening for other commands too given they all use the same structure, though I didn't see a similar string format error for it:

https://github.com/exercism/cli/blob/nextercism/config/api_config.go#L13-L16

(`troubleshoot` does not seem to indicate any connectivity problems.)

Error for `prepare` command:

```
Error: parse https://api.exercism.io/v1/tracks/%!!(string=python)s(MISSING): invalid URL escape "%!!"
```

Are these known errors? 

Thanks! I'll take another look this weekend when I'm not so tired. :smile:


kytrinyx:
Oh, shoot. My guess is that this is a problem in the library, not with you! 💙 

I'm running on empty myself, but my plan for today is to catch up on all the CLI-related issues and PRs so I'll see if I can test this out and provide any pointers or suggestions.

kytrinyx:
Ok! I had an almost-nap, and now I'm back to feeling quite human again.

It was most certainly my fault. I've submitted a fix for it here: #495

You'll also need to call `./testercism configure --token=YOUR-TOKEN`, and you can find your token on the https://v2.exercism.io/my/settings page.

I'm pretty sure that the empty error is because we haven't done any proper error handling for stuff like that yet (... note to self!).

✨ thanks, good luck etc etc and holler if you get stuck!

aldraco:
Aha, @kytrinyx thank you! I was also trying to use my old token from v1, so thanks for pointing that out as well.

aldraco:
Thanks again for your help! Removed the [wip] tag and is ready for a review.

PR #495, opened 2017-10-22
Fix string format error in prepare command

kytrinyx:
We changed the apiCfg to have a URL method which takes the URL identifier (e.g. prepare-track) and a slug. Then I forgot to pass the track ID to the method.

This re-arranges things to make the call correctly.

@aldraco this will fix the error you're seeing when calling `prepare`.

### DIFF
--- a/cmd/prepare.go
+++ b/cmd/prepare.go
@@ -56,7 +56,7 @@ func prepareTrack(id string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


kytrinyx:
@nywilken That's a good question. Yeah, we should probably improve the tests for this. Let me give this another whirl when my brain is fresh.

kytrinyx:
I fear that my brain will never be fresh ever again, unfortunately.

I'm going to merge this in as is, and open an issue for the other question.

PR #499, opened 2017-11-02
[nextercism] skip empty files on submit

aldraco:
Well, that was hairy for a moment - rebased against `master` instead of `nextercism`, resulting in much sadness for a moment ... apologies! Got the branch sorted out, but looks like I need to open a new PR because of all of the changes, so this is a re-creation of #493, addressing issue #472.

Copying the last discussion here for convenience:

> @kytrinyx 
I'm going to wait to merge it until my PR with the fix to the prepare endpoint is fixed, then I'll rebase this against the most recent master (unless you beat me to it). 🌻

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -148,6 +148,15 @@ figuring things out if necessary.
 		}
 
 		for _, path := range paths {
+			// Don't submit empty files
+			info, err := os.Stat(path)
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks good! I think we may end up printing to an `Out` variable, but I'm not sure what the status of that is, so I'm going to pull this in now, and we can adjust later if necessary.

Thanks for the contribution 🌻

Issue #485, opened 2017-10-01
Interactive confirmation when submitting entire directory


kytrinyx:
If someone uploads an entire directory during submit (`cmd/submit.go`) we need to

* show them a list of the files that will be selected (we use the IgnorePatterns from `config/CLIConfig.go` to filter things out)
* give the user two options: hit enter to accept or ctrl-C to cancel, then submit specific files.

aldraco:
I've started working on this one. Once I figure out how string formatting works in Go, I hope to have something to PR soon.

NobbZ:
There is the [`fmt`](https://golang.org/pkg/fmt/) package, providing a bunch of functions that work similar to Cs `printf` family of functions, also there is a good description in the packages documentation.

kytrinyx:
Closing this, as it's resolved in #503

Issue #480, opened 2017-09-04
Interactive confirmation when submitting explicit list of directories


kytrinyx:
A lot of people unintentionally try to submit files that they probably don't want to submit due to similarity in filenames and autocompletion.

When someone submits an explicit list of files, we need to filter against the filter patterns. If anything matches, interactively ask if they mean to submit it. If so, go ahead and submit it as normal, if not, filter it out.

For now don't worry about the wording of the message, just use something simple like: "[file] would normally not be submitted. Do you want to submit it anyway? [y/N]"

This came out of this discussion: https://github.com/exercism/v2-feedback/issues/48

Issue #463, opened 2017-08-18
Interactively resolve ambiguous track during download


kytrinyx:
If someone runs `exercism download clock` and they have joined multiple tracks that have that exercise, then the server will respond with a 400 error, and the following payload:

```
{
	"error": {
		"type": "track_ambiguous",
		"message": "Please specify a track id",
		"possible_track_ids": ["go", "ruby"]
	}
}
```

We should interactively ask which they intended, and then go ahead and download it.

People can also specify `--track` (or `-t`) as a flag to the download command to avoid ambiguity. We should probably explain that in the interactive messaging.

kytrinyx:
I'm reworking the specification for the CLI to lock down what the minimal adequate specification is.

As part of this process, I am going to close this issue, as it describes a more ambitious enhancement, and I'm not even sure we'll want to do this. After we launch v2 and get some experience with the new CLI, we will spend some time looking at what enhancements we'd like to implement.

PR #503, opened 2017-11-30
[nextercism] Interactive confirmation when submitting directory (#485)

aldraco:
When a user submits a directory, show them a list of (pre-filtered) files and allow them to confirm/pass on the whole selection.

Changes that support the interactive prompt:
 - made `Transmission.ArgDirs` public in order to check it in `submit.go`.

Otherwise, the interaction/prompts would need to take place inside code for a Transmission, which seemed like a bad separation of concerns. In our case, as long as `ArgDirs` == 1, then the user gave a whole directory, and we don't need to figure out anything else.

For testing:
 - added an In variable, like the Out, into `cmd/root`
 - added MockInput to `cmd_test`

This makes it possible to mock interactive responses in tests. There are several Issues talking about interactive submission/confirmation etc (#485, #480, #463, #462, #461 ), so I figured having the ability to test them would be nice. If the testing changes should be a separate PR, happy to split it out!

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -36,13 +36,13 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // ...
 // cmdTest.App.Execute()
 type CommandTest struct {
-	App            *cobra.Command
-	Cmd            *cobra.Command
-	InitFn         func()
-	TmpDir         string
-	Args           []string
-	MockInput      string
-	OriginalValues struct {
+	App                     *cobra.Command
+	Cmd                     *cobra.Command
+	InitFn                  func()
+	TmpDir                  string
+	Args                    []string
+	MockInteractiveResponse string
+	OriginalValues          struct {
 		ConfigHome string
 		Args       []string
 	}
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -19,7 +19,7 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // in the Args will be ignored. These represent the command (e.g. exercism)
 // and the subcommand (e.g. download).
 // Pass any interactive responses needed for the test in a single
-// String in MockInput, separated by newlines.
+// String in MockInput, delimited by newlines.
 //
 // Finally, when you have done whatever other setup you need in your
 // test, call the command by calling Execute on the App.
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -29,7 +29,7 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // 	Cmd:    myCmd,
 // 	InitFn: initMyCmd,
 // 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
-// 	MockInput: "first-input\nsecond\n",
+// 	MockInteractiveResponse: "first-input\nsecond\n",
 // }
 // cmdTest.Setup(t)
 // defer cmdTest.Teardown(t)
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -29,7 +29,7 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // 	Cmd:    myCmd,
 // 	InitFn: initMyCmd,
 // 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
-//  MockInput: "first-input\nsecond\n"
+// 	MockInput: "first-input\nsecond\n",
 // }
 // cmdTest.Setup(t)
 // defer cmdTest.Teardown(t)
@@ -41,7 +41,7 @@ type CommandTest struct {
 	InitFn         func()
 	TmpDir         string
 	Args           []string
-	MockInput	   string
+	MockInput      string
 	OriginalValues struct {
 		ConfigHome string
 		Args       []string
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -18,6 +18,8 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // The args are the faked out os.Args. The first two arguments
 // in the Args will be ignored. These represent the command (e.g. exercism)
 // and the subcommand (e.g. download).
+// Pass any interactive responses needed for the test in a single
+// String in MockInput, separated by newlines.
 //
 // Finally, when you have done whatever other setup you need in your
 // test, call the command by calling Execute on the App.
@@ -27,6 +29,7 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // 	Cmd:    myCmd,
 // 	InitFn: initMyCmd,
 // 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
+//  MockInput: "first-input\nsecond\n"
 // }
 // cmdTest.Setup(t)
 // defer cmdTest.Teardown(t)
@@ -38,6 +41,7 @@ type CommandTest struct {
 	InitFn         func()
 	TmpDir         string
 	Args           []string
+	MockInput	   string
 	OriginalValues struct {
 		ConfigHome string
 		Args       []string
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -19,6 +19,8 @@ var (
 	BinaryName string
 	// Out is used to write to the required writer.
 	Out io.Writer
+	// In used for testing
+	In io.Reader
 )
 
 // RootCmd represents the base command when called without any subcommands.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -19,7 +19,7 @@ var (
 	BinaryName string
 	// Out is used to write to the required writer.
 	Out io.Writer
-	// In used for testing
+	// In is used to provide mocked test input (i.e. for prompts).
 	In io.Reader
 )
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -43,5 +43,6 @@ func Execute() {
 func init() {
 	BinaryName = os.Args[0]
 	Out = os.Stdout
+	In = os.Stdin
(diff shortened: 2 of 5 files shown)


cmd/cmd_test.go
```diff
@@ -27,6 +29,7 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // 	Cmd:    myCmd,
 // 	InitFn: initMyCmd,
 // 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
+//  MockInput: "first-input\nsecond\n"
```
kytrinyx: Would you mind aligning this with the other arguments and adding a comma to the end of the line so that the syntax is valid?

cmd/cmd_test.go
```diff
@@ -38,6 +41,7 @@ type CommandTest struct {
 	InitFn         func()
 	TmpDir         string
 	Args           []string
+	MockInput	   string
```
kytrinyx: Did you run `go fmt` on this? The alignment seems off.

cmd/cmd_test.go
```diff
@@ -18,6 +18,8 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // The args are the faked out os.Args. The first two arguments
 // in the Args will be ignored. These represent the command (e.g. exercism)
 // and the subcommand (e.g. download).
+// Pass any interactive responses needed for the test in a single
+// String in MockInput, separated by newlines.
```
kytrinyx: This is great, thank you!

I agree that being able to test this properly is very useful indeed. No need to add it in a separate PR, it makes sense to keep it here.

cmd/root.go
```diff
@@ -19,6 +19,8 @@ var (
 	BinaryName string
 	// Out is used to write to the required writer.
 	Out io.Writer
+	// In used for testing
```
kytrinyx: This should be a full sentence, with punctuation.

```
// In is used to provide fake input during testing.
```
(or something like that).

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
 			return errors.New("no files found to submit")
 		}
 
+		// if the user submits a directory, confirm the list of files
```
kytrinyx: Would you mind capitalizing the `if` and adding a period to the end of this?

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+		if len(tx.ArgDirs) > 0 {
+			prompt := "You specified a directory. Here are the files you are submitting:\n"
+			for i, path := range paths {
+				prompt += fmt.Sprintf(" [%d]  %s\n", i+1, path)
+			}
+			prompt += "\nPress ENTER to submit, or control + c to cancel: "
```
kytrinyx: This is excellent.

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+			
+			conf_q := &comms.Question{
+				Prompt: prompt, 
+				DefaultValue: "y",
+				Reader: In,
+				Writer: os.Stdout,
```
kytrinyx: Should this use `cmd.Out`?

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
 
+		// if the user submits a directory, confirm the list of files
+		if len(tx.ArgDirs) > 0 {
+			prompt := "You specified a directory. Here are the files you are submitting:\n"
+			for i, path := range paths {
+				prompt += fmt.Sprintf(" [%d]  %s\n", i+1, path)
+			}
+			prompt += "\nPress ENTER to submit, or control + c to cancel: "
+			
+			conf_q := &comms.Question{
```
kytrinyx: Variables should use `mixedCase` to match the Go style guide.

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+				Prompt: prompt, 
+				DefaultValue: "y",
+				Reader: In,
+				Writer: os.Stdout,
+			}
+			answer, err := conf_q.Ask()
+			if err != nil {
+				fmt.Println(err)
```
kytrinyx: This should not print the error, only return it. The error will be printed to STDERR since we're mapping the command function to `RunE`

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+				return err
+			}
+			if answer != "y" {
```
kytrinyx: How about downcasing the answer in case people provide a `Y` out of habit?

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+				fmt.Println("OK, try submitting files individually instead.")
```
kytrinyx: Should this be printed to `Out`?

cmd/submit.go
```diff
@@ -147,6 +147,32 @@ figuring things out if necessary.
...
+				return nil
+			}
+			fmt.Println("OK, submitting files now...")
```
kytrinyx: Should this be printed to `Out`?

cmd/submit_test.go
```diff
@@ -112,6 +123,9 @@ func TestSubmit(t *testing.T) {
 	err = apiCfg.Write()
 	assert.NoError(t, err)
 
+	// write some mock input for interactive commands
```
kytrinyx: I don't understand why the mock input needs to be written to the file. Would you elaborate?

aldraco: Sure - perhaps there is a better solution than what I came up with.

The challenge for interactive prompt testing was that `cmd\submit.go` uses `comms.Question` to ask the user for input, and stdin is used as the input source. As I understand, `go test` overrides stdin to point to dev/null (or something that is empty), so the only response given to the prompts from stdin during the tests was `EOF`.

All that the prompt needs to get input is a file interface (`Open` and `Read` methods), and it expects to read the response to each prompt one line at a time, so writing the prompts delimited by newline to a temp file and reading from that, one line at a time, provided the necessary interface and can be easily cleaned up afterwards.

I'll see if some more comments could clarify that intent.

aldraco: Ah, and reading this again I see another thing -- I wasn't sure if it was better to have the test tmpFile setup and teardown handled in `cmd_test.go` or in the test file itself. On one hand, it would be nice to not have to manually set that up each time. On the other hand, input/output redirection is already happening in the test file, so perhaps it makes sense to put there with other similar tasks.

aldraco:
Thank you @kytrinyx for the review -- did not know about `go fmt`, handy tool. I think I have addressed all of your review comments.

cmd/submit_test.go
kytrinyx: I *think* you can override Out so that it goes to something that you capture instead of going to /dev/null, which would let you avoid the mock, but still get the same benefit.

With respect to the test setup and teardown, I want each test to do it for clarity, and so that we can customize. (I might be missing something here, though—if what I'm saying makes no sense, feel free to push back!)

kytrinyx: Sorry. I mean override `In`, not override `Out`.

So in your test, you would set `In` to `strings.NewReader("your\nfake\ninput")`. I think.

cmd/root.go
```diff
@@ -19,6 +19,8 @@ var (
 	BinaryName string
 	// Out is used to write to the required writer.
 	Out io.Writer
+	// In is used to provide mocked test input (i.e. for prompts).
+	In io.Reader
```
kytrinyx: We would need to set this in the init function to be os.Stdin, I think.

cmd/submit_test.go
kytrinyx: Ok, looking through all of this one more time (so sorry about the noise)... I think that it makes sense to provide the `MockInput` on the `CommandTest` value, but instead of writing it to a file, it would set `In` to be a reader that contains that mock input.

I think we might also want to rename `MockInput` to `FakeInteractiveResponse` for clarity.

kytrinyx:
Once you have updated the MockInput thing (or have figured out that what I suggested is not going to work) would you add a comment here? That way I'll know to do a final review.

aldraco:
Thank you @kytrinyx -- the `strings.NewReader` worked perfectly and now there's no need for the temp file.

kytrinyx:
@aldraco Awesome! I've been traveling for the past 10 days, and am about to get on a plane to make my way home. I'll take a look at this first thing when I've gotten back home and had some sleep.

cmd/cmd_test.go
```diff
@@ -27,18 +29,20 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // 	Cmd:    myCmd,
 // 	InitFn: initMyCmd,
 // 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
+// 	MockInput: "first-input\nsecond\n",
```
kytrinyx: `MockInput` needs to reflect the updated name `MockInteractiveResponse`

cmd/submit.go
```diff
@@ -207,8 +231,8 @@ figuring things out if necessary.
 		}
 
 		if solution.AutoApprove == true {
-			fmt.Fprintf(Out, "Your solution has been submitted " +
-				"successfully and has been auto-approved. You can complete " +
+			fmt.Fprintf(Out, "Your solution has been submitted "+
+				"successfully and has been auto-approved. You can complete "+
 				"the exercise and unlock the next core exercise at %s\n",
```
kytrinyx: Were these changes introduced by a change to `gofmt`? If not, can we revert them, since they're not related to this change?

aldraco: It was, as part of [this](https://github.com/exercism/cli/pull/503/commits/6e09ba18d1536f432563d0ff44ff951cf7d2cab5) commit. Not sure why it didn't change before.

aldraco:
```
>>> go version
go version go1.9.4 linux/amd64
```

aldraco:
Thank you, @kytrinyx! Appreciate all of the feedback.

kytrinyx:
Thank you @aldraco 🌷

PR #506, opened 2017-12-22
Uninstall probot-stale on this repository

kytrinyx:
We have a lot of work that needs to be done, and we tend to work in spurts. Probot Stale
hasn't really helped much, and recently (while I was not paying attention) it closed a bunch
of issues that I'd prefer to keep open (nextercism stuff).


kytrinyx:
Awesome, thanks all!

PR #507, opened 2017-12-22
[nextercism] cmd: Various simplifications

ferhatelmas:
* `if err != nil { return err } return nil`, just returns

there were 2 more issues:
* defer is called without checking error
* error wasn't checked on walk

### DIFF
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -62,12 +62,7 @@ You can also override certain default settings to suit your preferences.
 			return err
 		}
 
-		err = apiCfg.Write()
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return apiCfg.Write()
 	},
 }
 
--- a/cmd/submit_test.go
+++ b/cmd/submit_test.go
@@ -76,11 +76,10 @@ func TestSubmit(t *testing.T) {
 		files := mf.File["files[]"]
 		for _, fileHeader := range files {
 			file, err := fileHeader.Open()
-			defer file.Close()
-
 			if err != nil {
 				t.Fatal(err)
 			}
+			defer file.Close()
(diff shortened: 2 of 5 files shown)


cmd/configure.go
```diff
@@ -62,12 +62,7 @@ You can also override certain default settings to suit your preferences.
 			return err
 		}
 
-		err = apiCfg.Write()
-		if err != nil {
-			return err
-		}
-
-		return nil
+		return apiCfg.Write()
```
nywilken: Good catch on theses 👍

cmd/submit_test.go
```diff
@@ -76,11 +76,10 @@ func TestSubmit(t *testing.T) {
 		files := mf.File["files[]"]
 		for _, fileHeader := range files {
 			file, err := fileHeader.Open()
-			defer file.Close()
-
 			if err != nil {
 				t.Fatal(err)
 			}
+			defer file.Close()
 			body, err := ioutil.ReadAll(file)
```
nywilken: I believe by calling defer after checking the error you run the risk of stopping execution without ever closing the file.

kytrinyx: If I understand the source code for`fileHeader.Open()` correctly, it calls `os.OpenFile()` (https://golang.org/src/os/file_unix.go?s=4493:4559#L137) behind the scenes, which will either return a file _or_ an error.

That means that if we do have an error, the file will be `nil`.

So it is safe to call `defer file.Close()` after dealing with the error.

kytrinyx:
> Disregard the comment on defer - it's an open so if it fails there is no file handle to call Close on.

D'oh I only saw this after commenting. Disregard my comment about disregarding... well, whatever :)

This looks great. Thanks @ferhatelmas

Issue #509, opened 2018-01-01
Install from src instructions


jungle-boogie:
Hi All,

As this is a go project, I suspect I can do some `go get github.com/exercism/cli`, but instead I get:
`no Go files in /home/jungle/gopath/src/github.com/exercism/cli`

What's the proper way to install from src?

Asking because I run openBSD and freeBSD, I don't see binaries of those available.

thanks!

ferhatelmas:
@jungle-boogie try `go get -u -t github.com/exercism/cli/exercism` where `-t` for test libraries if you want to hack on it.

jungle-boogie:
Thanks, @ferhatelmas! Perhaps these steps can be posted on the site for reference.

ferhatelmas:
@jungle-boogie Since generally people just want to get a binary and don't bother with its code, installation doc doesn't mention about it but it's already documented in [Contributing Guide](https://github.com/exercism/cli/blob/nextercism/CONTRIBUTING.md). However, we might make it clearer in readme.

jungle-boogie:
> generally people just want to get a binary

That's me, but as I said above - no freebsd or openBSD binary is available:
https://github.com/exercism/cli/releases/tag/v3.0.0-alpha.3

Thank you for pointing me to the contrib guideline, though.

Happy new year!

ferhatelmas:
@jungle-boogie My pleasure and thank you for your interest in exercism. Happy new year and hacking too!

@kytrinyx any idea for binaries in other platforms ?

kytrinyx:
I'm completely open to adding other binaries. The build is in https://github.com/exercism/cli/blob/nextercism/bin/build-all, if you want to submit a pull request.

PR #510, opened 2018-01-02
bin: add freebsd and openbsd releases

ferhatelmas:
related to #509.

### DIFF
--- a/bin/build-all
+++ b/bin/build-all
@@ -77,6 +77,14 @@ createRelease darwin amd64
 createRelease linux 386
 createRelease linux amd64
 
+# FreeBSD Releases
+createRelease freebsd 386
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks, @ferhatelmas. The build was failing on tip, but I think it was a simple timeout (I can't see a particular failure, and I can't imagine how adding these two architectures to the build would cause the tests to fail). I'm rerunning the CI build and will merge this as soon as it passes.

ferhatelmas:
@kytrinyx I guess there was a (kind of) tiny outage in travis last night. For example, it didn't pick up PRs for ~2 hours in go track.

kytrinyx:
Ah, that makes sense then. Thanks!

PR #514, opened 2018-03-01
Adding note to CONTRIBUTING about the dep package

chrisjlebron:
I was gearing up to contribute to the project and ran into this issue while following the steps to get a development environment set up locally. Just pointing to the package repo, which is where I ended up after a quick search.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@ If you don't care how and why and just want something that works, follow these s
 1. `git remote set-url origin https://github.com/<your-github-username>/cli`
 1. `go get -u github.com/golang/dep/cmd/dep`
 1. `dep ensure`
+    * you may need to [install `dep`](https://github.com/golang/dep) if it's not already on your system
 
 Then make the change as usual, and submit a pull request. Please provide tests for the changes where possible.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,8 @@ Exercism would be impossible without people like you being willing to spend time
 
 You'll need Go version 1.7 or higher. Follow the directions on http://golang.org/doc/install
 
+You will also need `dep`, the Go dependency management tool. Follow the directions on https://github.com/golang/dep
+
 ## Development
 
 If you've never contributed to a Go project before this is going to feel a little bit foreign.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -21,6 +21,7 @@ If you don't care how and why and just want something that works, follow these s
 1. `git remote set-url origin https://github.com/<your-github-username>/cli`
 1. `go get -u github.com/golang/dep/cmd/dep`
 1. `dep ensure`
+    * you may need to [install `dep`](https://github.com/golang/dep) if it's not already on your system
```
nywilken: Step 5 is actually the install step for `dep`, but it may not work for everyone's setup nor do we call it out as a dependency. With that I suggest we update step 5 include the link to install dep and update the dependencies section up top so that future contributors know it is a requirement for the development setup.

chrisjlebron: Ah yeah, was wondering why that step was there. With my naive setup the command doesn't install/build the package, just fetches the project (as far as I can tell?). I'll fix this based on your requests.

CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@ Exercism would be impossible without people like you being willing to spend time
 
 You'll need Go version 1.7 or higher. Follow the directions on http://golang.org/doc/install
 
+You will also need `dep`, the Go dependency management tool. Follow the directions on https://github.com/golang/dep
```
nywilken: Let's link directly to the install docs `https://golang.github.io/dep/docs/installation.html`

CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@ Exercism would be impossible without people like you being willing to spend time
 
 You'll need Go version 1.7 or higher. Follow the directions on http://golang.org/doc/install
```
nywilken: I know this was not part of your PR, but would you mind bumping this up to 1.8 or higher? Go v1.7 is no longer supported by dep so this document is not correct.

CONTRIBUTING.md
```diff
@@ -20,6 +22,7 @@ If you don't care how and why and just want something that works, follow these s
 1. `cd $GOPATH/src/github.com/exercism/cli` (or `cd %GOPATH%/src/github.com/exercism/cli` on Windows)
 1. `git remote set-url origin https://github.com/<your-github-username>/cli`
 1. `go get -u github.com/golang/dep/cmd/dep`
+   * depending on your setup, you may need to install `dep` by following the instructions in the [`dep` repo](https://github.com/golang/dep)
```
nywilken: I think this still may be a bit confusing for new contributors. The `go get` command might work, but when it doesn't we don't actually explicitly say that its okay to fail and proceed with the `dep` install steps. What do you think, seeing as you are updating the documentation for this exact reason? 

Do you think if we move this step up to `2.` and make it read like below it would've simplified things for you? 

```
1. [Install dep](https://golang.github.io/dep/docs/installation.html)
```

chrisjlebron: I see your point. I'm not sure what the setup needs to look like in order for the existing instruction to work as expected. For me the direct install from the `dep` link was necessary, but I'm not sure how many people would have the same experience.

What you've suggested (just including the link and forgetting about the `go get` command) is foolproof, so that change would make sense to me.

kytrinyx: I like the sound of something a bit more foolproof—getting stuff installed and configured is hard enough as it is 👍

kytrinyx:
Ah, we never got a ping about this being updated.

@nywilken I'm going to go go ahead and merge this, and we can consider further improvements later.

Thank you @chrisjlebron!

PR #519, opened 2018-03-25
iadded suppoert for linux with ppc64

villadalmine:
I have my go 1.8 in my g5 ppc64 linux ubuntu xenial and now I compile exercism from master source and it is working ok. I added some code to build-all script in order to build ppc64. I tested the script in my linux ppc64 and Crete all releases I do not know what happen  if you run the build-all in amd64..

### DIFF
--- a/bin/build-all
+++ b/bin/build-all
@@ -27,8 +27,12 @@ createRelease() {
 	if [ "$arch" = amd64 ]
 	then
 		osarch=64bit
-	else
+	elif [ "$os" = linux ] && [ "$arch" = ppc64 ]
+        then
+                osarch=ppc64
+        else
 		osarch=32bit
+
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks so much for this 🌷 

I look forward to pulling this down and trying it all out. I will take a look this weekend!

kytrinyx:
This appears to have built properly. I added a commit that triggers the ppc64 build alongside the others when we call the build script. I'll also port this back to the master branch.

Thank you!

villadalmine:
Cool!! I will check it when will be ready to download binary from master !!! Also I want to get support from ppc but 32 bit, but golang do not support ppc32 anymore and I think no one is interested to do that :(

PR #524, opened 2018-05-11
Make verbose flag global to all commands

kytrinyx:
We are going to want to add verbose logging to the entire package for
debugging purposes. This moves the local flag onto the root command as
a persistent flag so that all subcommands inherit it.

FYI @nywilken -- I'm hoping to do a fair amount of work on the nextercism CLI in the next few days. I'll be merging a lot myself, just to stay unblocked, but if you have a time to look at things and flag improvements that you'd like to see us make in subsequent PRs, that would be lovely.


nywilken:
@kytrinyx that sounds good. Thanks for merging the pending PRs. I just enabled notifications for all the exercism repos to keep an 👀out for future PRs. Great work at ROSSConf!

kytrinyx:
> I just enabled notifications for all the exercism repos

You are a brave, brave human being.

Issue #529, opened 2018-05-11
Handle 401 Unauthorized


kytrinyx:
When the API returns a [401 Unauthorized](https://httpstatuses.com/401) response, then we need to provide a useful error message asking them to check https://v2.exercism.io/my/settings and run the configure command to set the token.

nywilken:
Closed in PR #533

PR #533, opened 2018-05-15
Add 401 status code case

jpreese:
Handles #529 

One question I did have -- what is the origin of the current error messages? 401 is handled in the FETCH command currently with the message:

2018/05/14 20:51:13 unable to submit (HTTP: 401) - unknown api key 'key-from-config-here', please check http://exercism.io/account/key and reconfigure

and sort of exists with SUBMIT (something about please check your api key)

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -86,6 +86,8 @@ func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	case http.StatusInternalServerError:
 		// TODO: if it's json, and it has an error key, print the message.
 		return nil, fmt.Errorf("%s", res.Status)
+	case http.StatusUnauthorized:
+		return nil, fmt.Errorf("unauthorized request. Please run the configure command to set the api token found at https://v2.exercism.io/my/settings")
 	default:
 		if v != nil {
 			defer res.Body.Close()


nywilken:
@jpreese the current message is coming from the non nextercism build of the CLI on master https://github.com/exercism/cli/blob/master/api/api.go#L187

jpreese:
Awesome thanks. I see how it's split up now.

api/client.go
```diff
@@ -67,6 +67,8 @@ func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	switch res.StatusCode {
 	case http.StatusNoContent:
 		return res, nil
+	case http.StatusUnauthorized:
+		return nil, fmt.Errorf("unauthorized request. Please run the configure command to set the api token found at https://v2.exercism.io/my/settings")
```
kytrinyx: The URL here needs to be http://exercism.io/account/key since this is the v1 codebase

jpreese: @kytrinyx yep, I just realized master is for v1 and theres a branch specifically for nextercism

jpreese:
No problem @nywilken I just noticed the difference myself. To keep the same PR I just changed the base and updated my repository master brancg to point to the current state of the nextercism branch.

nywilken:
@jpreese now that the change has been merged you should sync up the respective branches in your fork to make the next contribution even easier. Thanks again.

Issue #554, opened 2018-06-13
Build fails on Travis for Go: Tip


FabienTregan:
The build on last go toolchain is not happy :

```debug/debug.go:50: Println arg list ends with redundant newline
debug/debug.go:71: Println arg list ends with redundant newline```

nywilken:
@FabienTregan thanks for opening up this issue. But it is a little difficult for me to understand what the issue is. For the future, while the line number is helpful could you provide reproduction steps along with a clear description of the issue so that we can better understand how to debug, fix, and test. We like to flag trivial issues as "good first patch" to encourage other members to contribute so having a clear description of the issue and reproduction steps is always helpful for new comers. 

All the best!

FabienTregan:
Step to reproduce:
* trigger a new build on github (eg. by making a new pull request

Observed result:
* The debug/debug.go file, which caused no problem until yesterday and is not changed by this PR, now makes the build fail on job "GO: tip" with the error message I provided in first comment (`debug/debug.go:71: Println arg list ends with redundant newline`) An exemple a build having this problem can be seen [here](https://travis-ci.org/exercism/cli/builds/391737571)

This seems to be coming from a change on the go linter or compiler, which previously only raised this error on fmt.Println but now also raises it on our own  function (which is also named Println).

The PR #555 allows the build to pass by remplacing the trailing "\n" with a new call to Println("").

kytrinyx:
Solved in #555

PR #555, opened 2018-06-13
Avoids ending Println parameter with new lines.

FabienTregan:
May solve issue #554


kytrinyx:
This does, indeed, solve the problem on Go: tip. Thank you @FabienTregan!

Issue #543, opened 2018-06-11
Tweak behavior of bare configure command


kytrinyx:
If you call `nextercism configure` with no flags or arguments, then we should be more helpful:

1. If you don't have a token configured, say "call this again with `--token`" and where you can find your token. For now this is https://v2.exercism.io/my/settings.
1. If you do have a token configured, then we should verify that it is valid. Tell them what the result is, and then output the full configuration.

PR #556, opened 2018-06-18
Token validation and configure improvements

### LINKED ISSUE #537: [nextercism] Add validation of auth token on configure step
In #536 there was an empty error message printed at the user after he has configured an obviously malformed auth-token via `nextercism c -t` (it was taken from v1 instead of v2).

To make sure such things do not happen again, I think we should add validation of the auth token.

Validating if it is a valid UUID should be mandatory.

I do consider a verification request to the API as optional (a user might be offline when rolling back automated configuration tasks). Its not sure to me though if we should make the test request opt-in or opt-out though.

ccallergard:
Intention:
Close #537 and #543.

Not quite satisfied, there are some cosmetic/edge-case/error-handling issues, but essentially works.
Gotta go, so won't go over it right now, but opening this for any thoughts.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -101,3 +101,16 @@ func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 
 	return res, nil
 }
+
+// checkAuthorization calls the API to check if
+// the client is authorized.
+func (c *Client) checkAuthorization() error {
+	url := c.APIConfig.URL("validate")
+	req, err := c.NewRequest("GET", url, nil)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


kytrinyx:
Thank you so much for tackling this! I will do a proper review this weekend. My immediate thought is that we can likely merge this with only a few small changes, and then we can look at further error handling and edge cases in a subsequent PR. 💙 💚 💙

kytrinyx:
I didn't notice that this doesn't update the tests. I'll take a stab at fixing those before I merge.

Issue #545, opened 2018-06-11
Rework `download` command to use `--exercise` flag


kytrinyx:
Make the exercise a flag rather than an argument to the command. Currently the download command has the following format:

```
nextercism download hello-world --track=ruby
```

Instead, use `--exercise` (shortflag `-e`)

```
nextercism download --exercise=hello-world --track=ruby
```

Since the default behavior will be to copy/paste from the site, the extra verbosity is not a problem.

We will add an enhancement for power users so there is less typing.

FabienTregan:
I started having a look at this one and came across that code, and think that the implemented behaviour is not the intended one (there are no test to confirm either version) :

The download command is built with `Args: cobra.ExactArgs(1)`, hence if we give no args in addition to flags, the program has already stopped displaying `Error: accepts 1 arg(s), received 0` and `downloadCmd.RunE` is never called.

But then in `downloadCmd.RunE` we have code that seems to handle the case where the exercism name is not provided :

* Displaying a more explicit error message
https://github.com/exercism/cli/blob/eb19ecd50ea27b7943b42f768227527ce4878f82/cmd/download.go#L47-L50

* Only reading the exercism name if there are any argument :
https://github.com/exercism/cli/blob/eb19ecd50ea27b7943b42f768227527ce4878f82/cmd/download.go#L78-L81

And some the only code that actually uses the exercise name is ignored if an UUID is provided :
https://github.com/exercism/cli/blob/eb19ecd50ea27b7943b42f768227527ce4878f82/cmd/download.go#L83-L90

There is no test that can invalidate either behaviour. A possible implementation would be :
* `--uuid` is forbidden if `--exercise` is used, `latest` is then used as slug and exercise can't be empty
* `--uuid` is mandatory and can't be empty if `--exercise` is not used, `uuid` is then used as slug

kytrinyx:
> The download command is built with Args: cobra.ExactArgs(1)

We would need to get rid of that.

`--uuid` and `--exercise` are mutually exclusive. One of them must be present.

I don't understand the difference between the two bullet points that you have, they seem to be saying the same thing in two ways.

strickinato:
Does it make sense to close this since #563 was merged? Or is there more ways I can be helpful here?

PR #563, opened 2018-06-22
Rework download command to take exercise flag

strickinato:
Hi! This very minimally solves https://github.com/exercism/cli/issues/545

This is my first contribution (and actually the first time I've ever touched go code) - so I hope it's alright.

### The fix:

* Changed the command to take ONLY flags
* Check for the exercise flag instead of relying on the arg
* Lifted the check for exercise name to the beginning of the function so it can thow an error early if there is no exercise name OR uuid

### About me:

I've played very in a very minor way with exercism a while back and thought it was awesome (but not immediately useful to me).

A good friend of mine recently started programming and I showed her exercism. She immediately connected with it and is having an awesome experience which reinvigorated my excitement and I thought I would help where I can! (though I have no go experience)

❤️ Hope this helps and thanks for the amazing community you've created! ❤️

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -29,7 +29,7 @@ latest solution.
 
 Download other people's solutions by providing the UUID.
 `,
-	Args: cobra.ExactArgs(1),
+	Args: cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		token, err := cmd.Flags().GetString("token")
 		if err != nil {
@@ -45,9 +45,12 @@ Download other people's solutions by providing the UUID.
 		if err != nil {
 			return err
 		}
-		if uuid == "" && len(args) == 0 {
-			// TODO: usage
-			return errors.New("need an exercise name or a solution --uuid")
+		exercise, err := cmd.Flags().GetString("exercise")
+		if err != nil {
+			return err
+		}
+		if uuid == "" && exercise == "" {
+			return errors.New("need an --exercise name or a solution --uuid")
 		}
 		usrCfg, err := config.NewUserConfig()
 		if err != nil {
@@ -75,10 +78,6 @@ Download other people's solutions by providing the UUID.
 		track, err := cmd.Flags().GetString("track")
 		if err != nil {
 			return err
-		}
-		var exercise string
-		if len(args) > 0 {
-			exercise = args[0]
 		}
 
 		if uuid == "" {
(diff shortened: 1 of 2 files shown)


kytrinyx:
❤️ 💛 💚 💙 💜 🖤 Thank you! I'm going to take the time to review this properly on the weekend, I just wanted to say thanks so much for tackling it (and in an unfamiliar language, even!).

cmd/download.go
```diff
@@ -28,7 +28,7 @@ latest solution.
 
 Download other people's solutions by providing the UUID.
 `,
-	Args: cobra.ExactArgs(1),
+	Args: cobra.ExactArgs(0),
```
nywilken: @strickinato thanks for contribution and welcome. @kytrinyx will give a through review of the changes, but I will leave a note on this line. We should get rid of the call to ExactArgs as it adds its own error messaging which is not helpful to the user. Maybe showing the default help message? What do you think your make sense?

kytrinyx: I think we should show a specific error message rather than the default help if you've called the command with no arguments.

cmd/download.go
```diff
@@ -210,6 +210,7 @@ type downloadPayload struct {
 func initDownloadCmd() {
 	downloadCmd.Flags().StringP("uuid", "u", "", "the solution UUID")
 	downloadCmd.Flags().StringP("track", "t", "", "the track ID")
+	downloadCmd.Flags().StringP("exercise", "e", "", "the exercise ID")
```
kytrinyx: We typically refer to the value we're looking for as the _exercise slug_.

strickinato:
Thanks for the feedback!

I'm expecting to get to this in the late morning (EST) tomorrow.

On quick glance the conflict doesn't look bad - so I should be able to tackle without too much problem.

strickinato:
I think this should be good to go according to your feedback.

I am slightly concerned that I put the `CommandTest.Setup` and `CommandTest.Teardown` etc. within a loop. It's probably slower than it needs to be - but I wasn't sure a better way to test both cases.

Let me know if there is more I can do!

strickinato:
@kytrinyx Just a friendly bump in case you missed the updates (don't want it to go stale).

kytrinyx:
Oh! Yes I completely missed this. I'll get this merged today. Thank you ❤️

kytrinyx:
> I am slightly concerned that I put the CommandTest.Setup and CommandTest.Teardown etc. within a loop. It's probably slower than it needs to be - but I wasn't sure a better way to test both cases.

I'm going to be reworking the commands shortly to make it so we don't have to rely on the setup/teardown for the entire command, but rather inject the things that we care about.

PR #581, opened 2018-06-25
Move API error handling into the commands

kytrinyx:
I realized that we were unnecessarily complicating things by pushing some of the error handling into the global API wrapper, whereas it's in each of the commands that we'll know how to handle individual error types specific to the command that the API returns.

This is a pretty small change, but I'd appreciate your feedback on it, @nywilken, if you have time.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -95,9 +95,9 @@ Download other people's solutions by providing the UUID.
 			return err
 		}
 
-		payload := &downloadPayload{}
+		var payload downloadPayload
 		defer res.Body.Close()
-		if err := json.NewDecoder(res.Body).Decode(payload); err != nil {
+		if err := json.NewDecoder(res.Body).Decode(&payload); err != nil {
 			return fmt.Errorf("unable to parse API response - %s", err)
 		}
 
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -90,7 +90,7 @@ Download other people's solutions by providing the UUID.
 			req.URL.RawQuery = q.Encode()
 		}
 
-		res, err := client.Do(req, nil)
+		res, err := client.Do(req)
 		if err != nil {
 			return err
 		}
@@ -152,7 +152,7 @@ Download other people's solutions by providing the UUID.
 				return err
 			}
 
-			res, err := client.Do(req, nil)
+			res, err := client.Do(req)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


nywilken:
> I realized that we were unnecessarily complicating things by pushing some of the error handling into the global API wrapper, whereas it's in each of the commands that we'll know how to handle individual error types specific to the command that the API returns.

I agree with this approach. Handling the errors when they ocurr will definitely simplify things. I’ll give this a further review in the morning.

cmd/download.go
```diff
@@ -94,12 +95,22 @@ Download other people's solutions by providing the UUID.
...
 		}
 
-		payload := &downloadPayload{}
-		res, err := client.Do(req, payload)
+		res, err := client.Do(req)
 		if err != nil {
 			return err
 		}
 
+		payload := &downloadPayload{}
```
nywilken: Minor nitpick. Instead of initializing with an a pointer to a downloadPayload I would recommend using `var payload downloadPayload` and then in the call to Decode take the address, as that is the only place we actually need the pointer.

cmd/prepare.go
```diff
@@ -63,14 +64,20 @@ func prepareTrack(id string) error {
...
 
-	payload := &prepareTrackPayload{}
-	res, err := client.Do(req, payload)
+	res, err := client.Do(req)
 	if err != nil {
 		return err
 	}
+	defer res.Body.Close()
+
+	payload := &prepareTrackPayload{}
```
nywilken: ditto

cmd/prepare.go
```diff
@@ -63,14 +64,20 @@ func prepareTrack(id string) error {
...
+
+	if err := json.NewDecoder(res.Body).Decode(payload); err != nil {
+		return fmt.Errorf("unable to parse API response - %s", err)
+	}
 
 	if res.StatusCode != http.StatusOK {
-		return errors.New("api call failed")
+		return errors.New(payload.Error.Message)
```
nywilken: Without knowing too much about the API response. If for whatever reason Error is empty, or not provided in the response we will return an error with no message. If this is a non issue then we can ignore this observation

cmd/download.go
kytrinyx: Sweet, I'll update that.

cmd/prepare.go
kytrinyx: For now I'm ok with ignoring it. We're going to have to go through all the error handling and messaging with a fine-tooth comb, but I'd like to get the basic shape in place so that we have something reasonable to build off of.

PR #589, opened 2018-06-25
[WIP] Implement viper persister

kytrinyx:
🚧 🚧 🚧 work in progress 🚧 🚧 🚧 

Here's the direction I'm thinking of going in, based on some comments from @FabienTregan a few days ago.

The idea is to get rid of the various Config types in the `config` package, operating with a viper persister interface that writes to files when in production mode, and stores in memory when in test mode.

The cleanest way that I can see to move forward here is to set the persister we need as a package variable on `config`. That means that we would get in trouble if the config package is called from more than one package, because... well... race conditions. But I think we'll be fine as long as it's only used from the config package. Maybe. Unless I've forgotten how the parallelization of tests works in Go, which is likely. If we can't do a package variable the only other thing I can think of is to have the `RunE` function load the viper config and then immediately pass that on to another function, which would be the one we test. I don't much like it though.

As you can tell, I'm not yet sure what the shape of everything needs to be, that will be easier to experiment with once we've merged the open refactoring PRs which reduce some of the unnecessary indirection and abstractions.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -22,13 +22,28 @@
 [[projects]]
   branch = "master"
   name = "github.com/hashicorp/hcl"
-  packages = [".","hcl/ast","hcl/parser","hcl/scanner","hcl/strconv","hcl/token","json/parser","json/scanner","json/token"]
+  packages = [
+    ".",
+    "hcl/ast",
+    "hcl/parser",
+    "hcl/printer",
+    "hcl/scanner",
+    "hcl/strconv",
+    "hcl/token",
+    "json/parser",
+    "json/scanner",
+    "json/token"
+  ]
   revision = "392dba7d905ed5d04a5794ba89f558b27e2ba1ca"
 
 [[projects]]
   branch = "master"
   name = "github.com/inconshreveable/go-update"
-  packages = [".","internal/binarydist","internal/osext"]
+  packages = [
+    ".",
+    "internal/binarydist",
+    "internal/osext"
+  ]
   revision = "8152e7eb6ccf8679a64582a66b78519688d156ad"
 
 [[projects]]
@@ -70,7 +85,10 @@
 [[projects]]
   branch = "master"
   name = "github.com/spf13/afero"
-  packages = [".","mem"]
+  packages = [
+    ".",
+    "mem"
+  ]
   revision = "9be650865eab0c12963d8753212f4f9c66cdcf12"
 
 [[projects]]
@@ -101,7 +119,7 @@
   branch = "master"
   name = "github.com/spf13/viper"
   packages = ["."]
-  revision = "25b30aa063fc18e48662b86996252eabdcf2f0c7"
+  revision = "15738813a09db5c8e5b60a19d67d3f9bd38da3a4"
 
 [[projects]]
   name = "github.com/stretchr/testify"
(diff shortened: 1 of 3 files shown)


FabienTregan:
Another way I wanted to give a try is to inject the ConfigurationPersister (or the Configuration, which would aggregate the persister *). I think we can locate the places that will need the Persister by searching the calls to Dir().

Pros:
-No race condition
-More readable code (injection is made from main() or when the System Under Test in instanciated, no need to rely on side effect)

Cons:
-Instanciation of commands and adding them to rootCmd should then be done from main() instead of command's init() (then rootCmd is no longer a package variable. I like this pattern of having all the 'plumbing' code in main, but it may require more work).

I'm not sure you'll get what I'd like to explain, maybe I can make a proto this evening (14 hours from now). It will, at least, be a good coding exercise for me :)

kytrinyx:
> An alternative to package variable persister I wanted to give a try is to inject the ConfigurationPersister 

That was where I was headed first, but I could not find a place I like to inject it.

> I think we can locate the places that will need the Persister by searching the calls to Dir().

I'm trying to narrow down the places that call Dir() to exactly one.

> Instanciation of commands and adding them to rootCmd should then be done from main() instead of command's init()

I'd probably need to see what it looks like before having a proper opinion about it, but I resist the idea for two reasons:

1. It separates the place where the command is defined from where it is added, making it easier to forget to add it and harder to notice that one has forgotten
1. It moves away from the pattern that people are used to, which means that you have to think harder just to realize that there's nothing special going on.

I'd be interested in seeing what you have in mind.

kytrinyx:
After a bit more experimenting, I think what I like the most so far is this pattern:

```
package cmd

import (
	"fmt"

	"github.com/exercism/cli/config"
	"github.com/spf13/cobra"
	"github.com/spf13/viper"
)

// workspaceCmd outputs the path to the person's workspace directory.
var workspaceCmd = &cobra.Command{
	Use:     "workspace",
	Aliases: []string{"w"},
	Short:   "Print out the path to your Exercism workspace.",
		v, err := config.LoadViperConfig(config.TypeUser)
		if err != nil {
			return err
		}
		workspaceRun(v)
		return nil
	},
}

func runWorkspace(v *viper.Viper) {
	fmt.Fprintf(Out, "%s\n", v.GetString("workspace"))
}

func init() {
	RootCmd.AddCommand(workspaceCmd)
}
```

And the test:

```
package cmd

import (
	"bytes"
	"testing"

	"github.com/spf13/viper"
	"github.com/stretchr/testify/assert"
)

func TestWorkspace(t *testing.T) {
	Out = new(bytes.Buffer)

	v := viper.New()
	v.Set("workspace", "/home/alice")

	runWorkspace(v)

	b := Out.(*bytes.Buffer)
	assert.Equal(t, string("/home/alice\n"), b.String())
}
```

kytrinyx:
Ok, I pushed up a version that injects a viper configuration directly into the workspace command:

- workspace command: https://github.com/exercism/cli/blob/85dea083afd61a49d3b904433a96d3d532f2ac35/cmd/workspace.go
- tests: https://github.com/exercism/cli/blob/85dea083afd61a49d3b904433a96d3d532f2ac35/cmd/workspace_test.go

Issue #551, opened 2018-06-11
Clarify behavior of submit command


kytrinyx:
There will be three ways to call `submit`:

1. **Using a relative or absolute path to a directory**  (see #550)
    `exercism submit DIR`
1. **Passing a list of files** (see #549)
    `exercism submit --files`
1. **Passing an `--exercise`, `--track`, and `--team` to infer directory** (see #547 and #548)

These are incompatible. If someone calls something that seems to be trying to do two of these at the same time, we need to exit with a helpful error message.

kytrinyx:
Fixed in #594

Issue #550, opened 2018-06-11
Rework submit command to take simple DIR argument


kytrinyx:
Right now the `submit` command tries to be too smart, and to let you specify an argument which might be either a directory, or a file, or an exercise. Instead, that argument should _only_ be a relative or absolute path to a directory.

If we get no directory argument, then the `--exercise` and `--track` flags must be passed.

In other words, make it so that the submit command accepts:

```
exercism submit DIR
```

where DIR is the absolute or relative path to a directory within the workspace.

kytrinyx:
Fixed in #594

Issue #549, opened 2018-06-11
Rework submit to explicitly take --files flag


kytrinyx:
The default behavior of `submit` is to accept an `--exercise` and `--track` flag (and optionally `--team`). This will submit all the exercise files within the exercise directory. If someone wants to only submit some of them, they need to list out the files explicitly.

Currently the submit command accepts files as arguments. This should be changed to explicitly be a `--files` flag (shortflag `-f`).

If the files are not all for the same exercise, we need to exit with an error.

aldraco:
I have been looking at this group of related tickets around `submit` and would like to start working on them -- I'll go ahead and claim this one, if nobody else has yet.

kytrinyx:
@aldraco Thank you 🌻!

aldraco:
status update -- `go test` passes for the 3 new behaviors :fireworks: but it still needs a bit of cleanup, clarification, etc. Also, since #568 came up in the last few days, want to see if it's possible to check that the issue won't affect this command as well.

kytrinyx:
Fixed in #594

Issue #547, opened 2018-06-11
Rework submit command to take `--exercise` and `--track`


kytrinyx:
The default UI of `submit` should mirror that of `download`:

```
exercism submit --exercise=hello-world --track=ruby
```

Shortflags: `-e` for exercise, `-t` for track.

The CLI should read in the list of all the files in the tree under that exercise, and should provide an interactive prompt that shows:

1. the directory that corresponds to the exercise and track
1. the list of all the files that will be submitted (relative to the directory)
1. "to submit, hit enter"
1. the command to use (copy/paste) to submit individual files in case they don't want to submit everything

E.g.:

```
exercism submit --files path1 path2 path3
```

kytrinyx:
Fixed in #594

PR #594, opened 2018-07-02
Standardize submit command (#550, #551, #547, #549)

aldraco:
This PR should address  #550, #551, #547 and #549 by conforming the `submit` command to the following behaviors:
 - submit using a directory
 - submit using a combination of exercise/track flags
 - submit a list of files.
 - adds new tests for all 3 behaviors.

(It does not handle the --team feature requested in a separate ticket.)

The new functionality was also smoke tested against mentorcism-beta by reconfiguring the client to point to that, and submitting Hello World for a new language.

Thank you for the opportunity to contribute! 

ETA: Just noticed this comment on #547, which is not addressed here:
> (... should print out) the command to use (copy/paste) to submit individual files in case they don't want to submit everything

Happy to add it in afterwards, or in another PR.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -38,7 +38,7 @@ figuring things out if necessary.
 `,
 	Args: cobra.MaximumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		// check input before doing any other work
+		// Validate input before doing any other work
 		exercise, err := cmd.Flags().GetString("exercise")
 		if err != nil {
 			return err
@@ -88,7 +88,7 @@ figuring things out if necessary.
 
 		ws := workspace.New(usrCfg.Workspace)
 
-		// create directory from track and exercise slugs if needed
+		// Create directory from track and exercise slugs if needed
 		if trackId != "" && exercise != "" {
 			args = []string{filepath.Join(ws.Dir, trackId, exercise)}
 		} else if len(files) > 0 {
@@ -99,10 +99,12 @@ figuring things out if necessary.
 		if err != nil {
 			return err
 		}
+
 		dirs, err := ws.Locate(tx.Dir)
 		if err != nil {
 			return err
 		}
+
 		sx, err := workspace.NewSolutions(dirs)
 		if err != nil {
 			return err
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -44,7 +44,7 @@ figuring things out if necessary.
 			return err
 		}
 
-		trackId, err := cmd.Flags().GetString("track")
+		trackID, err := cmd.Flags().GetString("track")
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


cmd/submit.go
```diff
@@ -36,23 +36,45 @@ If called with the name of an exercise, it will work out which
...
 	RunE: func(cmd *cobra.Command, args []string) error {
-		usrCfg, err := config.NewUserConfig()
+		// Validate input before doing any other work
+		exercise, err := cmd.Flags().GetString("exercise")
 		if err != nil {
 			return err
 		}
 
-		cliCfg, err := config.NewCLIConfig()
+		trackId, err := cmd.Flags().GetString("track")
```
kytrinyx: In Go the idiom is to use all caps for ID in variable names, so this should be `trackID`.

cmd/submit.go
```diff
@@ -36,23 +36,45 @@ If called with the name of an exercise, it will work out which
...
+			return errors.New("must use --exercise and --track together")
+		}
+
+		if len(args) > 0 && (exercise != "" || trackId != "" || len(files) > 0) {
+			return errors.New("can't use flags and directory arguments together")
+		}
+
+		if len(files) > 0 && len(args) > 0 {
+			return errors.New("can't submit files and a directory together")
+		}
```
kytrinyx: Maybe to clarify:

> You can submit either a list of files or a whole directory, but not both.

cmd/submit.go
```diff
@@ -36,23 +36,45 @@ If called with the name of an exercise, it will work out which
...
+			return err
+		}
+
+		// Verify correct usage of flags/args
+		if len(args) == 0 && len(files) == 0 && !(exercise != "" && trackId != "") {
+			return errors.New("must use --exercise and --track together")
+		}
+
+		if len(args) > 0 && (exercise != "" || trackId != "" || len(files) > 0) {
+			return errors.New("can't use flags and directory arguments together")
```
kytrinyx: the combination of files and directories is handled below, so we can simplify this one and remove the part where we're checking files here.

For the error message, maybe something like:

> You are submitting a directory. We will infer the track and exercise from that. Please re-run the command again without the flags.

cmd/submit.go
```diff
@@ -36,23 +36,45 @@ If called with the name of an exercise, it will work out which
...
-			args = []string{cwd}
+		files, err := cmd.Flags().GetStringSlice("files")
+		if err != nil {
+			return err
+		}
+
+		// Verify correct usage of flags/args
+		if len(args) == 0 && len(files) == 0 && !(exercise != "" && trackId != "") {
+			return errors.New("must use --exercise and --track together")
+		}
```
kytrinyx: Let's be explicit here, and tell them what's missing.

> You specified --exercise. Please specify the --track to submit.

And

> You specified --track. Please specify the --exercise as well.

kytrinyx:
@aldraco I have rebased this against the `nextercism` branch and fixed the conflict. I pushed this up to your branch.

Would you please do  the following?
```
git checkout submit_cmd_standardize
git fetch origin submit_cmd_standardize
git reset --hard origin/submit_cmd_standardize
git rebase -i --exec "go test ./..." a70e30caeb45baae8609367782b1fc8f5c8dfa82
```

And then fix any of the failing tests in the commits?

kytrinyx:
Actually, now that I read your commit history, I see that I might be able to just squash two of your commits. Let me try that!

PR #595, opened 2018-07-05
Simplify configuration logic

kytrinyx:
I pulled some of the things I like from the spike out into separate commits, and applied them here.

All of it boils down to simplifying the configuration logic:

- remove the unmarshaling from client.Do(), since we want more control over the error handling and error messages
- get rid of the API config, folding in the base url into the user config
- pass token and base url directly into the API client, as they don't actually need any other configuration

The bottom line is that now the _only_ package that imports the `config` package is the `cmd` package, which will simplify a few things going forward.

FYI @nywilken I'm going to go ahead and merge this so that I can work on top of it.

### DIFF
--- a/cmd/configure_test.go
+++ b/cmd/configure_test.go
@@ -18,28 +18,41 @@ type testCase struct {
 func TestConfigure(t *testing.T) {
 	testCases := []testCase{
 		testCase{
-			desc:           "It writes the flags when there is no config file.",
-			args:           []string{"fakeapp", "configure", "--token", "a", "--workspace", "/a", "--api", "http://example.com", "--skip-auth"},
+			desc: "It writes the flags when there is no config file.",
+			args: []string{
+				"fakeapp", "configure", "--skip-auth",
+				"--token", "abc123",
+				"--workspace", "/workspace",
+				"--api", "http://api.example.com",
+			},
 			existingUsrCfg: nil,
-			expectedUsrCfg: &config.UserConfig{Token: "a", Workspace: "/a", APIBaseURL: "http://example.com"},
+			expectedUsrCfg: &config.UserConfig{Token: "abc123", Workspace: "/workspace", APIBaseURL: "http://api.example.com"},
 		},
 		testCase{
-			desc:           "It overwrites the flags in the config file.",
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/spf13/viper"
 )
 
+var (
+	defaultBaseURL = "https://v2.exercism.io/api/v1"
```
nywilken: I find this to be a bit confusing seeing as config/config is a generic wrapper around config with no specific exercism data. In my opinion it is probably better off in `user_config.go`

cmd/configure_test.go
```diff
@@ -32,11 +32,11 @@ func TestConfigure(t *testing.T) {
 		testCase{
 			desc:           "It overwrites the flags that are passed, without losing the ones that are not.",
 			args:           []string{"fakeapp", "configure", "--token", "c", "--skip-auth"},
-			existingUsrCfg: &config.UserConfig{Token: "token-c", Workspace: "/workspace-c"},
-			expectedUsrCfg: &config.UserConfig{Token: "c", Workspace: "/workspace-c"},
+			existingUsrCfg: &config.UserConfig{Token: "token-c", Workspace: "/workspace-c", APIBaseURL: "http://example.com"},
```
nywilken: For a second I thought the APIBaseURL was some how being obtained from a previous testcase using `http://example.com`. Maybe having a workspace and APIBaseURL that is different from other test cases or something like `Workspace: "/unmodified-workspace-path", APIBaseURL: "http://unmodified.example.com/"` would help illustrate the point. Or a description like "Unspecified flags do not overwrite previously configured settings"

kytrinyx:
> The PR description seems like the changes should be doing more, but guessing you might still be working on it.

Yes! I just wanted to get in the basics before it becomes massively difficult to review.

config/config.go
kytrinyx: Yepp, that's a good point. I'll make that change.

cmd/configure_test.go
kytrinyx: Nice, yeah, that would be much clearer.

kytrinyx: I went ahead and did a complete overhaul of the test data values in that test to help clarify intention throughout. I like it a lot better! Thanks for the suggestion.

nywilken: Looks good!

PR #596, opened 2018-07-07
Tweak config track test for readability

kytrinyx:
When naming a method that contains a small conditional, I find it easier
to name if the method contains both the exception and the rule.

The notIfNeeded method is naming only the case where it's not needed.
This changes the logic so that it contains two variations on the same
concept ("how acceptable is this").

In the second commit I differentiate between the test name and the failure message, as we don't need the same message in both places.


kytrinyx:
FYI @nywilken I'm going to merge this so that I can build on top of it.

PR #597, opened 2018-07-08
Use filepath rather than path for manipulating paths

### LINKED ISSUE #570: [Mentorcism] Windows: Configuration Failing
When I try to configure the token, I get a `Error: CreateFile` error. When I execute `mentorcism -s`:

```
$ mentorcism configure -s

Config dir:      C:\Users\jake\AppData\Roaming\C:\Exercism\mentorcism.exe
-t, --token
-w, --workspace  C:\Exercism\mentorcism.exe
-a, --api        https://mentors-beta.exercism.io/api/v1
```

OS: Windows 10
Shell: Git Bash

kytrinyx:
The documentation for the path package says:

> The path package should only be used for paths separated by forward
> slashes, such as the paths in URLs. This package does not deal with
> Windows paths with drive letters or backslashes; to manipulate operating
> system paths, use the path/filepath package.

Closes #570

### DIFF
--- a/config/dir.go
+++ b/config/dir.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"os"
-	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


kytrinyx:
FYI @nywilken -- merging so I can build on top of it (as always, if you see anything that I should fix, add feedback and I'll submit additional PRs to do that).

nywilken:
@kytrinyx nice catch. I really need to start reading the documentation more. I’ll keep my eyes open for those cases. With that said we should replace the string concatenation with a filepath.Join in user_config.
 
https://github.com/exercism/cli/blob/66e605e248c6d3d4ae706eb63822681681c60591/config/user_config.go#L75

PR #599, opened 2018-07-08
Fix token validation logic

kytrinyx:
The configure command was calling the token validation endpoint, but
was not actually bailing or providing an error in the case of an invalid
token.

We will need to improve the error message to include the link to the
url where the token can be found. Also, we should probably not print
the current configuration after printing the error message.

I am working on improvements to the configure command in a separate branch, and will make those changes there.

I have tested this by building the client and configuring with both an invalid and a valid token. In the new configure branch I am adding tests for the token validation.

FYI @nywilken I will merge this in order to unblock the configuration improvements. I don't think it's worth looking at this PR too closely; I mainly wanted to fix the api/client.go separately from the improvements to configure.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -72,14 +72,16 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, nil
 }
 
-// ValidateToken calls the API to determine whether the token is valid.
-func (c *Client) ValidateToken() error {
+// TokenIsValid calls the API to determine whether the token is valid.
+func (c *Client) TokenIsValid() (bool, error) {
 	url := fmt.Sprintf("%s/validate_token", c.APIBaseURL)
 	req, err := c.NewRequest("GET", url, nil)
 	if err != nil {
-		return err
+		return false, err
 	}
-	_, err = c.Do(req)
-
-	return err
+	resp, err := c.Do(req)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -72,14 +72,16 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, nil
 }
 
-// ValidateToken calls the API to determine whether the token is valid.
-func (c *Client) ValidateToken() error {
+// TokenIsValid calls the API to determine whether the token is valid.
+func (c *Client) TokenIsValid() (bool, error) {
 	url := fmt.Sprintf("%s/validate_token", c.APIBaseURL)
```
nywilken: This actually might be a good place to use `path.Join(c.APIBaseURL, validate_token)` :laughing:

kytrinyx: Hah! Now the distinction between `path` and `filepath` is starting to make sense.

kytrinyx: with path.Join it does this: (removes the // after the protocol)
```
http:/127.0.0.1:61851/validate_token
```

nywilken: :man_facepalming: so much for this statement 
> The path package should only be used for paths separated by forward slashes, such as the paths in URLs.

Sorry for that suggestion without actually testing.

PR #602, opened 2018-07-08
Do not double-print the command errors

kytrinyx:
I somehow was getting no errors, then when I printed them I got double errors,
then going back to how it was before, I'm getting exactly one error printed
... which is what I wanted in the first place.

I have no idea what I was doing.


kytrinyx:
> `fmt.Fprintln(...)`

Ah, yes! I think once I've gotten through the entire overhaul we should probably go through and fine tune all these little things.

PR #604, opened 2018-07-08
Rework the configure command to be more robust and testable

### LINKED ISSUE #542: Do not clobber existing directories when configuring workspace dir
When configuring the `nextercism` CLI for the first time (e.g. when `workspace_dir` in the config is blank), then we should verify that the workspace directory **does not exist** before setting a default. If it does, then we should exit with an error message saying that the directory already exists, and that if that's the correct directory, please call the command again, explicitly adding --directory="...".

We can print the full command for them to copy (including the token, which they just gave us).

### LINKED ISSUE #567: Consider rethinking design of configuration
From @FabienTregan:

### Things that were not that easy to understand when you first read the code 

* Relation between `Config` and `ViperConfig`

Currently when creating a `UserConfig`, we :
- Instantiate a `NewEmptyUserconfig()` (which is not really empty because it contains `Config: New(Dir(), "user")`)
- `cfg.Load` this `UserConfiguration` with a `viper.New()` (empty viper config)
- `cfg.Load` copies from `cfg` some of the config to the viper cfg (the one needed to load files) (`cfg.readIn()`)
- the viper cfg is loaded
- the viper cfg is `Unmarshaled()` back into `cfg`

Could we just have `ViperConfigPersister` (implementing a `ConfigPersister` interface) that would read or write the viper config from default location? Then for unit tests, we could just inject an `InMemoryConfigPersister` and be decoupled form ios/linux/windows file system difference.

---

I've been really unhappy with the current design of the configuration logic and this mess between viper and the various configs we rely on, but hadn't had the bandwidth to figure out how to untangle it.

I think this approach would make so much sense!

kytrinyx:
The configure command was a bit gnarly, and the test we had was testing everything at once, which made it fairly difficult to reason about.

This makes several related but fairly extensive changes:

- It adds a level of indirection by introducing a `runConfigure` function which takes a configuration object and the flags, but does not in any way deal with the Cobra command itself. The configuration object is new. By default, it will set a bunch of defaults, but since we're passing it to `runConfigure`, the tests can hard-code any of the defaults and things that we usually determine via the environment and OS.
- It uses the Out and Err streams more deliberately, and ensures that the tests do not print to them.
- It adds separate tests for each configuration option, so that all the logic related to each flag can be reasoned about separately.
- It removes all mention of the `config.UserConfig` type, which I am going to be working towards getting rid of. Instead, it uses the viper config directly, and adds a thin persistence wrapper to it, allowing us to avoid the filesystem for most of the tests.
- It fleshes out the verification logic for the `--no-verify` flag to verify both the base API URL (API is reachable) and the token (can successfully authenticate)

Additionally, this ensures that if we set a default workspace, that we do not set it to an existing directory. If there is something at the default workspace location, then we will print an error explaining that if they mean it, they can pass `--workspace` explicitly. We print the current command with the additional workspace in that case.

**Note: the new tests are quite repetitive. I did not want to introduce an abstraction here yet, as I'm afraid it will obscure more than it helps. I'm open to experimenting a bit with DRYing it up a bit, but I do not want to do it this week, as it's not crucial to moving the project forward.**

This pull request clarifies the behavior of the command in a number of ways:

If you just run `exercism configure` with no flags, then it will tell you how to configure the token if you haven't yet.

```
$ ./testercism configure
Error: There is no token configured. Find your token in your settings on the website, and call this command again with --token=<your-token>.
```

If you explicitly pass `--show` we will ignore any other flags, and just show you what is _currently_ configured.

```
$ ./testercism configure --show

Config dir:      /Users/kytrinyx/.config/testercism
-t, --token      11111111-2222-3333-4444-555555555555
-w, --workspace  /Users/kytrinyx/testercism
-a, --api        https://v2a.exercism.io/api/v1
```  

If you get an error, we will not print out the existing config:
```
$ ./testercism configure --token=abc123
Error: The token 'abc123' is invalid. Find your token on https://v2.exercism.io/my/settings.
```

If you successfully configure something, it will print the configuration (even if you don't pass --show).

```
$ ./testercism configure --token=11111111-2222-3333-4444-555555555555

Config dir:      /Users/kytrinyx/.config/testercism
-t, --token      11111111-2222-3333-4444-555555555555
-w, --workspace  /Users/kytrinyx/testercism
-a, --api        https://v2.exercism.io/api/v1
```    

Closes #567
Closes #542

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -22,13 +22,28 @@
 [[projects]]
   branch = "master"
   name = "github.com/hashicorp/hcl"
-  packages = [".","hcl/ast","hcl/parser","hcl/scanner","hcl/strconv","hcl/token","json/parser","json/scanner","json/token"]
+  packages = [
+    ".",
+    "hcl/ast",
+    "hcl/parser",
+    "hcl/printer",
+    "hcl/scanner",
+    "hcl/strconv",
+    "hcl/token",
+    "json/parser",
+    "json/scanner",
+    "json/token"
+  ]
   revision = "392dba7d905ed5d04a5794ba89f558b27e2ba1ca"
 
 [[projects]]
   branch = "master"
   name = "github.com/inconshreveable/go-update"
-  packages = [".","internal/binarydist","internal/osext"]
+  packages = [
+    ".",
+    "internal/binarydist",
+    "internal/osext"
+  ]
   revision = "8152e7eb6ccf8679a64582a66b78519688d156ad"
 
 [[projects]]
@@ -70,7 +85,10 @@
 [[projects]]
   branch = "master"
   name = "github.com/spf13/afero"
-  packages = [".","mem"]
+  packages = [
+    ".",
+    "mem"
+  ]
   revision = "9be650865eab0c12963d8753212f4f9c66cdcf12"
 
 [[projects]]
@@ -101,7 +119,7 @@
   branch = "master"
   name = "github.com/spf13/viper"
   packages = ["."]
-  revision = "25b30aa063fc18e48662b86996252eabdcf2f0c7"
+  revision = "15738813a09db5c8e5b60a19d67d3f9bd38da3a4"
 
 [[projects]]
   name = "github.com/stretchr/testify"
@@ -112,7 +130,11 @@
 [[projects]]
   branch = "master"
   name = "golang.org/x/net"
-  packages = ["html","html/atom","html/charset"]
+  packages = [
+    "html",
+    "html/atom",
+    "html/charset"
+  ]
   revision = "f5079bd7f6f74e23c4d65efa0f4ce14cbd6a3c0f"
 
 [[projects]]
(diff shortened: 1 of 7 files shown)


kytrinyx:
@FabienTregan would you have a chance to pull this down and see if any tests are failing? Or I guess... any _different_ tests than the ones that are failing on the `nextercism` branch?

kytrinyx:
@nywilken I'm going to merge this, but would *love* a review and suggestions for improvement when you have some time. There is no hurry, as I think that for the most part this is a significant improvement.

nywilken:
@kytrinyx I haven’t forgotten about this PR. I will review and give feedback where neccessary.

FabienTregan:
I've seen that in commit e9299ff you inlined configuration tests. In commit 818bc53b we tryed to break the out the tests to get better test-tools output and also allow for running next tests cases even if an earlier test failed. Is there something I missed that made you prefer to rool that back (e.g. some test that can't be run if previous test failed)?
Maybe it should be rolled back on other tests changed in 818bc53b too.

kytrinyx:
@FabienTregan the tests were not wrong, but they became difficult to reason about, which made it much harder to make changes.

I have kept the improved test names, but removed unnecessary duplication (the test name doesn't have to say "should not do X" since that is already in the failure message).

> allow for running next tests cases even if an earlier test failed

That is already the case. An assertion doesn't cause the test run to fail. The core test library distinguishes between `t.Error()` and `t.Fatal()`, so that we _can_ cause an entire test to fail if we want to.

I don't want to roll back any changes, but I will be making other updates to make tests more straight-forward to reason about, which as a first pass means more duplication, fewer loops, and especially fewer conditionals nested within loops. (I spent 3 hours this morning hunting down a problem that I wasn't seeing because test setup was checking what a particular argument was and doing different setup for a certain test inside the loop for all the tests, inside a conditional).

The important things for me with tests are that it is unambiguous what the setup is (which flags, which arguments, which files have we made up front, do we have a test server stubbing a response? If so, what is it doing).

I'm OK with some optimizations, but I want to make them only when I have a good grasp of what things a particular set of tests are communicating. So here for example: https://github.com/exercism/cli/pull/611/commits/9d609ea064f7babe9af8848d4d42f82158a206bd
the fake test server always does the same thing. Many tests don't need it, but at this point I was about to need it for a second endpoint. Since there wasn't going to be anything different in the server for different tests, I'm happy to hide it away in a method so I can say `ts := fakeSubmitServer`. I'm not super happy about passing in an argument that will get mutated, but it seems like a reasonable trade-off. I could have created it in the method and passed it back out, but it's not really all that much better. I don't know.

Anyway, I'm happy to discuss tweaks to the tests for readability and understandability, but my biggest concern with the tests is that they need to help us identify where and how to safely make a change, and a lot of the tests we've got aren't doing that, so I'm making some pretty significant changes as I work through the commands.

Issue #541, opened 2018-06-11
Set sensible defaults for workspace directory when configuring


kytrinyx:
When someone configures their `nextercism` CLI for the first time, `workspace_dir` will be blank in the config.

We need to set a sensible default based on their operating system.

- Mac: `~/Exercism`
- Linux: `~/exercism`
- Windows: `%HOMEDRIVE%%HOMEPATH%\Exercism`

@rpottsoh do you have any suggestions for what a good default Exercism directory would be for Windows?

rpottsoh:
I would think that c:\exercism would be ok.  The Installer also uses that as the, suggested, default location for the CLI as well.

Sent from my android device.

kytrinyx:
Perfect, thank you!

NobbZ:
I am not a windows user (as long as I have a choice), but `C:\exercism` is **not** OK.

Putting it anywhere outside of the user folder will cause trouble on corporate PCs or any system that houses more than one exercism student.

My choice where to stick to the same conventions as we do for Mac and Linux and put it under `%HOMEDRIVE%%HOMEPATH%\Exercism`.

rpottsoh:
@kytrinyx I can't disagree with @NobbZ's recommendation, he's not wrong.  My reservation to going this route is that some students may find it more difficult to navigate to that folder while at the command prompt.  

Likely when `cmd` or `powershell` are executed the student will land at `%homedrive%%homepath%`.  This is fine, they will just need to realize that the only need to `cd exercism`.

I am concerned for those students whose initial prompt is something other than `%homedrive%%homepath%`.  Changing to their solutions folder becomes more cumbersome since it will be several levels deep and the student may not know where to look or how to get there.

I don't have a anyway to gauge how savvy your average Windows user is with the command prompt.  I would tend to lean more towards them being less comfortable than most users of other operating systems where working at the command prompt is a more commonly practiced.

I suggested `c:\exercism` as I suspect in most cases this will be OK and is much easier for students to locate via the command line or via Windows Explorer.

rpottsoh:
Yes.  Realizing that the CLI can inform the student where it expects to find solutions.  :+1: :shipit:   I believe there is enough documentation to inform the student that they need to `cd` to the folder that contains their solution.  If there is a rash of Windows users that get stuck then we can deal with the nature of whatever the confusion might be at that time. 

In other words, at the moment, I can't think of how to make it any easier.

kytrinyx:
[...]

PR #605, opened 2018-07-09
Use sensible, os-specific workspace paths

kytrinyx:
This first adds the `Configuration` type, which is also added in https://github.com/exercism/cli/pull/604 (cherry-picked here). The `Configuration` type is a place to initialize all the environment-and-os-specific stuff, so that we can compute it only once and then pass around the object. It also lets us easily stub out env-related things for testing.

I also added stuff to that type in subsequent commits.

The sensible defaults that we've chosen are:

- Mac: `~/Exercism`
- Linux: `~/exercism`
- Windows: `%HOMEDRIVE%%HOMEPATH%\Exercism`

I'm using conditional compilation for the tests, since if we do `filepath.Join` on a *nix machine it will give `/` as the `os.PathSeparator`, and the same code will give `\` on a windows machine, so even stubbing out env things will give us nonsensical output. It doesn't make sense to make the test expectation check runtime environment, to determine which direction to make the slash (e.g. expect the windows path to be `"C:\\MyDocs/Exercism"` on *nix and `"C:\\MyDocs\\Exercism"` on windows). That just echoes the same underlying logic as the code, and doesn't really test anything.

Closes https://github.com/exercism/cli/issues/541


kytrinyx:
@FabienTregan Would you be able to pull this down and run the `config/configuration_windows_test.go` to see if it makes sense?

config/configuration.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func userHome() string {
+	var dir string
+	if runtime.GOOS == "windows" {
+		dir = os.Getenv("USERPROFILE")
+		if dir != "" {
+			return dir
+		}
+		dir = os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
```
nywilken: `filepath.Join`

config/configuration.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+	// If all else fails, use the current directory.
+	dir, _ = os.Getwd()
+	return dir
+}
+
+func DefaultWorkspaceDir(cfg Configuration) string {
+	dir := cfg.DefaultDirName
+	if cfg.OS != "linux" {
```
nywilken: On Mac it probably won’t matter as the file system is case insensitive by default. I think Window is the same. At least last time I checked.

config/configuration_notwin_test.go
```diff
@@ -0,0 +1,30 @@
+// +build !windows
```
nywilken: 👍🏽convention is to call this file `configuration_test.go` and on Windows call it `configuration_windows_test.go`

config/configuration.go
kytrinyx: The idiom on linux is to use lowercase directory names.

If you look at all the directories in the home directory on a mac by default, they're all capitalized (e.g. `Downloads`). Same with Windows.

So the defaults are not because the filesystems are case sensitive, but to match the pattern that they use for their other default directories.

kytrinyx:
> On that note we should probably add Windows to Travis if supported for Go. I have not tested YET!!!!

Agreed. I've added AppVeyor on a separate branch.

FabienTregan:
> On that note we should probably add Windows to Travis if supported for Go. I have not tested YET!!!!

Having a quick look, but https://docs.travis-ci.com/user/multi-os/ made me think it supports linux and osx only.

kytrinyx:
> it supports linux and osx only.

@FabienTregan I think you're right.

kytrinyx:
I've rebased this onto nextercism, which added the appveyor config... and it passes!

kytrinyx:
I'm going to merge this so we can rebase the configure-cmd and get appveyor added there as well.

PR #606, opened 2018-07-09
Ensure that we have a green build for Windows

kytrinyx:
This doesn't *fix* anything for windows, it simply splits out tests that are failing for windows and ensures that we don't build them.

**Configure Command**
The tests for the configure command are all being rewritten
in a separate branch. We will ensure that the new tests are passing.
The current ones are not worth fixing since they're going away.

**Default Workspace**
The default workspace logic is being rewritten in a separate branch.
We will make sure we have windows-specific tests there instead of fixing these only
to rewrite them later.

**Resolve Function**
The Resolve function will need to be fixed to be meaningful on
Windows.

**Locate**
We need to simplify the workspace logic.
As part of that work (later this week), we will ensure that it works for windows as well.


kytrinyx:
@FabienTregan I've gotten a green build! I cheated massively, but this should get us to where we can start fixing things sensibly.

kytrinyx:
FYI @nywilken

Issue #608, opened 2018-07-10
For launch, simplify submit command to only accept a list of files


kytrinyx:
We've gone around and around on how this will optimally work, and it keeps not being quite satisfactory.

After considering some of the friction reported from mentors, as well as reading through the discussion about allow-lists and block-lists in #458, and trying to figure out how we would document this, I'd like to simplify the submit command for launch to only accept a list of arguments that represent paths to files to be submitted.

PR #611, opened 2018-07-11
Rework submit command

kytrinyx:
The conclusion from the v2 beta is that the submit command is too hard to use.

In order to bring this back to something manageable I'm dramatically reducing the number of ways that you can interact with the submit command to exactly one: submitting a list of files as arguments:

    exercism submit path/to/file1 [path/to/file2...]

Once we get this worked out, and have added some features and tooling that will help us filter files more sensibly, we can judiciously add back a way to submit an entire directory (see https://github.com/exercism/meta/issues/107).

I've added extensive tests for the submit command, and I've made them very straight forward, which for now means quite a lot of duplication. I'd rather have duplication than conditionals within loops to do setup that isn't clear.

Closes https://github.com/exercism/cli/issues/608

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -123,15 +123,23 @@ Download other people's solutions by providing the UUID.
 			IsRequester: payload.Solution.User.IsRequester,
 		}
 
+		dir := filepath.Join(usrCfg.Workspace, solution.Track)
+		os.MkdirAll(dir, os.FileMode(0755))
+
 		var ws workspace.Workspace
 		if solution.IsRequester {
-			ws = workspace.New(filepath.Join(usrCfg.Workspace, solution.Track))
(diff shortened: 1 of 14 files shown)


kytrinyx:
@nywilken I'm going to merge this and keep working to allow us to launch the new site, but after this week is over (and when your life has settled a little bit) I'd love to discuss design decisions and direction to make sure I don't go too far down a path that is not useful to us.

nywilken:
@kytrinyx absolutely. I’ve been trying to looks at things as they get created but it has been hard for me to wrap my head around some of the changes. I need to sit down with a machine and really step through things so I can test and fully discuss. I think your approach of simplifying to base functionality is good for the launch.

PR #614, opened 2018-07-12
Do not clobber when configuring workspace

kytrinyx:
We already make sure we're not clobbering anything when setting the default workspace,
however if you pass the --workspace flag explicitly, we have just been trusting the
input.

There is one case where we don't want to trust the input, and that is if they
downloaded the CLI to the exact same place as the workspace that they
pass as a flag.

This will typically happen if:

1. They downloaded the CLI to their home directory
2. Extracted the archive right there
3. Followed the default instructions for configuring
4. At this point, the error message will
   - explain what is wrong, and
   - show how to use --workspace to explicitly configure it
5. If they then use --workspace with the default it blows up

Instead of trying to overwrite the CLI binary with a directory,
this adds an extra check to say that if there's a directory at the
location, it's fine, but if it's anything else, exit with an error
message.

### DIFF
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -150,6 +150,25 @@ func runConfigure(configuration config.Configuration, flags *pflag.FlagSet) erro
 		workspace = cfg.GetString("workspace")
 	}
 	workspace = config.Resolve(workspace, configuration.Home)
+
+	if workspace != "" {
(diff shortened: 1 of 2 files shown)


kytrinyx:
I'm going to merge this in, but as always I would love to hear it if you have thoughts on anything.

Issue #394, opened 2017-05-30
Homebrew bash/zsh completion


mulderje:
I noticed that the bash/zsh completion files aren't included in the install tar files, and am unsure what the generation process looks like. If not included in the tar files, have you considered leveraging the [Homebrew resource calls](https://github.com/Homebrew/brew/blob/master/docs/Formula-Cookbook.md#specifying-gems-python-modules-go-projects-etc-as-dependencies) the the cli website and installing that way?

Also happy to put this over in the Homebrew repo if more appropriate, but figured I would start here since the files didn't seem to be in the tars.

https://github.com/Homebrew/homebrew-core/blob/c53141dcda9d1a062eba89f77a5c02f09f6b5bd2/Formula/exercism.rb

Thanks,
jon

tengwar:
There's the same problem with Linux tarballs - they contain the executable and nothing else.

kytrinyx:
This is a great point. We should definitely consider adding these to the build script.

I need to figure out how to tell Homebrew where to put the auto-complete scripts, or if it has some good defaults.

kytrinyx:
I'm closing this as a duplicate of https://github.com/exercism/cli/issues/508

PR #616, opened 2018-07-13
Add shell completions and README to build script

### LINKED ISSUE #508: Distribute completion scripts along with executable
The completion scripts currently live in the CLI website repository, which—as @QuLogic mentions [here](https://github.com/exercism/cli-www/issues/34)—doesn't make much sense.

We should update the `bin/build-all` build script to produce a directory containing:

- the executable
- the completion scripts
- a README file that explains what to do with the executable (make sure it's in your path) and the completion scripts (pick the right one, source it in your shell config)

Then tar/zip as before.

jdsutherland:
Fixes #508. Possibly requires changes to facilitate #394? Additionally, might want to make relevant deletions from https://github.com/exercism/cli.exercism.io per https://github.com/exercism/DEPRECATED.cli-www/issues/34?

Completion scripts currently live in the CLI website repository which is
deemed inappropriate

Update build script to add shell completions and README to tar/zips


jdsutherland:
From #394:

> I need to figure out how to tell Homebrew where to put the auto-complete scripts, or if it has some good defaults.

I have no prior experience with the homebrew repo but it seems https://github.com/Homebrew/homebrew-core/blob/c53141dcda9d1a062eba89f77a5c02f09f6b5bd2/Formula/fzf.rb#L52 might be a good ref. 
`tree ~/code/fzf/shell`
shell
├── completion.bash
├── completion.zsh
I'm not sure if there's more to it than that.

bin/build-all
```diff
@@ -13,6 +13,15 @@ ARMVAR=github.com/exercism/cli/cmd.BuildARM
 # handle alternate binary name for pre-releases
 BINNAME=${NAME:-exercism}
 
+get_shell_completions() {
+        SHELLCOMP_BASEURL=http://cli.exercism.io/shell/exercism_completion
```
kytrinyx: Let's keep the scripts in this repository instead of in the static HTML repo (as you suggested). That way this will be a simple file inclusion rather than a network call.

kytrinyx:
Here's the error I got when I tried to extract the archive:

```
$ tar -xvzf exercism-mac-64bit.tgz 
x exercism
x ../shell/: Path contains '..'
x ../shell/exercism_completion.bash: Path contains '..'
x ../shell/exercism_completion.zsh: Path contains '..'
x ../BUILD.md: Path contains '..'
tar: Error exit delayed from previous errors.

```

kytrinyx:
FYI: I changed the base of this PR to point to master (I've been changing branches around, so what is on master now is what was previously on nextercism).

kytrinyx:
❤️ 💛 💚 💙 💜 🖤 thanks for this!

Issue #613, opened 2018-07-12
Debug windows-specific problem in transmission test


kytrinyx:
There is a test in `workspace/transmission_windows_test.go.go` that is failing on Windows.
I've added a `t.Skip()` in order to keep the build green.

I don't know whether this is a problem in the test-setup, or whether this represents an actual problem for Windows-users using the CLI.

nywilken:
Do you have any references to the error or a URL to the build that was failing on on Windows? I am testing on Windows and the test are passing at the moment. I am going to step through the test cases now to see if there is anything obvious

kytrinyx:
@nywilken I think we can actually get away with getting rid of transmission entirely. I just did an experiment where I inlined the bit that we need, and it's looking much better. I'l push it up now.

The builds are somewhere in https://ci.appveyor.com/project/kytrinyx/cli/history (I should have saved a link to the exact one!).

nywilken:
> The builds are somewhere in https://ci.appveyor.com/project/kytrinyx/cli/history (I should have saved a link to the exact one!).

No worries on that, you left a link on the submit failure which was super helpful. 

If we can drop the transmission entirely that would be great because it does add a level of complexity on top of file submissions and path look ups. I'll take a look at your changes. I opened a quick PR for #612 to see if Appveyor goes green. I mention that I am not 100% confident in file submission because of my lack of understanding around it, but made after the transmission drop things will become clear. 

I'll circle back later today to address comments, thoughts, and open issues.

PR #617, opened 2018-07-13
Get rid of transmission type

kytrinyx:
This inlines the little bit of logic that we needed from the workspace.Transition
and deletes the type.

We may find that we want to re-introduce an abstraction for some of this,
but I want to write it all down as simply as possible first and then go from there.

Closes https://github.com/exercism/cli/issues/613


kytrinyx:
Merging so I can move forward with the launch, but as always I'd love some eyes on this and am happy to put in some time improving the design based on feedback in new PRs.

PR #618, opened 2018-07-13
Fix relative path test on Windows

nywilken:
This changes fixes the one test for Windows, but I am not entirely
confident that the test passing means that it will work 100% of the time
on Windows. I am still wrapping my head around how the submission stuff
should play out.

### DIFF
--- a/cmd/submit_relative_path_windows_test.go
+++ b/cmd/submit_relative_path_windows_test.go
@@ -13,7 +13,7 @@ import (
 )
 
 func TestSubmitRelativePath(t *testing.T) {
-	t.Skip("The Windows build is failing and needs to be debugged.\nSee https://ci.appveyor.com/project/kytrinyx/cli/build/110")
+	//t.Skip("The Windows build is failing and needs to be debugged.\nSee https://ci.appveyor.com/project/kytrinyx/cli/build/110")
 
 	oldOut := Out
 	oldErr := Err
@@ -30,6 +30,7 @@ func TestSubmitRelativePath(t *testing.T) {
 
 	tmpDir, err := ioutil.TempDir("", "relative-path")
 	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
 
 	dir := filepath.Join(tmpDir, "bogus-track", "bogus-exercise")
 	os.MkdirAll(dir, os.FileMode(0755))
@@ -55,5 +56,5 @@ func TestSubmitRelativePath(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Equal(t, 1, len(submittedFiles))
-	assert.Equal(t, "This is a file.", submittedFiles["/file.txt"])
+	assert.Equal(t, "This is a file.", submittedFiles["\\file.txt"])
 }


kytrinyx:
> I am still wrapping my head around how the submission stuff should play out.

You and me both!

cmd/submit_relative_path_windows_test.go
```diff
@@ -55,5 +56,5 @@ func TestSubmitRelativePath(t *testing.T) {
 	assert.NoError(t, err)
 
 	assert.Equal(t, 1, len(submittedFiles))
-	assert.Equal(t, "This is a file.", submittedFiles["/file.txt"])
+	assert.Equal(t, "This is a file.", submittedFiles["\\file.txt"])
```
kytrinyx: OMG. Of course.

This should be `string(os.PathSeparator) + "file.txt"` or something like that. I wonder if that will automatically escape the string.

cmd/submit_relative_path_windows_test.go
```diff
@@ -30,6 +30,7 @@ func TestSubmitRelativePath(t *testing.T) {
 
 	tmpDir, err := ioutil.TempDir("", "relative-path")
 	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
```
kytrinyx: Nice. I need to add this logic everywhere.

cmd/submit_relative_path_windows_test.go
nywilken: I don't think it will escape, but escaping is not needed as the key actually is `\file.txt`. Changing to `string(os.PathSeparator) + "file.txt"` works as expected. However, I have a question why does the filename need to have the path separator? Should it not just be the filename?

kytrinyx: It's what the exercism API sends back, relative to the workspace directory, including the path separator. We could have done it either way, I don't think we had any reason to do one over the other.

kytrinyx: s/sends back/expects/ (sends back on the download)

kytrinyx: Wait. uhm.... is it just setup in the test that does this? In that case I think you're right, we should just remove it.

kytrinyx:
I made a tweak to the branches on this repository.

**TL;DR: You need to edit this PR to be against `master` instead of `nextercism`**

- I renamed the old master to `v2.x` (I might just make this a tag, but for now it's a branch)
- I pushed a copy of `nextercism` up to `master`

Once we've made sure everything looks right we can delete the nextercism branch.

cmd/submit_relative_path_windows_test.go
kytrinyx: I think the most reasonable thing right now is to keep the leading slash, since there are so many moving parts. It could be worth reconsidering down the road, as whether we change this now or later, we still have to be able to take into account the exercises that _do_ have the leading slash.

nywilken: I agree with you on this one; especially since I have not been able to give a lot of time. I would hate to start a larger thing right now and not be able to follow it all the way through. 

If you are okay with it I will merge this in.

kytrinyx: Yepp, I think that's fine. I'll go ahead and merge.

PR #622, opened 2018-07-13
Fix multi-file submission bug

kytrinyx:
When I rewrote the tests to go through the runSubmit function so I can
inject stuff, I forgot to actually consider what the command setup was
configuring.

In this case we had left a strict requirement that the submit command
only accept one single argument.

I've deleted the restriction, and tested manually.

I'll add a proper test when things slow down a bit.


nywilken:
Tested on Linux. Will confirm on Windows as well.

PR #630, opened 2018-07-14
Move .solution.json to hidden subdirectory

### LINKED ISSUE #557: Access denied to .solution.json on Windows
# Environement

* Windows command line (cmd.exe)
* Latest exercism build (from nextercism branch)
* Empty workspace

# Observed behaviour

If you try to download an exercism twice, the second time it will fail because with an access denied. Submiting also gives the same error 👍 
```
D:\code\Exercism>v2 download hello-world --track=rust

D:\code\Exercism>v2 download hello-world --track=rust
Error: open D:\code\Exercism\workspace\rust\hello-world\.solution.json: Access is denied.
```
The problem probably occures anytime the .solution file is modified.

#Expected behaviour

Just working :)

#Analysis

* The problem comes from the fact that go's WriteFile fails on hidden file under windows (see https://github.com/FabienTregan/TestWritingToHiddenFilesInGo ).
* I came accross the problem while trying to make the tests pass on windows, and worked around the problem by deleting old configuration file in the test : https://github.com/FabienTregan/cli/blob/19c10054e0621104f9f53c617946c72329b4864b/cmd/prepare_test.go#L47

# Possible solutions

Since I could not find a way to write to hidden files on windows with go, two possible solutions come to me:
* avoid hiding the file
* delete the file before writting it

### LINKED ISSUE #607: Put .solution.json in a hidden directory in the exercise
Rather than putting the `.solution.json` into the exercise directory directly, we should put it in a hidden subdirectory (per @ErikSchierboom's suggestion in https://github.com/exercism/cli/issues/458#issuecomment-399670848)

Let's use `.exercism`, even though for Windows we'll have to do extra work to hide it.

This may solve https://github.com/exercism/cli/issues/557 as we would no longer be trying to write to a hidden file, but rather a file in a hidden directory.

jdsutherland:
Closes #607
Closes #557


jdsutherland:
It seems the AppVeyor error is the same as #557. Is the intended solution to not hide .solution.json?

kytrinyx:
> s the intended solution to rename .solution.json, (i.e., .exercism/solution.json)?

Yeah we're aiming for `.exercism/solution.json`

cmd/download_test.go
```diff
@@ -84,7 +84,7 @@ func TestDownload(t *testing.T) {
 		},
 		{
 			desc:     "It creates the .solution.json file.",
-			path:     filepath.Join(cmdTest.TmpDir, "bogus-track", "bogus-exercise", ".solution.json"),
+			path:     filepath.Join(cmdTest.TmpDir, "bogus-track", "bogus-exercise", "./.exercism/.solution.json"),
```
kytrinyx: We can't hard-code slashes here, as on Windows the slashes go the other way.

```
filepath.Join(cmdTest.TmpDir, "bogus-track", "bogus-exercise", ".exercism", "solution.json")
```

nywilken:
> s the intended solution to rename .solution.json, (i.e., .exercism/solution.json)?

To be clear, the path on *nix will be `EXERCISEROOT/.exercism/.solution.json` and on Windows it should be `EXERCISEROOT\.exercism\.solution.json`

cmd/download.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const hiddenSolutionDir = ".exercism"
```
nywilken: This is not being used anywhere. Is it needed?

workspace/solution.go
```diff
@@ -12,8 +12,11 @@ import (
 	"github.com/exercism/cli/visibility"
 )
 
+const ignoreSubdir = ".exercism"
 const solutionFilename = ".solution.json"
 
+var solutionRelPath = filepath.Join(ignoreSubdir, solutionFilename)
```
nywilken: Seems like this is only used within NewSolution does it need to be a package level variable?

cmd/download.go
jdsutherland: Oops, not sure how that slipped through.

workspace/solution.go
nywilken: Im speaking about solutionRelPath. I did a quick search and didn’t see other references. If that variable is used three times and we don’t plan on changing it for testing should it be const?

nywilken:
Something to think about. When we roll this change out it will break already downloaded exercises as the solution json file will be in a new path. Forcing them to redownload to fix. 

If that is the case then our code should detect if an existing solution file exists and move it for the user. Thoughts?

workspace/solution.go
jdsutherland: It can't be const: `[gometalinter] const initializer filepath.Join(ignoreSubdir, solutionFilename) is not a constant (vet) [Error]`

nywilken: Sorry I wasn’t clear, if we make solutionFilename a const with the new path we wouldn’t need the other variables as they are just being used to construct the final solution path 
`const SolutionFilename = filepath.Join(“.exercism”,”solution.json”)`

I also think the it should be exported to denote that it is actually shared or is expected to be.

workspace/solution.go
```diff
@@ -67,11 +70,13 @@ func (s *Solution) Write(dir string) error {
 		return err
 	}
 
-	path := filepath.Join(dir, solutionFilename)
+	path := filepath.Join(dir, ignoreSubdir)
```
nywilken: Just a heads up if we go with one const for filename path then we have to use `filepath.Dir(SolutionFilename)` to just get the directory path and create that if it doesn’t already exist

kytrinyx:
> When we roll this change out it will break already downloaded exercises as the solution json file will be in a new path. Forcing them to redownload to fix.

Yes, I was hoping to get this sorted before the launch, but it's been an eventful few days and I didn't get to it.

We need to handle detecting the old location and moving the file for the user if it is found.

workspace/solution.go
jdsutherland: Gotcha. I don't think `filepath.Join` can be const: `[gometalinter] const initializer filepath.Join(“.exercism”, ”solution.json”) is not a constant (vet) [Error]`

jdsutherland: Gotcha. It appears that `filepath.Join` can be const: `[gometalinter] const initializer filepath.Join(".exercism", "solution.json") is not a constant (vet) [Error]`

nywilken: Well then it seems like the best advise I can give is to carry on as you were. Disregarding my coments on this file. 

 I didnt think about the the separator not being known until run time making it unfit to use as a const, my fault. 

Thanks for vetting and for your time.

jdsutherland: No problem. I appreciate your feedback.

jdsutherland:
> We need to handle detecting the old location and moving the file for the user if it is found.

I gave this a shot. I have questions about writing tests for this but it depends on if these changes are serviceable so I wanted to get feedback. Given the `os.Rename` call, tests can't simply add a fixture like most of the existing so I'm wondering if you'd recommend mocking it or using `ioutil.TempFile`?

jdsutherland:
> We need to handle detecting the old location and moving the file for the user if it is found.

I gave this a shot and think it's in a good state for review.

> Interesting enough removing the call to visibility.Hidefile fixes all the tests.
>> How strongly do we feel that the metadata directory needs to be hidden?

Do we want to remove `visibility.Hidefile`?

workspace/solution.go
```diff
@@ -87,3 +95,22 @@ func (s *Solution) PathToParent() string {
...
+	}
+	return nil
+}
+
+func migrateLegacySolutionFile(path string, legacySolutionPath string, solutionPath string) error {
+	if err := createIgnoreSubdir(filepath.Join(path, IgnoreSubdir)); err != nil {
+		return err
+	}
+	if err := os.Rename(legacySolutionPath, solutionPath); err != nil {
+		return errors.New("failed migrating legacy solution file")
```
jdsutherland: This doesn't actually end up getting handled (it gets called in the path traversal loop in `workspace.SolutionDir`). Would an acceptable solution be to `panic`?

nywilken:
@jdsutherland I’ve been a little caught up. Thanks for updating.  I’ll take a look at your latest changes and get back to you by tomorrow.

workspace/solution.go
```diff
@@ -67,11 +71,15 @@ func (s *Solution) Write(dir string) error {
...
-	path := filepath.Join(dir, solutionFilename)
+	path, err := createIgnoreSubdir(dir)
+	if err != nil {
+		return err
+	}
 
 	// Hack because ioutil.WriteFile fails on hidden files
 	visibility.ShowFile(path)
 
+	path = filepath.Join(dir, solutionRelPath)
```
nywilken: I think this should now be `filepath.Join(path, solutionRealPath)`

workspace/solution.go
```diff
@@ -67,11 +71,15 @@ func (s *Solution) Write(dir string) error {
...
 	}
 
-	path := filepath.Join(dir, solutionFilename)
+	path, err := createIgnoreSubdir(dir)
+	if err != nil {
+		return err
+	}
 
 	// Hack because ioutil.WriteFile fails on hidden files
 	visibility.ShowFile(path)
```
nywilken: So we are calling ShowFile on a directory then down below we are returning the result of visibility.HideFile on the full path to the solution file. 

At this point I wonder if it is needed since it is now the directory that is a dot file.

workspace/solution.go
```diff
@@ -87,3 +95,23 @@ func (s *Solution) PathToParent() string {
...
+	}
+	return path, nil
+}
+
+func migrateLegacySolutionFile(legacySolutionPath string, solutionPath string) error {
+	if _, err := createIgnoreSubdir(filepath.Dir(legacySolutionPath)); err != nil {
+		return err
+	}
+	if err := os.Rename(legacySolutionPath, solutionPath); err != nil {
+		return errors.New("failed migrating legacy solution file")
```
jdsutherland: This doesn't actually end up getting handled (it gets called in the path traversal loop in `workspace.SolutionDir`). If `os.Rename` fails, path traversal will continue until a [missing metadata error](https://github.com/exercism/cli/commit/53bb62de80806f504109901e95e931ea956884df) is reached.

Would an acceptable solution be to `panic`?  It seems that `os.Rename` really shouldn't fail and would be truly exceptional if it did. On the other hand, the "missing metadata error" is more useful to the user and provides an actionable solution (download a new exercise and copy files over).

config/track.go
```diff
@@ -7,7 +7,7 @@ import (
 
 var defaultIgnorePatterns = []string{
 	".*[.]md",
-	"[.]solution[.]json",
+	"solution[.]json",
```
nywilken: Is there any drawbacks to using `[.]?solution[.]json` to ensure both old and new solution files are ignored?

workspace/solution.go
```diff
@@ -87,3 +95,23 @@ func (s *Solution) PathToParent() string {
...
+		}
+	}
+	return path, nil
+}
+
+func migrateLegacySolutionFile(legacySolutionPath string, solutionPath string) error {
+	if _, err := createIgnoreSubdir(filepath.Dir(legacySolutionPath)); err != nil {
+		return err
+	}
+	if err := os.Rename(legacySolutionPath, solutionPath); err != nil {
```
nywilken: What happens if there is no solution file old or legacy? 

Should we be checking if the file exists before trying to migrate it? With that said, I think we should print to Stderr that the solution is being migrated so that the user is made aware that a file is being changed for them. 

This could prove helpful for those users who have their files under version control.

workspace/solution.go
nywilken: IMO the missing metadata message informs the user on how to resolve. If we panic with this error message the user has no where to go. 

Error messages with context and possible remediation steps is the way to go.

jdsutherland:
> I am curious to see how the code behaves if you drop the calls to the visibility package now that we are using .exercism directory.

Do you mean remove *all* refs to `visibility` - isn't `visibility.HideFile(path)` needed to hide the directory in Windows?

workspace/solution.go
jdsutherland: > What happens if there is no solution file old or legacy?

Then errMissingMetadata gets handled (`migrateLegacySolutionFile` never gets called).

> Should we be checking if the file exists before trying to migrate it?

In this implementation, `migrateLegacySolutionFile` only gets called when a legacy solution (`$EXERCISE_ROOT/.solution.json`) is found (this is checked before checking for `$EXERCISE_ROOT/.exercism/solution.json`). It is potentially dangerous and poor defensive programming practice not to check this inside the actual function though. Good call.

In the case that both legacy and modern solution files exist (due to user manually moving files around) the legacy would overwrite the modern solution. It's not apparent to me that this would be undesirable but I don't have a ton of experience with the product and it's quite possible I'm ignorant here.

In the case where both legacy and modern solutions exist, what should be done? Which solution file is more likely to be correct (I'd bet the legacy because the only sensible way I imagine this case occuring is if a user downloads a new exercise and copies the contents of a legacy solution over)? I see the options as:

1. Don't migrate, print Stderr, no error => user should be able to successfully submit (but has a junk legacy solution file left in the workspace)
2. Same as 1. but delete the legacy metadata file for the user
2. Same as 1. but replace the modern metadata file with the legacy
4. Don't migrate, print Stderr, return error => user will get the missing metadata error (this seems confusing and bad)

What to print to Stderr? Perhaps: "Attempted to migrated solution metadata to %s but file already exists."


> With that said, I think we should print to Stderr that the solution is being migrated so that the user is made aware that a file is being changed for them.

Good call, I'll add it.

nywilken:
@jdsutherland I haven’t forgotten about this PR. Thanks for keeping it updated. I will tackle this PR with the team PR later today.

workspace/solution.go
```diff
@@ -87,3 +83,37 @@ func (s *Solution) PathToParent() string {
...
+		return err
+	}
+	if _, err := os.Lstat(solutionPath); err != nil {
+		if err := os.Rename(legacySolutionPath, solutionPath); err != nil {
+			return err
+		}
+		fmt.Fprintf(os.Stderr, "\nMigrated solution metadata to %s\n", solutionPath)
+	} else {
+		// TODO: decide how to handle case where both legacy and modern metadata files exist
+		fmt.Fprintf(os.Stderr, "\nAttempted to migrate solution metadata to %s but file already exists\n", solutionPath)
```
nywilken: I think the new metadata takes precendence which leads into my next comment about workspace,checkSolutionFile. It currently checks for a legacy metdata first, but I think we should check for the new path first and only check for legacy if it doesn’t exist.

workspace/workspace.go
```diff
@@ -198,9 +198,21 @@ func (ws Workspace) SolutionDir(s string) (string, error) {
...
 		path = filepath.Dir(path)
 	}
 }
+
+func checkSolutionFile(path string) error {
+	legacySolutionPath := filepath.Join(path, ".solution.json")
+	solutionPath := filepath.Join(path, SolutionMetadataFilepath())
+
+	if _, err := os.Lstat(legacySolutionPath); err == nil {
+		return migrateLegacySolutionFile(legacySolutionPath, solutionPath)
```
nywilken: If we don’t rename and instead warn the use that a deprecated solution files exists would it make the migration easier. @kytrinyx @nobbz thoughts on leaving the file or renaming it for the user. 

I recommended renaming but I’m not sure how users will feel about the CLI renaming

jdsutherland:
>Do you mean remove all refs of visibility package? Isn't visibility.HideFile(path) needed to hide the directory in Windows?

>>That was the question I was asking and it seems like it may not be needed now that tests are passing on Windows. We need to thoroughly test with an actual build and file downloads.

If testing on Windows is successful, should the `visibility` package be deleted as a result of these changes? It wouldn't have any remaining referrers.

cmd/submit_test.go
```diff
@@ -260,6 +261,50 @@ func TestSubmitOnlyEmptyFile(t *testing.T) {
...
+	v := viper.New()
+	v.Set("token", "abc123")
+	v.Set("workspace", tmpDir)
+	v.Set("apibaseurl", ts.URL)
+	cfg := config.Configuration{
+		Persister:       config.InMemoryPersister{},
+		Dir:             tmpDir,
+		UserViperConfig: v,
+	}
+	expectedSolutionPathAfterMigration := filepath.Join(dir, workspace.SolutionMetadataFilepath())
```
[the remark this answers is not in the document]
jdsutherland: Yeah, I think it would help clarify intent and ensure completeness.

cmd/submit_test.go
```diff
@@ -260,6 +261,50 @@ func TestSubmitOnlyEmptyFile(t *testing.T) {
 	assert.Regexp(t, "No files found", err.Error())
 }
 
+func TestSubmitExerciseWithLegacySolutionMetadataFileAndGetsMigrated(t *testing.T) {
```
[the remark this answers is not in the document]
jdsutherland: `TestLegacySolutionMetadataMigration` is more readable. The only potential concern about naming it `TestLegacySolutionMetadataMigration` is: the test isn't just testing migration but that a submission is still successful given a legacy solution file - is it worth indicating that in the name?

kytrinyx:
> If testing on Windows is successful, should the visibility package be deleted as a result of these changes? It wouldn't have any remaining referrers.

Yes, if we aren't referring to it, I think we should delete it.

workspace/solution.go
```diff
@@ -87,3 +83,37 @@ func (s *Solution) PathToParent() string {
 	}
 	return filepath.Join(dir, s.Track)
 }
+
+// SolutionMetadataFilepath is the path of the solution metadata file relative to the workspace.
+func SolutionMetadataFilepath() string {
```
kytrinyx: I don't understand how this works. Each exercise has its own metadata file, so this is relative to a subdirectory within the workspace.

kytrinyx:
I've started working on a change in https://github.com/exercism/cli/pull/682 that I think would help clarify things here.

With the `workspace.Exercise` type we could have two methods:

- `exercise.MetadataFilepath()`
- `exercise.LegacyMetadataFilepath()`

I think it would be worth making the change in multiple steps, for clarity.

1. Rename `workspace.Solution` to `workspace.ExerciseMetadata` or `workspace.SolutionMetadata` for clarity
1. Once the #682 gets merged, make a single PR that refactors to use `workspace.Exercise` and rely on `exercise.MetadataFilepath()` instead of hard-coding `.solution.json`
1. Finally, a PR (most of what is in this PR) that changes the default location of `MetadataFilepath()` and updates the behavior to fall back to looking in `LegacyMetadataFilepath()` if the default doesn't exist, as well as moving the legacy file to the new path.

Does that make sense?

workspace/solution.go
jdsutherland: I mistakenly used 'workspace' when I meant 'exercise' there.

kytrinyx: Ok, that makes sense. Thanks!

jdsutherland:
Yeah, that makes sense. I think what you've suggested will improve clarity. It seems that "exercise" and "solution" are used interchangeably in places but are defined differently at https://github.com/exercism/docs/blob/master/about/glossary.md. If I understand correctly, it seems that the metadata is static for an exercise - it's generated once when the user runs a download cmd and doesn't change when a solution gets submitted. If it doesn't change when a solution is submitted, calling it SolutionMetadata seems counterintuitive to me.

kytrinyx:
> If I understand correctly, the metadata is static - it's generated once when the user runs a download cmd and doesn't change when a solution gets submitted.

I think you're right that `ExerciseMetadata` makes more sense here. Thanks for helping think this through--getting terminology right is tricky (and so important!).

kytrinyx:
@jdsutherland How are you getting on?

I think we decided to not rename the solution and exercise metadata until after this work is done.

As far as I can tell, the next step would be to open a separate PR that ensures that we never hard-code `.solution.json` anywhere (that we always rely on the configured constant).

kytrinyx:
> I made a new PR

Awesome, I'll review that now!

PR #631, opened 2018-07-14
Don't accept token flag on download command

### LINKED ISSUE #544: Remove `--token` behavior from download command
The download command is overly complex.

Instead of accepting a token and passing it through to configure, we will print an error message and exit, per the description in https://github.com/exercism/cli/issues/522.

kytrinyx:
We need to know who you are in order to download the correct exercise with all the requisite
metadata.

Originally the download command allowed you to pass a token on the fly, during download.
That added a lot of complexity. This changes it so that we instead bail with a helpful error message
that explains how to configure the client.

The error message is the same as the one we are using for the submit command. I moved the error message to a constant. I'm not convinced this is the right choice, but it is one which will be cheap to back out of if we decide to.

My goal for today is to do the same sort of simplification/cleanup in download as I recently did in submit. I'm going to merge this to provide a basis for the next changes, but as with all of the changes in the last couple of weeks, feedback, thoughts, and ideas are very welcome.

Closes #544

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -36,22 +36,7 @@ Download other people's solutions by providing the UUID.
 		}
 		if usrCfg.Token == "" {
 			tokenURL := config.InferSiteURL(usrCfg.APIBaseURL) + "/my/settings"
-			msg := `
-
-    Welcome to Exercism!
-
-    To get started, you need to configure the the tool with your API token.
-    Find your token at
-
-        %s
-
-    Then run the configure command:
-
-
-        %s configure --token=YOUR_TOKEN
-
-    `
-			return fmt.Errorf(msg, tokenURL, BinaryName)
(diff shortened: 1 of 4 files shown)


cmd/download.go
```diff
@@ -30,16 +30,15 @@ latest solution.
...
 			return err
 		}
-		if token != "" {
-			RootCmd.SetArgs([]string{"configure", "--token", token})
-			if err := RootCmd.Execute(); err != nil {
-				return err
-			}
+		if usrCfg.Token == "" {
+			tokenURL := config.InferSiteURL(usrCfg.APIBaseURL) + "/my/settings"
+			return fmt.Errorf(msgWelcomePleaseConfigure, tokenURL, BinaryName)
```
nywilken: If a user calls this for the first time without having run configure ever tokenURL is empty. Now that we have made the migration should it just be https://exercism.io/my/settings? Or do we still want to try an infer the Token URL?

PR #635, opened 2018-07-14
Update minimum supported Go version to 1.9

nywilken:
This change drops support for Go1.8 and sets the minimum to Go1.9.
By making this change we are able to use the same testing steps on *nix and Windows. 

Travis and Appveyor build scripts have been updated to reflect the said testing step. I also included the `-cover` flag to `go test` so that we have a better idea which features need some testing :heart:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ language: go
 sudo: false
 
 go:
-  - 1.8
-  - 1.9
+  - "1.9"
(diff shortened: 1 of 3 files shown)


nywilken:
Looks like I got myself into a bit of history conflicts. Not feeling the merge commit so when I get behind my machine I’ll clean up the history (rebase/drop) and merge.

PR #642, opened 2018-07-14
Add more guards to the download command

kytrinyx:
This will give better error messages when downloading a solution without having configured the CLI for the new site.

Merging so I can cut a release.


kytrinyx:
Thanks @nywilken. I wasn't sure you were around and I need to get this release cut before I get on this next flight. This will slow down shortly, I promise!

PR #646, opened 2018-07-16
Fix issue with upgrading on Windows

### LINKED ISSUE #645: Windows 10: Resulting CLI after upgrade broken
Called `upgrade` from 2.4.1 CLI on Windows 10.  Upgrade was successful:
```
C:\Users\rpott>exercism
NAME:
   exercism - A command line tool to interact with http://exercism.io

USAGE:
   exercism [global options] command [command options] [arguments...]

VERSION:
   2.4.1

COMMANDS:
     configure     Writes config values to a JSON file.
     debug         Outputs useful debug information.
     download, dl  Downloads a solution given the ID of the latest iteration.
     fetch, f      Fetches the next unsubmitted problem in each track.
     list, li      Lists the available problems for a language track, given its ID.
     open, op      Opens exercism.io to your most recent iteration of a problem given the track ID and problem slug.
     restore, r    Downloads the most recent iteration for each of your solutions on exercism.io.
     skip          Skips a problem given a track ID and problem slug.
     status, st    Fetches information about your progress with a given language track.
     submit, s     Submits a new iteration to a problem on exercism.io.
     tracks, t     Lists the available language tracks.
     unsubmit, u   REMOVED
     upgrade       Upgrades the CLI to the latest released version.

GLOBAL OPTIONS:
   --config value, -c value  path to config file [%EXERCISM_CONFIG_FILE%, %XDG_CONFIG_HOME%]
   --verbose                 turn on verbose logging
   --help, -h                show help
   --version, -v             print the version


C:\Users\rpott>exercism upgrade
Downloading exercism-windows-64bit.zip
Successfully upgraded!

C:\Users\rpott>
```
Upon executing `exercism` after the upgrade I am presented with the following dialog:
image

My instinct is to go through, by hand, and remove all remnants of the old CLI and to try again.  I presumed that `upgrade` would be successful.

`exercism version` also yields the same error.

I'll keep my setup as is for the time being until I receive some feedback on what you might like me to try.

nywilken:
This changes ensures that the upgrade process does not try to replace
the exercism binary with a non executable file. There is probably a more
extensive check we can do to ensure we have the correct file. I did not
choose to use the full name as Configlet is also using the cli pkg for
upgrading its binary.

closes #645

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -164,6 +164,10 @@ func extractBinary(source *bytes.Reader, os string) (binary io.ReadCloser, err e
 		}
 
 		for _, f := range zr.File {
+			info := f.FileInfo()
+			if info.IsDir() || !strings.HasSuffix(f.Name, ".exe") {
+				continue
+			}
 			return f.Open()
 		}
 	} else {


cli/cli.go
```diff
@@ -164,6 +164,10 @@ func extractBinary(source *bytes.Reader, os string) (binary io.ReadCloser, err e
 		}
 
 		for _, f := range zr.File {
+			info := f.FileInfo()
+			if info.IsDir() || !strings.HasSuffix(f.Name, ".exe") {
```
NobbZ: Why iterating over all files?

Why not just extract `exercism.exe`?

nywilken: > Why iterating over all files?

I don't want to assume that the first file is the executable, even though that is probably the case.

> Why not just extract exercism.exe?

If I extract the whole zip I would have to create the structure locally, then write each of the files. I am essentially extracting the executable without writing the file to disk (less to clean up). I could check for the exact name, read it's contents, then return in case we decide to add more binaries to the release package.

nywilken:
@kytrinyx I realized that we are not closing the zip or the file after reading their contents. I want to add that in before merging. 

For future reference, am I okay to cut releases? During the mentorcism releases I think you were doing something custom. But now it should be a standard release, right?

cli/cli.go
nywilken: Anytime! Thanks for the question and review.

nywilken:
Thanks for cutting the release and info on mentorcism. If the situation araises again I cut any neccessary releases to help get fixes out and lighten your load 😉

PR #652, opened 2018-07-18
Add --team flag to download command

### LINKED ISSUE #546: Add `--team` flag to download command
We need to support teams on the download command.

The format is:

```
exercism download --exercise=hello-world --track=ruby --team=funfun
```

Call the same API endpoint as the rest of the download command, but with the addition of `&team_id`, passing the slug that was found.

Store the files under:

```
$WORKSPACE/teams/$team/$track/$exercise
```

kytrinyx:
We're working on implementing teams, which is pretty different from how it worked on the old site. In short, teams are a whole separate thing. If you are working on `bob` in Python on the main site, and you're a member of a team, then your `bob` does not automatically end up in the team pool of solutions. You need to explicitly submit it with a `--team` flag (not yet implemented in the CLI). Likewise, to work on an exercise for a team, you need to download it so that you get the correct metadata, then you can copy over your solution from your main track.

Closes #546

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -90,11 +90,19 @@ func runDownload(cfg config.Configuration, flags *pflag.FlagSet, args []string) 
 		return err
 	}
 
+	team, err := flags.GetString("team")
+	if err != nil {
+		return err
+	}
+
 	if uuid == "" {
 		q := req.URL.Query()
 		q.Add("exercise_id", exercise)
 		if track != "" {
 			q.Add("track_id", track)
+		}
+		if team != "" {
+			q.Add("team_id", team)
 		}
 		req.URL.RawQuery = q.Encode()
 	}
@@ -127,6 +135,7 @@ func runDownload(cfg config.Configuration, flags *pflag.FlagSet, args []string) 
 	solution := workspace.Solution{
 		AutoApprove: payload.Solution.Exercise.AutoApprove,
 		Track:       payload.Solution.Exercise.Track.ID,
+		Team:        payload.Solution.Team.Slug,
 		Exercise:    payload.Solution.Exercise.ID,
 		ID:          payload.Solution.ID,
 		URL:         payload.Solution.URL,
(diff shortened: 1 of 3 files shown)


cmd/download.go
```diff
@@ -136,6 +145,9 @@ func runDownload(cfg config.Configuration, flags *pflag.FlagSet, args []string)
 	}
 
 	dir := usrCfg.GetString("workspace")
+	if solution.Team != "" {
+		dir = filepath.Join(dir, "teams", solution.Team)
```
nywilken: This is a bit confusing. In payload it is a struct here Team is a string. I understand the difference but maybe it should be TeamSlug. Or maybe have a different name in the Solution struct

kytrinyx: I made this change.

kytrinyx:
Yeah, I submitted small parts of this as other PRs, so I need to rework it so it's just the smallest possible change.

kytrinyx:
I think this is ready for another round of review. I've cleaned it up and the PR is as small as I can get it and still be a coherent change :)

kytrinyx:
Wait. The first two commits are in a separate PR already. Hold off on reviewing this until the other PR is merged, then I'll rebase and this will be smaller.

workspace/solution.go
```diff
@@ -19,6 +19,7 @@ type Solution struct {
 	Track       string     `json:"track"`
 	Exercise    string     `json:"exercise"`
 	ID          string     `json:"id"`
+	Team        string     `json:"team,omitempty"`
```
nywilken: what is the intended empty state for Team “”, null? I ask because I am wondering if this should be of type *string. 

If the empty state is “” then your example of “null” in the test should be changed.

cmd/download_test.go
```diff
@@ -165,11 +170,16 @@ func fakeDownloadServer(requestor string) *httptest.Server {
 		fmt.Fprint(w, "")
 	})
 
-	payloadBody := fmt.Sprintf(payloadTemplate, requestor, server.URL+"/")
 	mux.HandleFunc("/solutions/latest", func(w http.ResponseWriter, r *http.Request) {
+		team := "null"
```
nywilken: See my comment about the Team field for the Solution type.

workspace/solution.go
kytrinyx: The empty state of the solution is `""`. The empty state of the payload we receive from the server is `null`.

kytrinyx:
Thank you!

PR #655, opened 2018-07-19
Replace accidental \ in path to /.

virtualsue:
Fixed a typo


virtualsue:
I should have noticed that originally. Done.

PR #659, opened 2018-07-20
Move assertion for solution metadata out of helper

kytrinyx:
The solution metadata varies on a test-by-test basis.
I don't like hiding that away in a helper.

### DIFF
--- a/cmd/download_test.go
+++ b/cmd/download_test.go
@@ -130,7 +130,15 @@ func TestDownload(t *testing.T) {
 		err = runDownload(cfg, flags, []string{})
 		assert.NoError(t, err)
 
-		assertDownloadedCorrectFiles(t, filepath.Join(tmpDir, tc.expectedDir), tc.requestor)
+		targetDir := filepath.Join(tmpDir, tc.expectedDir)
+		assertDownloadedCorrectFiles(t, targetDir, tc.requestor)
+
+		metadata := `{"track":"bogus-track","exercise":"bogus-exercise","id":"bogus-id","url":"","handle":"alice","is_requester":%s,"auto_approve":false}`
+		metadata = fmt.Sprintf(metadata, tc.requestor)
+		path := filepath.Join(targetDir, "bogus-track", "bogus-exercise", ".solution.json")
+		b, err := ioutil.ReadFile(path)
+		assert.NoError(t, err)
+		assert.Equal(t, metadata, string(b), "the solution metadata file")
 	}
 }
 
@@ -165,7 +173,6 @@ func fakeDownloadServer(requestor string) *httptest.Server {
 }
 
 func assertDownloadedCorrectFiles(t *testing.T, targetDir, requestor string) {
-	metadata := `{"track":"bogus-track","exercise":"bogus-exercise","id":"bogus-id","url":"","handle":"alice","is_requester":%s,"auto_approve":false}`
 	expectedFiles := []struct {
 		desc     string
 		path     string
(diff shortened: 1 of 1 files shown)


cmd/download_test.go
```diff
@@ -114,7 +114,15 @@ func TestDownload(t *testing.T) {
 		err = runDownload(cfg, flags, []string{})
 		assert.NoError(t, err)
 
-		assertDownloadedCorrectFiles(t, filepath.Join(tmpDir, tc.expectedDir), tc.requestor)
+		targetDir := filepath.Join(tmpDir, tc.expectedDir)
+		assertDownloadedCorrectFiles(t, targetDir, tc.requestor)
+
+		metadata := `{"track":"bogus-track","exercise":"bogus-exercise","id":"bogus-id","url":"","handle":"alice","is_requester":%s,"auto_approve":false}`
+		metadata = fmt.Sprintf(metadata, tc.requestor)
+		path := filepath.Join(targetDir, "bogus-track", "bogus-exercise", ".solution.json")
```
nywilken: For some reason I was under the impression that downloading the same exercise twice would result in different directories. Specifically `exercise` and `exercise-2`. But the CLI doesn't add a suffix, it just supports reading from a suffix.

nywilken: For the future maybe in #630 there should probably be an exported variable or function for getting the path to the solution metadata file.

cmd/download_test.go
```diff
@@ -114,7 +114,15 @@ func TestDownload(t *testing.T) {
...
+		b, err := ioutil.ReadFile(path)
+		assert.NoError(t, err)
+		assert.Equal(t, metadata, string(b), "the solution metadata file")
```
nywilken: Actually we should be cleaning up files here `os.RemoveAll(filepath.join(targetDir, "bogus-track")`

cmd/download_test.go
```diff
@@ -114,7 +114,15 @@ func TestDownload(t *testing.T) {
 		err = runDownload(cfg, flags, []string{})
 		assert.NoError(t, err)
 
-		assertDownloadedCorrectFiles(t, filepath.Join(tmpDir, tc.expectedDir), tc.requestor)
+		targetDir := filepath.Join(tmpDir, tc.expectedDir)
+		assertDownloadedCorrectFiles(t, targetDir, tc.requestor)
+
+		metadata := `{"track":"bogus-track","exercise":"bogus-exercise","id":"bogus-id","url":"","handle":"alice","is_requester":%s,"auto_approve":false}`
```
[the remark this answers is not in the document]
kytrinyx: The solution file that we write is minimized. I can take a look to see if I find a good way to strip out whitespace, or do a comparison that ignores whitespace.

cmd/download_test.go
kytrinyx: If they have different solution UUIDs, then they will download to separate directories. I think we have a bug in the separate directory thing, so I might strip that out until we need to support that feature.

kytrinyx: Yes, agreed!

cmd/download_test.go
kytrinyx: Good catch, thank you!

cmd/download_test.go
kytrinyx: I'm going to do that in a separate PR, as it touches other code than this download test.

cmd/download_test.go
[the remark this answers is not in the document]
kytrinyx: Found it. `json.Compact()`!

PR #668, opened 2018-07-23
Don't pass files to payload template in download test

kytrinyx:
The fake download server hard-codes three files.
We were passing these three files into the fake payload, which doesn't really make sense.

### DIFF
--- a/cmd/download_test.go
+++ b/cmd/download_test.go
@@ -159,22 +159,19 @@ func fakeDownloadServer(requestor string) *httptest.Server {
 	mux := http.NewServeMux()
 	server := httptest.NewServer(mux)
 
-	path1 := "file-1.txt"
-	mux.HandleFunc("/"+path1, func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


cmd/download_test.go
```diff
@@ -238,9 +235,9 @@ const payloadTemplate = `
 		},
 		"file_download_base_url": "%s",
 		"files": [
-		"%s",
-		"%s",
-		"%s"
+		"/file-1.txt",
```
nywilken: How do you feel about moving the constants, template, and fakeDownServer declarations to the top of the file? So reading the test file from top to bottom give a complete picture of what those variables are. 

Right now I have to scan towards the bottom of the file to see what requestorSelf is or even what we are doing with the template.

kytrinyx: I don't much like the idea, to be honest. I don't want to have to scan through a couple hundred lines of code before we get to the tests.

kytrinyx: Thanks! And thanks for taking the time to ask the question. I didn't know I didn't like it until you asked!

PR #675, opened 2018-07-24
Tweak output of troubleshoot command when the CLI is unconfigured

### LINKED ISSUE #671: Clarify configuration status in troubleshoot command
If you run `exercism troubleshoot` when nothing has been configured, we should show the default configuration directory as "configuration home", and explicitly say that it is unconfigured.

kytrinyx:
Right now the output of troubleshoot is a bit unhelpful when the CLI is unconfigured.

```
Configuration
----------------
Home:      
Workspace: 
Config:    
API key:   <not configured>
Find your API key at https://exercism.io/my/settings

API Reachability
----------------
Exercism:
    * /ping
    * [Get /ping: unsupported protocol scheme &#34;&#34;]
    * 52.222µs
```

The configuration section should be able to tell us what the home directory is, no matter what (that should not rely on configuring the CLI).

We can output what the default workspace _would_ be, as long as we clarify that it's not actually configured.

The `Config` field should show the default config directory, even if we haven't create it.

Lastly, we should use the default site URL for ping.

This PR fixes all of the above:

```
Configuration
----------------
Home:      /Users/kytrinyx
Workspace: /Users/kytrinyx/Test-On-Branch (default)
Config:    /Users/kytrinyx/.config/test-on-branch
API key:   <not configured>
Find your API key at https://exercism.io/my/settings

API Reachability
----------------
Exercism:
    * https://api.exercism.io/v1/ping
    * [connected]
    * 663.848701ms

```

It is probably easiest to review commit-by-commit.

Closes #671

### DIFF
--- a/cmd/troubleshoot.go
+++ b/cmd/troubleshoot.go
@@ -42,7 +42,9 @@ command into a GitHub issue so we can help figure out what's going on.
 		// Ignore error. If the file doesn't exist, that is fine.
 		_ = v.ReadInConfig()
 
-		status := newStatus(c, v)
+		cfg.UserViperConfig = v
+
+		status := newStatus(c, cfg)
 		status.Censor = !fullAPIKey
 		s, err := status.check()
 		if err != nil {
@@ -61,8 +63,8 @@ type Status struct {
 	System          systemStatus
 	Configuration   configurationStatus
 	APIReachability apiReachabilityStatus
+	cfg             config.Configuration
 	cli             *cli.CLI
-	cfg             *viper.Viper
 }
 
 type versionStatus struct {
@@ -99,10 +101,10 @@ type apiPing struct {
 }
 
 // newStatus prepares a value to perform a diagnostic self-check.
-func newStatus(c *cli.CLI, v *viper.Viper) Status {
+func newStatus(cli *cli.CLI, cfg config.Configuration) Status {
 	status := Status{
-		cli: c,
-		cfg: v,
+		cfg: cfg,
+		cli: cli,
 	}
 	return status
 }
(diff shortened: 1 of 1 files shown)


nywilken:
Is the `<not configured>` for the API token enough to denote that the CLI was never configured? Or should we be a bit more explicit with something like the following
```

Home:   <not configured>  (default /Users/kytrinyx)
Workspace:  <not configured>  (default /Users/kytrinyx/Test-On-Branch)
Config:    <not configured>  (default /Users/kytrinyx/.config/test-on-branch)

```

cmd/troubleshoot.go
```diff
@@ -112,7 +114,7 @@ func (status *Status) check() (string, error) {
 	status.Version = newVersionStatus(status.cli)
 	status.System = newSystemStatus()
 	status.Configuration = newConfigurationStatus(status)
-	status.APIReachability = newAPIReachabilityStatus(status.cfg.GetString("apibaseurl"))
+	status.APIReachability = newAPIReachabilityStatus(status.cfg)
```
nywilken: Passing the config here seems a bit heavy handed. What if instead of the original value you passed `config.InferSiteURL(v.GetString("apibaseurl"))`? Would that not achieve the same thing?

kytrinyx: No, the site URL is not the same as the API, so it would call https://exercism.io/ping (which would fail with a 404).

kytrinyx:
> Is the `<not configured>` for the API token enough to denote that the CLI was never configured? 

The `<not configured>` for the API token means that the API token is not configured. It doesn't speak for the rest of the values.

I went with `(default)` because if you don't explicitly overwrite the workspace, that is what will be written, but I think that your suggestion would be fine as well. I don't have very strong feelings either way. Which do you prefer?

nywilken:
> I went with (default) because if you don't explicitly overwrite the workspace, that is what will be written, but I think that your suggestion would be fine as well. I don't have very strong feelings either way. Which do you prefer?

If you modify the config file and remove only the workspace config the CLI reports that the API token is also not configured even if it is part of the config. It appears to me that we should should report <not configured> for the Workspace so that if a user opens an issue we can quickly see that their CLI is entirely unconfigured (I.e missing the most important part the workspace)

PR #676, opened 2018-07-24
Improve the assertions for the downloaded solution metadata

kytrinyx:
The hard-coded JSON for the metadata was annoying me. The big JSON is hard to read. We improved it a bit by writing the JSON as multiline, and then compacting it using `json.Compact()` but it still felt brittle.

This improves it further by reading the JSON into a solution struct and just checking the fields we care about.

### DIFF
--- a/cmd/download_test.go
+++ b/cmd/download_test.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -9,9 +8,11 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 
 	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/workspace"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


cmd/download_test.go
```diff
@@ -113,7 +113,7 @@ func TestDownload(t *testing.T) {
 		defer os.RemoveAll(tmpDir)
 		assert.NoError(t, err)
 
-		ts := fakeDownloadServer(tc.requestor)
+		ts := fakeDownloadServer(fmt.Sprintf("%v", tc.requester))
```
nywilken: `strconv.FormatBool(tc.requester)` is what you want here.

cmd/download_test.go
```diff
@@ -87,22 +87,22 @@ func TestDownload(t *testing.T) {
...
 			flags:       map[string]string{"exercise": "bogus-exercise"},
 		},
 		{
-			requestor:   requestorSelf,
+			requester:   true,
 			expectedDir: "",
 			flags:       map[string]string{"uuid": "bogus-id"},
 		},
 		{
-			requestor:   requestorOther,
```
nywilken: These constants are still defined below but no longer used

cmd/download_test.go
kytrinyx: Ooooh, love it. Thanks!

cmd/download_test.go
kytrinyx: Nice, thank you for catching that.

kytrinyx:
I fixed the last details and force pushed, @nywilken.

PR #677, opened 2018-07-25
Delete speculative implementation of prepare command

kytrinyx:
When I first started writing the new CLI I had some grand ideas about how submit should work,
which required all sorts of complicated guessing, which in turn required a command to
store settings on a per-track basis.

We may still need to prepare things on a per-track basis, but the current implementation was
highly speculative, and I decided that it would be better to delete it and start fresh
when we know exactly what we need.


nywilken:
So I reviewed these changes  and it looks good. I just wanted to do a sanity check to ensure that nothing we need is being deleted. But it doesn’t seem that way. 

Should you have one extra commit that deletes prepare.go since it is an empty comman? I’m suggesting it be a separate command so we can revert it when/if we wan to reimplement.

kytrinyx:
I would like to keep `prepare.go`, as having it does no harm, and I intend to fill it out with behavior. I just wanted to get rid of the things that I don't think we actually will end up using, or if we do will not be implemented in the same way. I don't feel like `prepare.go` complicates things, whereas the rest of it does.

PR #679, opened 2018-07-25
Improve the API ping in configure command

kytrinyx:
This distinguishes between an error in the HTTP request (e.g.
no service at URL), and an error returned from the API (e.g. 500 error).

Previous output in the case of an HTTP 500

```
$ ./testercism configure --api=http://localhost:3000/api/v1
Error: The base API URL 'http://localhost:3000/api/v1' cannot be reached.



```

Now:

```
$ ./testercism configure --api=http://localhost:3000/api/v1
Error: The base API URL 'http://localhost:3000/api/v1' cannot be reached.

API returned 500 Internal Server Error
```

### DIFF
--- a/cmd/configure.go
+++ b/cmd/configure.go
@@ -91,8 +91,7 @@ func runConfigure(configuration config.Configuration, flags *pflag.FlagSet) erro
 			return err
 		}
 
-		ok, err := client.IsPingable()
-		if !ok || err != nil {
+		if err := client.IsPingable(); err != nil {
 			return fmt.Errorf("The base API URL '%s' cannot be reached.\n\n%s", baseURL, err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/configure.go
```diff
@@ -93,7 +93,11 @@ func runConfigure(configuration config.Configuration, flags *pflag.FlagSet) erro
 
 		ok, err := client.IsPingable()
 		if !ok || err != nil {
```
nywilken: The IsPingable method seems to imply if there is an error than the client is not technically not reachable. If that is always going to be the case should we just check if `err != nil` and drop the `!ok`?

kytrinyx: Yeah, now that the IsPingable() returns an error, we can simplify.

kytrinyx:
I've tweaked based on your feedback, @nywilken

PR #686, opened 2018-07-27
Rename config.Configuration to config.Config

kytrinyx:
Now that we got rid of the complicated viper wrapper, we
can name this Config, which is a bit more idiomatic.


kytrinyx:
Good catch, yeah. I had to go with `config_notwin_test.go` rather than `config_test.go` since we already have a config_test.go that needs to be run for all OSes.

PR #693, opened 2018-08-04
Rename solution to exercise metadata

kytrinyx:
This is a pure rename, so while the diff is quite big, the change is small:

- `workspace.Solution` is now `workspace.ExerciseMetadata`
- `workspace.Solutions` is now `workspace.MetadataCollection`

What we previously called Solution is actually the exercise metadata, which also
includes the user's solution UUID.

In order to keep things as clear as possible, this also renames the Solutions
type to MetadataCollection. ExerciseMetadataCollection seemed to long,
and I couldn't think of anything better than collection to deal with the
multiples of a multiple.


nywilken:
> In order to keep things as clear as possible, this also renames the Solutions
type to MetadataCollection. ExerciseMetadataCollection seemed to long,
and I couldn't think of anything better than collection to deal with the
multiples of a multiple.

ExerciseMetadataCollection makes sense to me.

cmd/download.go
```diff
@@ -132,7 +132,7 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		}
 	}
 
-	solution := workspace.Solution{
+	solution := workspace.ExerciseMetadata{
```
nywilken: Would it be too much of a change, confusion, if we renamed this to metadata? If we can't change the name with out changing a ton of code right now what do you think about adding a `TODO`?

cmd/submit.go
```diff
@@ -130,7 +130,7 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
-	sx, err := workspace.NewSolutions(dirs)
+	sx, err := workspace.NewMetadataCollection(dirs)
```
nywilken: Lets call this one collection like you do in the tests.

cmd/download.go
kytrinyx: I went back and forth on `Metadata` vs `ExerciseMetadata` and was worried that `Metadata` was too unspecific. I like the shorter name better, though, so I'll go with that.

cmd/submit.go
kytrinyx: Oh, yes! I agree.

kytrinyx:
@nywilken I changed the name to `Metadata` instead of `ExerciseMetadata`. I also updated the local variables to match (including the `sx` that you found, and a couple of stray ones that were `solution` instead of `metadata`).

jdsutherland:
@nywilken I'm happy to adapt #630 to fruition. Having glanced at #696 and #693, it seems like the most sensible path forward would be to wait until they get merged and start a new PR off master since it seems so much of #630 will be stale due to refactoring.

Is that what you would suggest?

jdsutherland:
@kytrinyx No problem at all - it's useful to learn from how you improved things.

cmd/download.go
```diff
@@ -132,7 +132,7 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		}
 	}
 
-	solution := workspace.Solution{
+	metadata := workspace.Metadata{
```
nywilken: Something about this type is off to me , but I can't put my finger on why.  It may have to do with the two properties `IsRequester` and `Team.

cmd/download.go
nywilken: Sorry for not being explicit when I mentioned renaming this. I meant just changing `solution` to `metadata`. I'm on the fence about the name. `ExerciseMetadata` is clearer, but longer.

kytrinyx:
@nywilken I don't feel strongly either way. I'll close this and we can re-evaluate after the file gets moved.

nywilken:
Awesome. 🙌🏽

PR #695, opened 2018-08-05
Simplify open command

kytrinyx:
This no longer accepts just an exercise name. It mirrors the behavior of
the submit command where you pass the path to the exercise you want to
open in the browser.

In the future we can expand this to take --exercise, --track, and --team.

We don't have integration tests for this command, but I tested it locally by compiling the binary and opening one of my exercises.

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -1,15 +1,9 @@
 package cmd
 
 import (
-	"errors"
-	"fmt"
-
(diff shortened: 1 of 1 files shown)


kytrinyx:
Actually, I'm going to combine this change with another one where we get rid of the `Locate` functionality entirely. Ignore this!

Issue #661, opened 2018-07-20
Cannot download executeable from exercise solution.


desdulianto:
Hi, I opened this ticket because @kotp told me that when he downloaded my exercise solution of ruby track rna-transcription, it did not deliver executable file.

Here is the screenshot of the command history when I submitted the solution.
![image](https://user-images.githubusercontent.com/3291789/42983817-142140ec-8c13-11e8-94f1-5cbc1341d9a1.png)

I first submitted the solution when exercism haven't launch the new version yet. And after that I just manually download the exercise as instructed and submit it as usual.

Let me know if you need anything for helping closing the bug.

Thank you very much.

NB. I've learn much from the new mentor feature. Thanks very much exercism and the mentors.

kytrinyx:
@desdulianto Thanks so much for following up here!

Would you show me the output of `exercism troubleshoot`?

desdulianto:
Here it is @kytrinyx 
```bash
david@pegasus:~/exercism/ruby/rna-transcription-2$ exercism troubleshoot

Troubleshooting Information
===========================

Version
----------------
Current: 3.0.4
Latest:  3.0.5

Call 'exercism upgrade' to get the latest version.
See the release notes at https://github.com/exercism/cli/releases/tag/3.0.5 for details.


Operating System
----------------
OS:           linux
Architecture: amd64


Configuration
----------------
Home:      
Workspace: /home/david/exercism
Config:    /home/david/.config/exercism/user.json
API key:   bce0*****************************309

API Reachability
----------------

GitHub:
    * https://api.github.com
    * [connected]
    * 298.841827ms

Exercism:
    * https://api.exercism.io/v1/ping
    * [connected]
    * 1.256193759s


If you are having trouble please file a GitHub issue at
https://github.com/exercism/exercism.io/issues and include
this information.
```

is it because of my version is lagging?

kotp:
I believe a bug was found, but it does not hurt at all to update the client.

desdulianto:
Ok, I have upgraded the client and resubmit the solution, it's on 7th iteration now :D
Here is the command line output:
```bash
[root@pegasus rna-transcription-2]# exercism upgrade
[root@pegasus rna-transcription-2]# exit
exit
david@pegasus:~/exercism/ruby/rna-transcription-2$ exercism troubleshoot

Troubleshooting Information
===========================

Version
----------------
Current: 3.0.5
Latest:  3.0.5


Operating System
----------------
OS:           linux
[...]

PR #696, opened 2018-08-05
Get rid of the workspace.Locate behavior and comms package

kytrinyx:
This is big in the sense that it deletes a lot of code, but the core of the PR is three small simplifications:

- the open command - https://github.com/exercism/cli/commit/20ebf1956904f21a1aa9a255f898a08aa6ac5f7b
- the submit command - https://github.com/exercism/cli/commit/693b42910ed2913a1cbef98e8d5595c419480075
- the download command - https://github.com/exercism/cli/commit/dd6b948a28b3e98eff94cc676ade19ccc30fe2ca

Once all of those were simplified, we no longer used `SolutionPath`, `IsSolutionPath`, `ResolveSolutionPath`, or `Locate`, and could delete all of the functions and their associated tests. And the entire `comms` package!

This removes the behavior of adding a numeric suffix when downloading an exercise if the exercise already exists, which is the first step in fixing https://github.com/exercism/cli/issues/661
The second step is to make the download command more robust so it doesn't clobber existing solution files: https://github.com/exercism/cli/issues/697


cmd/download.go
```diff
@@ -59,21 +59,21 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
 	if err != nil {
 		return err
 	}
-	if uuid == "" && exercise == "" {
+	if uuid == "" && slug == "" {
 		return errors.New("need an --exercise name or a solution --uuid")
 	}
 
-	var slug string
+	var param string
```
nywilken: What is param supposed to represent an exercise `iteration`?

cmd/download.go
```diff
@@ -59,21 +59,21 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
 	if uuid == "" {
```
nywilken: Yon can simplify this a bit
```
param := "latest"
if uuid != "" {
 param = uuid
}
```

cmd/download.go
kytrinyx: param is the param to the URL that we're calling on the API. It's either the string `latest`, or an actual UUID.

cmd/download.go
kytrinyx: Thanks! Yeah, that's much better. Updated.

PR #703, opened 2018-08-11
Omit the leading slash on the relative path of exercise files

kytrinyx:
This updates the download and submit commands to no longer use a leading slash on the relative path for exercise files.

I added an extra file in the download test that has a leading slash, to show that we still handle the leading slash where present.

### DIFF
--- a/cmd/download_test.go
+++ b/cmd/download_test.go
@@ -170,6 +170,10 @@ func fakeDownloadServer(requestor, teamSlug string) *httptest.Server {
 		fmt.Fprint(w, "this is a special file")
 	})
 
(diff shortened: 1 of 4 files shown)


nywilken:
@kytrinyx I’ll take a deeper dive in the am and give it a proper review. The description implies updates to submit and download but I don’t see any changes to the download command init the test file.

kytrinyx:
@nywilken I added only a test for the download, that shows that it does the right thing already when given a file that doesn't have a leading slash.

nywilken:
I confirmed all works as expected even the case I called out. As it is a non-issue.

kytrinyx:
Thanks!

I am planning on doing some cleanup on the server site, just for consistency, but we're handling this better now than we were.

PR #704, opened 2018-08-11
Fix linting errors in cmd package

kytrinyx:
Not sure how I missed these before.


kytrinyx:
I think I haven't upgraded my Go in a while, as I'm not getting that lint error. I'll get my Go sorted, and think a bit on how to accomplish both making the linter happy and giving the end-user a nice, friendly, properly formatted message.

Issue #698, opened 2018-08-05
Distinguish between paths and filepaths for solutions


kytrinyx:
When submitting a solution we should ensure that the path that we send to the server is a normalized path relative to the exercise directory. This normalized path should always use forward slashes.

So for Hello World in Scala the path, when submitting from any operating system, would be:

```
src/main/scala/HelloWorld.scala
```

When downloading, we need to use Go's [`filepath.FromSlash`](https://golang.org/pkg/path/filepath/#FromSlash) to translate back to something that will work on the filesystem in question.

I will be writing a data transition on the backend to clean up all the existing filepaths.

Issue #665, opened 2018-07-22
Ensure filepaths are all normalized on submit/download


kytrinyx:
When submitting files on Windows, we need to ensure that the server receives normalized file separators (`/`), and when downloading files on windows, we need to ensure that any file paths are normalized to Windows path separators (`\`).

See https://github.com/exercism/exercism.io/issues/3870 for original report.

ptillemans:
I have an example for this: solution dae72e893ea042729b35608a16584ad1

A student submitted a solution for rust track challenge Pythagorean Triplets.

It displays fine on the site, but when downloading the solution contains:

``` rust
rust/pythagorean-triplet » cat src/lib.rs
pub fn find() -> Option<u32> {
    unimplemented!();
}
```

examining the content of the folder:

``` sh
rust/pythagorean-triplet » ls -al
total 40
drwxr-xr-x  9 pti  staff   288 24 jul 10:46 .
drwxr-xr-x  3 pti  staff    96 24 jul 10:46 ..
-rw-r--r--  1 pti  staff   249 24 jul 10:46 .gitignore
-rw-------  1 pti  staff   233 24 jul 10:46 .solution.json
-rw-r--r--  1 pti  staff    73 24 jul 10:46 Cargo.toml
-rw-r--r--  1 pti  staff  2421 24 jul 10:46 README.md
-rw-r--r--  1 pti  staff  1183 24 jul 10:46 \\src\\lib.rs
drwxr-xr-x  3 pti  staff    96 24 jul 10:46 src
drwxr-xr-x  3 pti  staff    96 24 jul 10:46 tests
```

note the \\src\\lib.rs

ptillemans:
Oops, sorry, didn't saw this was closed already. sorry.

kytrinyx:
This isn't closed -- it's being worked on.

Stargator:
I'm on a MacOS and 1 out of 5 of the first solutions I downloaded were like the one @ptillemans posted. But this is on the Dart track.

ErikSchierboom:
@kytrinyx Just out of curiosity: is this high on the priority list? It is my number one pain when using the CLI (which is otherwise working perfectly).

kytrinyx:
Yes, it's next on my list!

ErikSchierboom:
Fantastic!

PR #705, opened 2018-08-11
Distinguish between path and filepath

kytrinyx:
Use a normalized path relative to the exercise directory and using forward slashes when submitting to the server. This is converted to a proper filepath on the file system, using `/` or `\` depending on the operating system.

This builds on top of the pull request in https://github.com/exercism/cli/pull/703 -- too look at just the difference between the two branches, compare [no-leading-slash to path-vs-filepath](https://github.com/exercism/cli/compare/no-leading-slash...path-vs-filepath?expand=1).

Closes https://github.com/exercism/cli/issues/698
Closes https://github.com/exercism/cli/issues/665

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -162,7 +162,11 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 			fmt.Fprintf(Err, msg, file)
 			continue
 		}
-		exercise.Documents = append(exercise.Documents, workspace.NewDocument(exercise.Filepath(), file))
+		doc, err := workspace.NewDocument(exercise.Filepath(), file)
+		if err != nil {
+			return err
+		}
+		exercise.Documents = append(exercise.Documents, doc)
 	}
 
 	if len(exercise.Documents) == 0 {
@@ -178,7 +182,7 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	writer := multipart.NewWriter(body)
 
 	for _, doc := range exercise.Documents {
-		file, err := os.Open(doc.Filepath)
+		file, err := os.Open(doc.Filepath())
 		if err != nil {
 			return err
 		}
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -162,7 +162,7 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 			fmt.Fprintf(Err, msg, file)
 			continue
 		}
-		exercise.Documents = append(exercise.Documents, exercise.NewDocument(file))
+		exercise.Documents = append(exercise.Documents, workspace.NewDocument(exercise.Filepath(), file))
 	}
 
 	if len(exercise.Documents) == 0 {
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -8,7 +8,6 @@ import (
 	"mime/multipart"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/exercism/cli/api"
 	"github.com/exercism/cli/config"
@@ -125,6 +124,8 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		exerciseDir = dir
 	}
 
+	exercise := workspace.NewExerciseFromDir(exerciseDir)
+
 	solution, err := workspace.NewSolution(exerciseDir)
 	if err != nil {
 		return err
@@ -143,7 +144,7 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return fmt.Errorf(msg, BinaryName, solution.Exercise, solution.Track)
 	}
 
-	paths := make([]string, 0, len(args))
+	exercise.Documents = make([]workspace.Document, 0, len(args))
 	for _, file := range args {
 		// Don't submit empty files
 		info, err := os.Stat(file)
@@ -161,10 +162,10 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 			fmt.Fprintf(Err, msg, file)
 			continue
 		}
-		paths = append(paths, file)
+		exercise.Documents = append(exercise.Documents, exercise.NewDocument(file))
 	}
 
-	if len(paths) == 0 {
+	if len(exercise.Documents) == 0 {
 		msg := `
 
     No files found to submit.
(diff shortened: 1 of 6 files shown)


kytrinyx:
I've rebased this onto master.

nywilken:
Thanks @kytrinyx. I’ll have time to review later today.

workspace/document.go
```diff
@@ -0,0 +1,29 @@
...
+	return Document{
+		Root:     root,
+		Filepath: file,
+	}
+}
+
+// Path is the normalized path.
+// It uses forward slashes regardless of the operating system.
+func (doc Document) Path() string {
+	path := strings.Replace(doc.Filepath, doc.Root, "", 1)
```
nywilken: Are these two lines doing the same thing as `path, _ := filepath.Rel(doc.Root, doc.Filepath)`? Ignoring error right now.

workspace/document.go
```diff
@@ -0,0 +1,29 @@
...
+	"strings"
+)
+
+// Document is a file in a directory.
+type Document struct {
+	Root     string
+	Filepath string
+}
+
+// NewDocument creates a document from a filepath.
```
nywilken: Lets add a comment on what is Root and what is Filepath. It seems like Filepath is more of a relative path to the file.

cmd/submit.go
```diff
@@ -161,10 +162,10 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 			fmt.Fprintf(Err, msg, file)
 			continue
 		}
-		paths = append(paths, file)
+		exercise.Documents = append(exercise.Documents, exercise.NewDocument(file))
```
nywilken: Would it be clearer here if it was `workspace.NewDocument(exercise.Filepath, file)`

workspace/exercise.go
```diff
@@ -49,3 +59,8 @@ func (e Exercise) HasMetadata() (bool, error) {
 	}
 	return false, err
 }
+
+// NewDocument creates a document relative to the exercise.
+func (e Exercise) NewDocument(file string) Document {
```
nywilken: Method on this type really feels like it should be a single function or just deleted as workspace.NewDocument does the same thing.

workspace/document.go
```diff
@@ -0,0 +1,29 @@
...
+func NewDocument(root, file string) Document {
+	return Document{
+		Root:     root,
+		Filepath: file,
```
nywilken: Should this be Filename?

kytrinyx: It's a path relative to the root. So for clojure exercises it would be `src/anagram.clj`

cmd/submit.go
kytrinyx: I don't feel strongly either way.

workspace/exercise.go
kytrinyx: I'll go ahead and make the change--I really don't have a preference one way or the other.

workspace/document.go
kytrinyx: Oh. My. Godness.

I didn't know about `filepath.Rel`. That's _exactly_ what I need.

workspace/document.go
kytrinyx: Yepp, good call.

cmd/submit.go
```diff
@@ -176,18 +177,14 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	body := &bytes.Buffer{}
 	writer := multipart.NewWriter(body)
 
-	for _, path := range paths {
-		file, err := os.Open(path)
+	for _, doc := range exercise.Documents {
+		file, err := os.Open(doc.Filepath)
```
nywilken: I haven’t verified this yet. But I think we are losing the ability to submit via full path here. 

doc.Filepath is a relative path so if I submit outside of my track calling open here may result in file not exist error. We should add a test for TestSubmitWithFullPath

cmd/submit.go
```diff
@@ -176,18 +177,14 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
 			return err
 		}
 		defer file.Close()
 
-		dirname := fmt.Sprintf("%s%s%s", string(os.PathSeparator), solution.Exercise, string(os.PathSeparator))
-		pieces := strings.Split(path, dirname)
-		filename := pieces[len(pieces)-1]
-
-		part, err := writer.CreateFormFile("files[]", filename)
+		part, err := writer.CreateFormFile("files[]", doc.Path())
```
nywilken: `doc.Path()` now returns an error as a second value which now needs to be handled

cmd/submit.go
kytrinyx: Good catch! I'll add a test.

cmd/submit.go
kytrinyx: Dang. Why did this compile?

kytrinyx: I guess maybe it didn't but I was sure I had run all the tests...

kytrinyx:
I think I got this sorted out. I've moved the `filepath.Rel` logic into the `NewDocument()` function instead of keeping it in the normalized `Path()` method. @nywilken would you take another look at this?

workspace/document.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Document is a file in a directory.
+type Document struct {
+	Root         string
+	RelativePath string
+}
+
+// NewDocument creates a document from a relative filepath.
+// The root is typically the root of the exercise, and
+// path is the relative path to the file within the root directory.
```
nywilken: Path is no longer a relative path.  Path needs to be an absolute path, rooted with root, or filepath.Rel will return an error. 

calling filepath.Abs before calling filepath.Rel in this method is an option but it can result in unwanted behavior. https://play.golang.org/p/cLE1R9dsRcH


I think a `filepath.HasPrefix(root, path)`  check before .Rel would catch it.

kytrinyx:
>  If the documentation is clear we may not need to change anything. It really depends on how flexible we want the code to be.

I actually went back and forth on that a bit, but concluded that I don't want the flexibility until we find that we need it. Every time I get myself in trouble with complexity it's because I thought we needed flexibility that we really didn't need.

I'll fix the doc comment!

nywilken:
> I actually went back and forth on that a bit, but concluded that I don't want the flexibility until we find that we need it. 

Good call!  Comment is good and will save us time in the future.

kytrinyx:
Yay, thanks so much for your patience working through this one. I'll cut a release tonight.

PR #706, opened 2018-08-12
configure: fix reported lint errors

nywilken:
Go is a stickler when it comes to capitalizing the first word within an error string in case it gets consumed upstream for additional formatting. So instead of using the default error printing mechanism for Cobra. The
configure command now prints the token error message to Stderr and exits.

## What does it look like?
Before change
```
> golint cmd/...
cmd/configure.go:64:21: error strings should not be capitalized or end with punctuation or a newline
cmd/configure.go:114:21: error strings should not be capitalized or end with punctuation or a newline
cmd/configure.go:128:22: error strings should not be capitalized or end with punctuation or a newline
```

Now
```
> golint cmd/...

```


nywilken:
Need to rethink this a bit to not break tests.

kytrinyx:
Hm. Yeah... I know Go is a stickler, but since these errors are chat-y end-user messages I'm torn about how to accomplish that.

Issue #712, opened 2018-08-22
Can't download exercise by uuid with 3.0.7


angelikatyborska:
I have just updated to 3.0.7 via `exercism upgrade` and now every time I try to download an exercise by uuid, I get this:
```
$ exercism download --uuid=d7e094f32c9545a2863a851e25ba112a
Error: The exercise you specified could not be found
```

I tried a few different exercises (all Elixir) and it's all the same.

I was forced to downgrade to 3.0.6 to be able to continue mentoring. To anyone that has to downgrade too, I did it by running:
```
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/21d0da6bc774d1713c4542e4d7cf8adba2474a30/Formula/exercism.rb
```

Brew allows you to install any version of a formula by just passing the formula file.

nywilken:
@angelikatyborska thanks for calling this issue out. @kytrinyx I submitted a fix for this issue and assigned it to you for review. I tested locally and confirmed the download command is working with both `download --uuid 90cc8d0c5d5c494aad97a8c10b52f7cd` and `download --exercise rna-transcription --track ruby`

crane-yuan:
@nywilken Hello, I was in trouble when I was using exercism-cli under OSX. Please help me look at it.

And my exercism-cli version is 3.0.7.

I get this when i try to download exercise in the first time, but  it worked somehow when I tried again.
``` 
net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

![image](https://user-images.githubusercontent.com/8643491/44472040-e6dac100-a65f-11e8-8eeb-8f4f47041665.png)

kytrinyx:
I've released a fix for the `uuid` flag on the `download` command. I'm going to go ahead and close this.

@crane-yuan Please subscribe to https://github.com/exercism/cli/issues/701 which is the issue that you are describing.

crane-yuan:
@kytrinyx Thanks for your reply, I will try it.

PR #713, opened 2018-08-22
download: fix support for uuid flag

nywilken:
This change fixes the issue #712 where calling download with a solution UUID errors because the command is overriding the provided UUID with the string latest. This change also adds a mutual exclusive check for the uuid and exercise flag used by the download command.


nywilken:
> Shoot! I wish we had a failing test that caught this.

Agreed. I started looking into writing some tests this am. I’ll open a PR when ready.

kytrinyx:
@nywilken thanks! It was more critical to get the fix out than to wait for a test.

Issue #701, opened 2018-08-11
Linux: CLI- Configuration Error


MarkyMarkz:
I tried to configure CLI with my token, but got these errors:

```
[meck@lilsaint ~]$ exercism configure --token=<my token>
Error: The base API URL 'https://api.exercism.io/v1' cannot be reached.

Get https://api.exercism.io/v1/ping: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Sudo:

```
[meck@lilsaint ~]$ sudo exercism configure --token=<my token>
[sudo] password for meck: 
Error: The base API URL 'https://api.exercism.io/v1' cannot be reached.

Get https://api.exercism.io/v1/ping: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

Troubleshoot:

```
[meck@lilsaint ~]$ exercism troubleshoot

Troubleshooting Information
===========================

Version
----------------
Current: 3.0.6
Latest:  3.0.6


Operating System
----------------
OS:           linux
Architecture: amd64


Configuration
----------------
Home:      /home/meck
Workspace: /home/meck/exercism (default)
Config:    /home/meck/.config/exercism
API key:   <not configured>
Find your API key at https://exercism.io/my/settings

API Reachability
----------------

GitHub:
    * https://api.github.com
    * [connected]
    * 117.199064ms

Exercism:
    * https://api.exercism.io/v1/ping
    * [connected]
    * 272.994092ms


If you are having trouble please file a GitHub issue at
https://github.com/exercism/exercism.io/issues and include
this information.
```

Does someone know any solution to this?
Thanks in advance!

kytrinyx:
You should not need sudo.

This is interesting because it says that you're unable to reach the URL:

```
Get https://api.exercism.io/v1/ping: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

However when you call the troubleshoot command, you are able to connect. It's not fast, but it's less than a second:

```
Exercism:
    * https://api.exercism.io/v1/ping
    * [connected]
    * 272.994092ms
```

The timeout in the configure command is 10 seconds, whereas the one in the troubleshoot command is 20, which should not affect this.

You could try this:

```
exercism configure --token=<your token> --no-verify 
```

But if this is a timeout problem, you'd probably get a timeout again when calling `download`.

What does `time curl -v https://api.exercism.io/v1/ping` return?

MarkyMarkz:
[...]

Issue #719, opened 2018-08-26
Design question: what should the cli and api packages actually look like?


kytrinyx:
@nywilken I was looking at the contents of the `cli` package and the `api` package. I think we've got a bit of ambiguity around where things belong.

I'm wondering if we should have the CLI-specific _logic_ in the `cli` package (build stuff, logic around versions, maybe), and then have everything that talks to remote servers in `api`. But maybe `api` is the wrong name. Maybe `remote`?

nywilken:
> I'm wondering if we should have the CLI-specific logic in the cli package (build stuff, logic around versions, maybe), and then have everything that talks to remote servers in api. But maybe api is the wrong name. Maybe remote?

I think this makes complete sense. Maybe even have a pkg specific for the exercism api. I know I mentioned the upgrade pkg in the past but this might be a good opportunity to revisit. 

I’m blanking right now on what is specific to the cli. But it is worth breaking up. I have time this week to put together a rough proposal for the pkg break out.

kytrinyx:
> I have time this week to put together a rough proposal for the pkg break out.

That would be fantastic. I've been spending an hour or so each evening checking in on discussions, PRs, and issues, so I will have time to look at a proposal if you have one.

nywilken:
#### Current Structure
An outline of the current HTTPClient structure being used by the Exercism CLI

Packages which define an HTTPClient:
 ```
 githubm.com/exercism/cli/api
 githubm.com/exercism/cli/cli
```

Commands which use the defined HTTPClients:
 ``` 
  githubm.com/exercism/cli/cmd/configure
  githubm.com/exercism/cli/cmd/download
  githubm.com/exercism/cli/cmd/submit
  githubm.com/exercism/cli/cmd/troubleshoot
  githubm.com/exercism/cli/cmd/version
```

Import Dependencies:
```
  githubm.com/exercism/cli/api
  |_cli/api.go
  |_cmd/configure.go
  |_cmd/download.go
  |_cmd/submit.go

  githubm.com/exercism/cli/cli
  |_cmd/troubleshoot.go
  |_cmd/upgrade.go
  |_cmd/version.go
```
----
#### Proposed Structure
```
cmd/... (command files; minor changes)
api/ (Exercism configured API client; we can restructure further)
|_api.go
|_debug.go
pkg/ (Re-usable packages)
|_cli/ (great candidate for the GitHub Release Updater pkg; needs to be renamed)
   |_ cli.go
   |_ asset.go
[...]

PR #720, opened 2018-08-26
Make the timeout configurable globally

kytrinyx:
This tweaks the HTTP clients that we create to use a package variable for the timeout instead of hard-coding it.

Then it adds a persistent flag to the root command, which lets the 'pre-run' logic override the package variables if the flag is populated with something other than the zero value.

Closes https://github.com/exercism/cli/issues/701

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -15,7 +15,7 @@ var (
 	UserAgent = "github.com/exercism/cli"
 
 	// TimeoutInSeconds is the timeout the default HTTP client will use.
-	TimeoutInSeconds = 10
+	TimeoutInSeconds = 60
 	// HTTPClient is the client used to make HTTP calls in the cli package.
 	HTTPClient = &http.Client{Timeout: time.Duration(TimeoutInSeconds) * time.Second}
 )
--- a/api/client.go
+++ b/api/client.go
@@ -14,8 +14,10 @@ var (
 	// It's overridden from the root command so that we can set the version.
 	UserAgent = "github.com/exercism/cli"
 
-	// HTTPClient configures a timeout to use by default.
-	HTTPClient = &http.Client{Timeout: 10 * time.Second}
+	// TimeoutInSeconds is the timeout the default HTTP client will use.
+	TimeoutInSeconds = 10
+	// HTTPClient is the client used to make HTTP calls in the cli package.
+	HTTPClient = &http.Client{Timeout: time.Duration(TimeoutInSeconds) * time.Second}
 )
 
 // Client is an http client that is configured for Exercism.
--- a/api/client.go
+++ b/api/client.go
@@ -14,8 +14,8 @@ var (
 	// It's overridden from the root command so that we can set the version.
 	UserAgent = "github.com/exercism/cli"
 
-	// DefaultHTTPClient configures a timeout to use by default.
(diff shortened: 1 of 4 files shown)


api/client.go
```diff
@@ -14,8 +14,10 @@ var (
 	// It's overridden from the root command so that we can set the version.
 	UserAgent = "github.com/exercism/cli"
 
-	// DefaultHTTPClient configures a timeout to use by default.
-	DefaultHTTPClient = &http.Client{Timeout: 10 * time.Second}
+	// TimeoutInSeconds is the timeout the default HTTP client will use.
+	TimeoutInSeconds = 10
```
nywilken: Judging from the user comments we’ve seen pertaining to this issue, and from what I’ve seen after diving into http Client connection errors. I recommend we bump the default to 30s even 60s. 

I’m 75% sure the  issue is that the CLI is not giving the user enough time to complete the connection (dns, tls handshake, handle response, etc), may it be a slow DNS, slow response time from backend or slow Internet connections with large roundtip times. 

We can use the [httptrace pkg](https://golang.org/pkg/net/http/httptrace), maybe when passed the debug flag, to gain more insight on the issue. 


If you're wondering what's the 75% make: 25% user comments (which ruled out OS, CLI), 50% my research and local testing. Whats missing is the httptrace stats.

cmd/troubleshoot.go
```diff
@@ -30,7 +29,7 @@ If you're running into trouble, copy and paste the output from the troubleshoot
 command into a GitHub issue so we can help figure out what's going on.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		cli.HTTPClient = &http.Client{Timeout: 20 * time.Second}
+		cli.TimeoutInSeconds = 20
```
nywilken: What if we multiply the default value as opposed to hard coding to 20s?  If a user is having a problem pining the API and setting a timeout it would appear that the timeout is not enough, but in fact it is always 20s which the user may not know or care to look at the source.

cmd/root.go
```diff
@@ -39,6 +40,10 @@ Download exercises and submit your solutions.`,
 		if verbose, _ := cmd.Flags().GetBool("verbose"); verbose {
 			debug.Verbose = verbose
 		}
+		if timeout, _ := cmd.Flags().GetInt("timeout"); timeout > 0 {
+			cli.TimeoutInSeconds = timeout
+			api.TimeoutInSeconds = timeout
```
nywilken: I see what you’re talking about in issue #719. I will follow up shortly.

api/client.go
kytrinyx: I think that's fine. As long as we have _some_ default.

I'll set the default to 60s everywhere.

cmd/troubleshoot.go
kytrinyx: You mean, in troubleshoot say something like `cli.TimeoutInSeconds = cli.TimeoutInSeconds * 5` or something?

kytrinyx: Let's start with 2 and see how we fare :)

kytrinyx:
Alrighty, updated the timeouts per the discussion. Let me know what you think!

MarkyMarkz:
It is working now, thank you very much!

PR #721, opened 2018-08-26
Handle windows filepaths that accidentally got submitted to the server

kytrinyx:
Some of the older CLIs did not submit a normalized path for the file, but rather submitted the
filepath. This means that we sometimes had forward slashes and sometimes had backslashes.

The CLI now always submits the normalized path with forward slashes.

This ensures that when we receive files submitted with a buggy client we rewrite the paths
correctly.

Closes https://github.com/exercism/exercism/issues/4260
Closes https://github.com/exercism/exercism/issues/4101
Closes https://github.com/exercism/exercism/issues/4072

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -199,6 +199,10 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
 		// TODO: if there's a collision, interactively resolve (show diff, ask if overwrite).
 		// TODO: handle --force flag to overwrite without asking.
+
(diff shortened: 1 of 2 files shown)


kytrinyx:
> to actually fix on the backend the user would have to submit the solution, correct?

Yeah, though the latest release of the CLI does submit the correct one, I believe. I also have on my list to clean up old data on the server.

> At some point I got the double path error on Windows 

Damn, I wish I knew what situation we get that in, I've not yet got a good reproduction case to work from to fix that one.

Thanks for the review!

PR #722, opened 2018-08-26
 Handle exercise directories with numeric suffixes

kytrinyx:
An early design decision was to allow people to work on multiple solutions to the same
exercise at the same time. We later decided not to do this, but the CLI still had
logic that supported it.

This caused us to (incorrectly) create exercise directories that were named with a
numeric suffix, which would then get submitted to the backend server, which doesn't
care about filepaths.

Because the exercise with the numeric suffix doesn't match the expected path, we did
not correctly trim off any leading directories on submit, which further caused
the download command to put the solution in a weirdly and deeply nested directory,
making the solution hard to find and review.

Closes https://github.com/exercism/exercism/issues/4275
Closes https://github.com/exercism/exercism/issues/4066

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -9,6 +9,7 @@ import (
 	netURL "net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/exercism/cli/api"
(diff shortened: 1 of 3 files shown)


nywilken:
@kytrinyx I’m reviewing this, but I’m having a hard time verifying  what it is actually doing. Without this change the origin reported issue, at least on Linux, is part of the payload but gets skipped by the continue block. 

That aside the regex in the change  captures the whole file and replaces it with an empty string. Is that right or am I just reading this wrong? I’m going to give it another run through, I’m curious to know how the tests fail if we remove the regex from the download command.

kytrinyx:
Thanks @nywilken. I was a but unsure how much context to add to the comment.

PR #723, opened 2018-08-27
Metadata hidden subdir

### LINKED ISSUE #557: Access denied to .solution.json on Windows

### LINKED ISSUE #607: Put .solution.json in a hidden directory in the exercise

jdsutherland:
Continuation of #630
Closes #630
Closes #607
Closes #557


workspace/exercise_test.go
```diff
@@ -17,9 +17,9 @@ func TestHasMetadata(t *testing.T) {
 	exerciseA := Exercise{Root: ws, Track: "bogus-track", Slug: "apple"}
 	exerciseB := Exercise{Root: ws, Track: "bogus-track", Slug: "banana"}
 
-	err = os.MkdirAll(filepath.Join(exerciseA.Filepath()), os.FileMode(0755))
+	err = os.MkdirAll(filepath.Join(exerciseA.Filepath(), ignoreSubdir), os.FileMode(0755))
```
jdsutherland: Tests having to know to create the `.exercism` ignore subdirectory seems smelly. I'm not sure how to avoid it though.

workspace/workspace_test.go
```diff
@@ -60,10 +60,11 @@ func TestWorkspaceExercises(t *testing.T) {
 	b2 := filepath.Join(tmpDir, "track-b", "exercise-two")
 
 	for _, path := range []string{a1, a2, b1, b2} {
+		path := filepath.Join(path, ignoreSubdir)
 		err := os.MkdirAll(path, os.FileMode(0755))
 		assert.NoError(t, err)
 
-		if path != a2 {
+		if path != filepath.Join(a2, ignoreSubdir) {
```
jdsutherland: > Tests having to know to create the `.exercism` ignore subdirectory seems smelly. I'm not sure how to avoid it though.

cmd/submit_test.go
```diff
@@ -184,6 +185,54 @@ func TestSubmitFiles(t *testing.T) {
...
+	_, err = os.Stat(expectedPathAfterMigration)
+	assert.Error(t, err)
+
+	err = runSubmit(cfg, pflag.NewFlagSet("fake", pflag.PanicOnError), []string{file})
+	assert.NoError(t, err)
+	assert.Equal(t, "This is a file.", submittedFiles["file.txt"])
+
+	_, err = os.Stat(expectedPathAfterMigration)
+	assert.NoError(t, err)
+}
```
kytrinyx: This looks good. Can we also check that there's no file at the `legacy` location?

workspace/solution.go
```diff
@@ -88,3 +86,36 @@ func (s *Solution) PathToParent() string {
...
+		return err
+	}
+	if _, err := os.Lstat(metadataPath); err != nil {
+		if err := os.Rename(legacyMetadataPath, metadataPath); err != nil {
+			return err
+		}
+		fmt.Fprintf(os.Stderr, "\nMigrated solution metadata to %s\n", metadataPath)
+	} else {
+		// TODO: decide how to handle case where both legacy and modern metadata files exist
+		fmt.Fprintf(os.Stderr, "\nAttempted to migrate solution metadata to %s but file already exists\n", metadataPath)
```
kytrinyx: I think if they both exist, we delete the old one.

workspace/workspace_test.go
kytrinyx: I was thinking `os.MkdirAll` on the `filepath.Dir` of the full metadata file filepath  -- I'd rather do that than have to have the code know about the various pieces of the metadata file path.

workspace/exercise.go
```diff
@@ -37,7 +37,12 @@ func (e Exercise) Filepath() string {
 
 // MetadataFilepath is the absolute path to the exercise metadata.
 func (e Exercise) MetadataFilepath() string {
-	return filepath.Join(e.Filepath(), solutionFilename)
+	return filepath.Join(e.Filepath(), ignoreSubdirMetadataFilepath())
```
kytrinyx: How would you feel about making the relative path to the metadata filepath a variable defined at compile time rather than a function call?

```
var metadataFilepath = filepath.Join(".exercism", "solution.json")
```

workspace/workspace.go
```diff
@@ -113,9 +113,22 @@ func (ws Workspace) SolutionDir(s string) (string, error) {
 		if _, err := os.Lstat(path); os.IsNotExist(err) {
 			return "", err
 		}
-		if _, err := os.Lstat(filepath.Join(path, solutionFilename)); err == nil {
+		if err := checkMetadataFile(path); err == nil {
```
kytrinyx: I'm not quite comfortable with how this method not only checks the files, but also does the migration.

What do you think about inlining the checking logic into this function, so that it explicitly checks the metadata file path and returns if it's there, then checks the legacy metadata file path and returns if it's there.

Then, after we've found the solution directory, call a `workspace.MigrateLegacyMetadataFile()` function (or some such name), which is a noop if the metadata file isn't legacy.

workspace/solution.go
```diff
@@ -88,3 +86,36 @@ func (s *Solution) PathToParent() string {
...
+	path = filepath.Join(path, ignoreSubdir)
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if err := os.Mkdir(path, os.FileMode(0755)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func migrateLegacySolutionFile(legacyMetadataPath string, metadataPath string) error {
```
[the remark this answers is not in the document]
kytrinyx: Actually, having looked at the code, I think I'd like to export it now, but not for future use in the `doctor` command, but rather to explicitly call the migration method from the `submit` command.

cmd/submit_test.go
```diff
@@ -434,3 +483,19 @@ func writeFakeSolution(t *testing.T, dir, trackID, exerciseSlug string) {
 	err := solution.Write(dir)
 	assert.NoError(t, err)
 }
+
+func writeFakeLegacySolution(t *testing.T, dir, trackID, exerciseSlug string) {
```
kytrinyx: This is only used in one place, right? If so I'd rather inline it into the test.

workspace/exercise.go
jdsutherland: Sure. I'd made it a function call as a workaround to keep `const`.

workspace/workspace.go
jdsutherland: Yeah, I didn't like

workspace/exercise_test.go
jdsutherland: Good call. I think I fixed this up in all the tests that were explicitly knowing about metadata filepaths.

workspace/exercise_test.go
```diff
@@ -42,3 +42,19 @@ func TestNewFromDir(t *testing.T) {
 	assert.Equal(t, "the-track", exercise.Track)
 	assert.Equal(t, "the-exercise", exercise.Slug)
 }
+
+func TestMigrateLegacyMetadataFile(t *testing.T) {
```
jdsutherland: Since this is exported and dangerous (os.Rename & os.Remove) I imagine we want to unit test this, right?

These tests would need to setup both fake metadata (legacy/modern). `writeFakeSolution()` currently lives here https://github.com/exercism/cli/blob/32d8ff8c016e78ab49068681e084901d31a58056/cmd/submit_test.go#L426 and gets used a few places. I tried putting it in `exercise_test.go` and exporting it but that gives errors back in the `cmd` package. I have little experience with Go testing. If we want to share `WriteFakeSolution()` should we put it in `workspace/workspace.go` or would you suggest something else?

cmd/submit_test.go
```diff
@@ -184,6 +185,68 @@ func TestSubmitFiles(t *testing.T) {
...
+	assert.Error(t, err)
+
+	err = runSubmit(cfg, pflag.NewFlagSet("fake", pflag.PanicOnError), []string{file})
+	assert.NoError(t, err)
+	assert.Equal(t, "This is a file.", submittedFiles["file.txt"])
+
+	_, err = os.Stat(expectedPathAfterMigration)
+	assert.NoError(t, err)
+	_, err = os.Stat(exercise.LegacyMetadataFilepath())
+	assert.Error(t, err)
```
jdsutherland: Added check that legacy no longer exists.

jdsutherland:
I think all of the review concerns are now implemented. Assuming you're ok with `func (e Exercise) MigrateLegacyMetadataFile()` being a method on `Exercise`, I think the next step is the question I have about testing it.

jdsutherland:
I believe I've implemented all of the review concerns. I also attempted writing unit tests for the migration (I assume we want it unit tested since it's exported and potentially dangerous). I think it's ready for another round.

workspace/exercise.go
kytrinyx: Ah right. Personally, I'd rather have a var than a function call

workspace/exercise.go
```diff
@@ -59,3 +65,46 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+	}
+	if err == nil {
+		return true, nil
+	}
+	return false, err
+}
+
+// MigrateLegacyMetadataFile migrates a legacy metadata to the modern location.
+// This is a noop if the metadata file isn't legacy.
+// If both legacy and modern metadata files exist, the legacy file will be deleted.
```
kytrinyx: 👍 😍 This is such a great comment.

workspace/exercise.go
```diff
@@ -59,3 +65,46 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+	}
+	if err := createIgnoreSubdir(filepath.Dir(legacyMetadataFilepath)); err != nil {
+		return "", err
+	}
+	if _, err := os.Lstat(metadataFilepath); err != nil {
+		if err := os.Rename(legacyMetadataFilepath, metadataFilepath); err != nil {
+			return "", err
+		}
+		str = fmt.Sprintf("\nMigrated metadata to %s\n", metadataFilepath)
+		fmt.Fprintf(os.Stderr, str)
```
kytrinyx: I'd rather not print anything in packages other than the `cmd` package. Let's make this return some sort of state or conclusion (a custom type, maybe an iota?) that the command package (or whoever) can check to decide whether or not to print stuff.

workspace/exercise.go
```diff
@@ -59,3 +65,46 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+		return true, nil
+	}
+	return false, err
+}
+
+// MigrateLegacyMetadataFile migrates a legacy metadata to the modern location.
+// This is a noop if the metadata file isn't legacy.
+// If both legacy and modern metadata files exist, the legacy file will be deleted.
+func (e Exercise) MigrateLegacyMetadataFile() (string, error) {
+	var str string
```
kytrinyx: The `str` variable doesn't seem to be doing anything at the moment. Can we ditch it?

workspace/exercise.go
```diff
@@ -59,3 +65,46 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+		str = fmt.Sprintf("\nMigrated metadata to %s\n", metadataFilepath)
+		fmt.Fprintf(os.Stderr, str)
+	} else {
+		if err := os.Remove(legacyMetadataFilepath); err != nil {
+			return "", err
+		}
+		str = fmt.Sprintf("\nRemoved legacy metadata: %s\n", legacyMetadataFilepath)
+		fmt.Fprintf(os.Stderr, str)
+	}
+	return str, nil
```
kytrinyx: I think we can get rid of the `else` and move everything from the else down here.

workspace/exercise.go
```diff
@@ -59,3 +65,46 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+// MigrateLegacyMetadataFile migrates a legacy metadata to the modern location.
+// This is a noop if the metadata file isn't legacy.
+// If both legacy and modern metadata files exist, the legacy file will be deleted.
+func (e Exercise) MigrateLegacyMetadataFile() (string, error) {
+	var str string
+	legacyMetadataFilepath := e.LegacyMetadataFilepath()
+	metadataFilepath := e.MetadataFilepath()
+
+	if _, err := os.Lstat(legacyMetadataFilepath); err != nil {
+		return "", nil
```
kytrinyx: Is the assumption here that if we get an error, it's an `os.IsNotExist` error? Is that a safe assumption?

workspace/solution.go
```diff
@@ -67,17 +69,15 @@ func (s *Solution) Write(dir string) error {
...
 		return err
 	}
-
-	path := filepath.Join(dir, solutionFilename)
-
-	// Hack because ioutil.WriteFile fails on hidden files
-	visibility.ShowFile(path)
-
-	if err := ioutil.WriteFile(path, b, os.FileMode(0600)); err != nil {
+	if err = createIgnoreSubdir(dir); err != nil {
```
kytrinyx: I'd probably inline this. I know `createIgnoreSubdir` is used in a couple of places, but it doesn't feel like an actual abstraction to me, just an accidental bit of logic that happens to have to happen twice.

workspace/solution.go
```diff
@@ -88,3 +88,13 @@ func (s *Solution) PathToParent() string {
 	}
 	return filepath.Join(dir, s.Track)
 }
+
+func createIgnoreSubdir(path string) error {
+	path = filepath.Join(path, ignoreSubdir)
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if err := os.Mkdir(path, os.FileMode(0755)); err != nil {
```
kytrinyx: If you use `MkdirAll` instead of `Mkdir` you don't have to check that it doesn't exist first.

workspace/exercise.go
jdsutherland: Yeah, this is problematic. I replaced this by calling exercise.HasMetadata() since it contains the logic needed here. The trade-off is that this adds excess calls to MetadataFilepath() but it might be worth it for improved readability and avoiding duplication?

workspace/solution.go
jdsutherland: Agreed. As you mentioned below, using `MkdirAll` makes this just 2 lines.

workspace/exercise.go
```diff
@@ -59,3 +64,54 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+	}
+	return false, err
+}
+
+// MigrationStatus represents the result of migrating a legacy metadata file.
+type MigrationStatus int
+
+// MigrationStatus
+const (
+	MigrationStatusErrorMkdir MigrationStatus = iota
```
jdsutherland: It seems a bit odd to have an 'Error' status when we also have an `error` return type. I'm not sure if it's worth preserving for distinguishing these error cases as part of the status or if `error` should be relied on. If we want to ditch `ErrorX` statuses, should they be changed to `...Noop` or something else?

workspace/exercise.go
```diff
@@ -59,3 +64,53 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+		return false, nil
+	}
+	if err == nil {
+		return true, nil
+	}
+	return false, err
+}
+
+// MigrationStatus represents the result of migrating a legacy metadata file.
+type MigrationStatus int
```
jdsutherland: Do we want to add a `String()` method for consumers to print?

cmd/submit.go
```diff
@@ -125,7 +125,9 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	}
 
 	exercise := workspace.NewExerciseFromDir(exerciseDir)
-
+	if _, err = exercise.MigrateLegacyMetadataFile(); err != nil {
```
jdsutherland: Do we want to print a notice if the legacy metadata gets moved or deleted?

workspace/exercise.go
kytrinyx: > The trade-off is that this adds excess calls to MetadataFilepath() but it might be worth it for improved readability and avoiding duplication?

Yeah, I wouldn't worry about that. If we find that we have performance problems we can do proper profiling etc to see what's up and fix it 👍

cmd/submit.go
kytrinyx: How about only printing it if in `--verbose` mode?

workspace/exercise.go
kytrinyx: I wouldn't worry about having statuses for the errors since (as you mention) we return the error.

I think `MigrationStatusNoop` is fine to return in the error condition, since it would not have actually changed anything if there is an error. The name is fine, too, I think. Could perhaps do `Unchanged` instead of `Noop` but I actually don't feel strongly either way.

workspace/exercise.go
```diff
@@ -59,3 +64,53 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+
+// MigrateLegacyMetadataFile migrates a legacy metadata file to the modern location.
+// This is a noop if the metadata file isn't legacy.
+// If both legacy and modern metadata files exist, the legacy file will be deleted.
+func (e Exercise) MigrateLegacyMetadataFile() (MigrationStatus, error) {
+	if ok, _ := e.HasLegacyMetadata(); !ok {
+		return MigrationStatusNoop, nil
+	}
+	legacyMetadataFilepath := e.LegacyMetadataFilepath()
+	metadataFilepath := e.MetadataFilepath()
```
kytrinyx: Let's just call these inline. It's a couple of extra operations, potentially, but they're really cheap, and I'd rather not have to keep track of the local variables.

workspace/exercise_test.go
```diff
@@ -42,3 +67,92 @@ func TestNewFromDir(t *testing.T) {
...
+	err = ioutil.WriteFile(metadataFilepath, []byte{}, os.FileMode(0600))
+	assert.NoError(t, err)
+
+	ok, _ := exercise.HasLegacyMetadata()
+	assert.False(t, ok)
+	ok, _ = exercise.HasMetadata()
+	assert.True(t, ok)
+
+	status, err := exercise.MigrateLegacyMetadataFile()
+	assert.Equal(t, MigrationStatus(MigrationStatusNoop), status)
```
kytrinyx: You shouldn't need this type cast, since MigrateLegacyMetadataFile() returns a MigrationStatus not an int.

workspace/exercise_test.go
```diff
@@ -42,3 +68,86 @@ func TestNewFromDir(t *testing.T) {
 	assert.Equal(t, "the-track", exercise.Track)
 	assert.Equal(t, "the-exercise", exercise.Slug)
 }
+
+func TestMigrateLegacyMetadataFileWithoutLegacy(t *testing.T) {
```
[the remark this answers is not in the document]
kytrinyx: Yeah, I think these are great. I'm all for explicit, thorough tests :)

You could try turning it into a single table test if you wanted to--not sure what that would end up looking like (sometimes it's better, sometimes it's worth).

cmd/submit.go
```diff
@@ -125,7 +125,13 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	}
 
 	exercise := workspace.NewExerciseFromDir(exerciseDir)
-
+	migrationStatus, err := exercise.MigrateLegacyMetadataFile()
+	if err != nil {
+		return err
+	}
+	if verbose, _ := flags.GetBool("verbose"); verbose {
+		fmt.Fprintf(os.Stderr, migrationStatus.String(exercise))
```
jdsutherland: Should this be covered in tests?

workspace/exercise.go
```diff
@@ -59,3 +65,59 @@ func (e Exercise) HasMetadata() (bool, error) {
...
+type MigrationStatus int
+
+// MigrationStatus
+const (
+	MigrationStatusNoop MigrationStatus = iota
+	MigrationStatusMigrated
+	MigrationStatusRemoved
+)
+
+func (m MigrationStatus) String(e Exercise) string {
```
jdsutherland: Is it awkward that this takes an `Exercise`. I'm not sure if there's a better way to do this.

workspace/solution.go
```diff
@@ -67,17 +69,15 @@ func (s *Solution) Write(dir string) error {
...
 		return err
 	}
-
-	path := filepath.Join(dir, solutionFilename)
-
-	// Hack because ioutil.WriteFile fails on hidden files
-	visibility.ShowFile(path)
-
-	if err := ioutil.WriteFile(path, b, os.FileMode(0600)); err != nil {
+	metadataAbsoluteFilepath := NewExerciseFromDir(dir).MetadataFilepath()
```
jdsutherland: 'metadataAbsoluteFilepath' is a bit awkward but 'metadataFilepath' would shadow the var in Solution.

workspace/workspace.go
```diff
@@ -113,7 +113,10 @@ func (ws Workspace) SolutionDir(s string) (string, error) {
 		if _, err := os.Lstat(path); os.IsNotExist(err) {
 			return "", err
 		}
-		if _, err := os.Lstat(filepath.Join(path, solutionFilename)); err == nil {
+		if _, err := os.Lstat(filepath.Join(path, metadataFilepath)); err == nil {
+			return path, nil
+		}
+		if _, err := os.Lstat(filepath.Join(path, legacySolutionFilename)); err == nil {
```
jdsutherland: Is it ok that above we're using the name `metadataFilepath` but here, `legacySolutionFilename`? Will we presumably clean up all related incoherencies in the future rename PR?

workspace/exercise.go
kytrinyx: Only if we need it now. If not, we can add it when we need it.

workspace/exercise.go
kytrinyx: >  Making it a method on Exercise doesn't seem right.

Why is that? I actually think that seems fine, but I fear that I'm missing something.

workspace/workspace.go
kytrinyx: > Will we presumably clean up all related solution/metadata incoherencies in the future rename PR?

Yeah, I think it makes sense to do that separately.

cmd/submit.go
kytrinyx: Oh, I see what you mean by it seems wrong that it takes an exercise! Yes, somehow I didn't see that it was the `String()` method.

Yes, I agree that this isn't ideal.

workspace/exercise.go
kytrinyx: No, let's not include the metadata location.

jdsutherland:
I think it's ready.

jdsutherland:
I think it's ready. Do you want me to squash some of this?

kytrinyx:
I'll go ahead and squash and merge on my end.

Thank you ever so much for your patience in working through this! ❤️ 💛 💚 💙 💜 🖤

nywilken:
@jdsutherland great stuff on getting this in. I'm glad @kytrinyx was able to help you get this into a good place. With that said, we should probably test, maybe as a pre-release, to ensure exercises get migrated over successfully. 

@kytrinyx thoughts?

kytrinyx:
Yepp--we can test this by building locally:

```
go build -o testercism exercism/main.go
```

Then configuring with `./testercism configure`, and then running a download and submit against the live site.

@jdsutherland want to give this a first run?

nywilken:
@kytrinyx I was thinking of making a pre-release version available and asking members of the Exercism group to test as well; if they are up for it. Is that okay with you?

jdsutherland:
I tested this locally to ensure stderr output cases were correct but I don't have an exhaustive local workspace that would function as an effective fixture.

kytrinyx:
@nywilken that would be great.

nywilken:
@jdsutherland @kytrinyx I pushed out an alpha release so that we can test this furhter. Please feel free to download and give it a run (https://github.com/exercism/cli/releases/tag/v3.0.10-alpha.1).

There are no breaking changes here, but I would like to document how to roll back changes once done with the testing. Seeing as 3.0.9 binaries won't find the newly migrated metadata file. So I will make an official issue in the AM with testing steps.

jdsutherland:
> There are no breaking changes here, but I would like to document how to roll back changes once done with the testing.

A shell script could create a copy workspace to test with avoiding touching the actual workspace. Are you suggesting a rollback mechanism incase there is an issue after live rollout? What do you have in mind?

PR #724, opened 2018-08-29
Update submit error message when submitting a directory

sfairchild:
This it updating the error message when trying to submit a directory to help new students.
Issue: [https://github.com/exercism/exercism/issues/4309](https://github.com/exercism/exercism/issues/4309)

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -87,6 +87,8 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
         %s
 
+    Please provide the path to the file you wish to submit, e.g. : exercism submit FILENAME
+
 			`
 			return fmt.Errorf(msg, arg)
 		}
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -87,7 +87,9 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
         %s
 
-    Please provide the path to the file you wish to submit, e.g. : %s submit FILENAME
+    Please change into the directory and provide the path to the file(s) you wish to submit
+
+        %s submit FILENAME
 
 			`
 			return fmt.Errorf(msg, arg, BinaryName)
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -87,10 +87,10 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
         %s
 
-    Please provide the path to the file you wish to submit, e.g. : exercism submit FILENAME
+    Please provide the path to the file you wish to submit, e.g. : %s submit FILENAME
 
 			`
-			return fmt.Errorf(msg, arg)
+			return fmt.Errorf(msg, arg, BinaryName)
 		}
 
 		src, err := filepath.EvalSymlinks(arg)
--- a/cmd/submit_test.go
+++ b/cmd/submit_test.go
@@ -123,7 +123,7 @@ func TestSubmitFilesAndDir(t *testing.T) {
 	}
 	err = runSubmit(cfg, pflag.NewFlagSet("fake", pflag.PanicOnError), files)
 	assert.Regexp(t, "submitting a directory", err.Error())
-	assert.Regexp(t, "Please provide the path to the file you wish to submit", err.Error())
+	assert.Regexp(t, "Please change into the directory and provide the path to the file\(s\) you wish to submit", err.Error())
 }
 
 func TestSubmitFiles(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cmd/submit.go
```diff
@@ -87,6 +87,8 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
         %s
 
+    Please provide the path to the file you wish to submit, e.g. : exercism submit FILENAME
```
nywilken: We have a variable called `BinaryName` that contains the name of the command. You should use that as opposed to hard coding `exercism` https://github.com/exercism/cli/blob/master/cmd/submit.go#L141

nywilken:
@sfairchild \please update the formatting so that it is consistent with the message format being used for the download message error

https://github.com/exercism/cli/blob/e9877e2a172e4004124983488918ec68292fe9d3/cmd/submit.go#L138-L144

sfairchild:
Thanks for the feedback. I updated and pushed the changes

cmd/submit.go
```diff
@@ -87,8 +87,10 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
         %s
 
+    Please provide the path to the file you wish to submit, e.g. : %s submit FILENAME
```
nywilken: One more minor nitpick. We generally put the example command on a separate line to avoid it being pushed off the terminal window (longer strings) and to make is easier for copying/pasting. 

Can we update the message so that it reads and looks like the following.
```
    You are submitting a directory, which is not currently supported.

        /home/wilken/Development/exercism/vimscript

    Please change into the directory and provide the path to the file(s) you wish to submit

        exercism submit FILE1 FILE2
```

Note: I am making the assumption that if a user is trying to submit a directory they probably want to submit multiple files; hence the reason for calling out FILE1 FILE2. If that is not the case feel free to just keep FILENAME.

cmd/submit.go
```diff
@@ -87,8 +87,10 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+
 			`
-			return fmt.Errorf(msg, arg)
+			return fmt.Errorf(msg, arg, BinaryName)
```
nywilken: :+1:

cmd/submit.go
sfairchild: The use case for this was a new student trying to submit hello_world for the very first time and not knowing how to fix their issue with just the line 

```You are submitting a directory, which is not currently supported.```

Listing the multiple files might be helpful for letting users know that you can submit more than 1 file. I didn't even know this was an option until I dug into the source. My fear is listing multiple files will be confusing for someone who if very new. I am happy to make the change based on what everyone else thinks. @kytrinyx

```

    Please change into the directory and provide the path to the file(s) you wish to submit

        exercism submit FILE1 FILE2
```
vs
```

    Please change into the directory and provide the path to the file(s) you wish to submit

        exercism submit FILE
```

Comment (nywilken) [reply cmd/submit.go]: Let’s  drop the multiple files message. I agree that it can be confusing. Which is why I called out the note.

Comment (NobbZ) [reply cmd/submit.go]: It is a common convention to put optional arguments (named and positional) into square brackets, also using ellipsis to signal "or more", so that line should IMHO read `exercism submit FILE1 [FILE2 ...]`

Comment (NobbZ) [reply cmd/submit.go]: But I think, a single file example is sufficient if we extend the usage (`exercism help submit`) as described below and remind in the error that you can get more info there.

1. be more explicit about the fact that at least one file is necessary
2. be more explicit that one or more files can be specified

The current help output only states in the description that a list of files is needed, the actual usage does not mention it, but flags only as optional (square brackets).

Comment (sfairchild) [reply cmd/submit.go]: I agree with @NobbZ that the documentation for submit could be improved. Maybe a seperate issue could be opened to update that.

```
➜  ~ exercism version
exercism version 3.0.9
➜  ~ exercism submit -h
Submit your solution to an Exercism exercise.

	Call the command with the list of files you want to submit.

Usage:
   submit [flags]

Aliases:
  submit, s

Flags:
  -h, --help   help for submit

Global Flags:
      --timeout int   override the default HTTP timeout (seconds)
  -v, --verbose       verbose output
```

For now I updated the error message to 
```
    Please change into the directory and provide the path to the file(s) you wish to submit

        exercism submit FILENAME
```

nywilken:
> so that line should IMHO read exercism submit FILE1 [FILE2 ...]

@NobbZ thanks for pointing this out. That is the right way and we should definitely add that output to our usage details. 

> extend the usage ( exercism help submit ) ... and remind in the error that you can get more info there. 

Let’s open a separate PR for this one. If you come across other info that should be part of the usage, please let us know. Ty

nywilken:
@sfairchild I got caught up working on issue #719 and forgot to open. Ty.

PR #725, opened 2018-08-29
717 submit enormous files

### LINKED ISSUE #717: Do not allow submitting enormous files
The `file_contents` column in the database is a MySQL `BINARY` type (65,535 bytes).

https://github.com/exercism/website/blob/40515d13b4a8ae177aea8813eae86ca9229eae81/db/schema.rb#L137

See https://github.com/exercism/exercism/issues/4277 for the original report.

sfairchild:
Closes #717 

Throws an error if file is larger than 65535 bytes

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -151,13 +151,14 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		if err != nil {
 			return err
 		}
-		if info.Size() >= int64(65535) {
+		const MaxFileSize int64 = 65535
+		if info.Size() >= MaxFileSize {
 			msg :=`
 
-      The file you are trying to submit is %d bytes. Please reduce the file to below 65535 bytes and try again.
+      The submitted file is larger than the max allowed file size of %d bytes. Please reduce the size of the file and try again.
 
 			`
-			return fmt.Errorf(msg, info.Size())
+			return fmt.Errorf(msg, MaxFileSize)
 		}
 		if info.Size() == 0 {
 
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -151,14 +151,14 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		if err != nil {
 			return err
 		}
-    if info.Size() >= int64(65535) {
+		if info.Size() >= int64(65535) {
 			msg :=`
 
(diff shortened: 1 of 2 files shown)


cmd/submit_test.go
```diff
@@ -230,6 +230,48 @@ func TestSubmitWithEmptyFile(t *testing.T) {
...
+		Persister:       config.InMemoryPersister{},
+		UserViperConfig: v,
+	}
+
+	file := filepath.Join(dir, "file.txt")
+	err = ioutil.WriteFile(file, make([]byte, 65535), os.FileMode(0755))
+
+	err = runSubmit(cfg, pflag.NewFlagSet("fake", pflag.PanicOnError), []string{file})
+
+	assert.Regexp(t, "Please reduce the file to below 65535 bytes and try again.", err.Error())
```
nywilken: See my suggestion above about the messaging and using a const variable.

sfairchild:
Thank you @nywilken. Go isn't my primary language and I just started learning it. I updated based on your comments and let me know if you have other concerns.

cmd/submit_test.go
```diff
@@ -230,6 +230,52 @@ func TestSubmitWithEmptyFile(t *testing.T) {
...
+	v.Set("workspace", tmpDir)
+	v.Set("apibaseurl", ts.URL)
+
+	cfg := config.Config{
+		Persister:       config.InMemoryPersister{},
+		UserViperConfig: v,
+	}
+
+	file := filepath.Join(dir, "file.txt")
+	err = ioutil.WriteFile(file, make([]byte, 65535), os.FileMode(0755))
```
nywilken: I like that we don't use `maxFileSize` here. This test helps gives context to what is considered an enormous file. Good stuff.

cmd/submit.go
```diff
@@ -151,6 +151,15 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		if err != nil {
 			return err
 		}
+		const MaxFileSize int64 = 65535
```
nywilken: In Go variable names starting with a capital letter usually denote exported variables (shared across the package). Given the scope in which this constant is defined it can not be exported. 

In an attempt to keep things inline with the Code review section for Go https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps this const should be called `maxFileSize`.

PR #727, opened 2018-08-31
WIP: general working propsal for separating HTTP services in the CLI

nywilken:
This particular change is a WIP, proposal, for restructuring the HTTP clients being used by the Exercsim CLI. The rationale behind the changes found in the PR have been outlined in the issue https://github.com/exercism/cli/issues/719#issuecomment-417534271


api/api.go
```diff
@@ -1,51 +1,102 @@
...
 
-	encoding, _, certain := charset.DetermineEncoding(b, mimeType)
-	if !certain {
-		// We don't want to use an uncertain encoding.
-		// In particular, doing that may mangle UTF-8 files
-		// that have only ASCII in their first 1024 bytes.
-		// See https://github.com/exercism/cli/issues/309.
-		// So if we're unsure, use UTF-8 (no transformation).
-		s := string(b)
-		return &s, nil
```
kytrinyx: Are we losing something here with the removal of the uncertain encoding check? Or was it not being used?

cmd/root.go
```diff
@@ -38,12 +45,16 @@ Download exercises and submit your solutions.`,
...
-			cli.TimeoutInSeconds = timeout
-			api.TimeoutInSeconds = timeout
+			TimeoutInSeconds = timeout
 		}
+
+		httpClient = &http.Client{Timeout: time.Duration(TimeoutInSeconds) * time.Second}
+		// share the configured HTTPClient with other web related services
+		api.HTTPClient = httpClient
+		cli.HTTPClient = httpClient
 	},
```
kytrinyx: I really like this. It's so much cleaner.

pkg/cli/cli.go
```diff
@@ -15,7 +15,6 @@ import (
 	"time"
 
 	"github.com/blang/semver"
-	"github.com/exercism/cli/debug"
```
kytrinyx: I like that we don't have the debug package here anymore. That feels cleaner.

pkg/cli/cli.go
```diff
@@ -43,10 +42,8 @@ var (
 )
 
 var (
-	// TimeoutInSeconds is the timeout the default HTTP client will use.
-	TimeoutInSeconds = 60
 	// HTTPClient is the client used to make HTTP calls in the cli package.
-	HTTPClient = &http.Client{Timeout: time.Duration(TimeoutInSeconds) * time.Second}
+	HTTPClient = &http.Client{Timeout: 60 * time.Second}
```
kytrinyx: This is the default that gets overridden by the `cmd.httpClient`, correct?

cmd/troubleshoot.go
```diff
@@ -201,7 +204,7 @@ func (ping *apiPing) Call(wg *sync.WaitGroup) {
 	defer wg.Done()
 
 	now := time.Now()
-	res, err := cli.HTTPClient.Get(ping.URL)
+	res, err := httpClient.Get(ping.URL)
```
kytrinyx: Yeah, this feels cleaner.

api/api.go
nywilken: I didn’t see it being used anywhere so I removed it. It seems like an artifact from CLI 2, but I have yet to confirm.

pkg/cli/cli.go
nywilken: Yeah me too. It had one odd sprinkle in the previous cli pkg. I like having the option to create a custom API client with all the bells and whistles that we may want.

api/api.go
kytrinyx: Ooooh, this makes sense. We used to serialize the files into JSON. Now we're just sending the raw files.

nywilken:
@kytrinyx I'll start working on cleaning things up a bit here. Thanks for the comments.

kytrinyx:
@nywilken will you let me know when you're ready for another review?

nywilken:
@kytrinyx will do. Sorry for the longer than normal delay.

nywilken:
Given its age I think it’s safe to close and revisit if there is still a need.

Issue #637, opened 2018-07-14
Path issues with a mac (capitalisation)


NobbZ:
In exercism/exercism.io#3820 we discovered that when beeing in `~/exercism/javascript/hello-world` canceled submission because of not beeing in the workspace, as the configured workspace was `~/Exercism` (capital `E`!). Downloading exercises worked and files were accessible at the aforementioned uncapitalized path.

It seems as if we need to normalize pathes on mac as well similar how it is planned or already done for windows.

nywilken:
@nobbz thanks for calling this out. We have not yet fixed this issue, but it’s on our radar. 

File paths on configure are titlized but we don’t normalize for paths on submit.

kytrinyx:
@NobbZ, yes! I completely agree. We need to make sure that we pay attention to what is configured and ensure that _that_ is what we use to submit.

kytrinyx:
@NobbZ If you have a moment, would you help fleshing out the reproduction steps we need to ensure that we fix this correctly both for the case sensitive and case insensitive platforms?

The reported case was:

- workspace directory is capitalized
- CLI is configured with capitalized directory name (matches actual name)
- exercise is submitted using lowercase for workspace

It sounds like we want two different outcomes, depending on whether or not the filesystem is case sensitive or not.

1. Case insensitive file system: the exercise is submitted
1. Case sensitive file system: we get an error saying that the file that is being submitted it outside the workspace.

My mac is case sensitive, but only to a certain degree:

If I have a directory named `test`, then:

- `mkdir Test` will fail with `File exists`
- `ls test` will show `test`
- `ls te<tab>` will auto-complete
- `ls Test` will show `Test`
- `ls Te<tab>` will not auto-complete

So fundamentally, Mac is case insensitive in that it will not allow you to have two directories where the name differs only in capitalization.

I think that the right way to move forward in this is to have two separate test files, one that compiles on linux, and the other that compiles for non-linux.

kytrinyx:
Note from @NobbZ:

> I googled a bit in the meantime, perhabs the `HasFilepathPrefix` function from this internal `dep` module can help you? https://github.com/golang/dep/blob/master/internal/fs/fs.go

This looks exactly like what we need.

kotp:
Bumping this issue, as quite a few "support requests" come about regarding this issue.

PR #730, opened 2018-09-03
[WIP] Checking whether user is in workspace shouldn't be case-sensitive 🤔

andrewlouis93:
Here's a problem I observed: 

```
➜  cli (workspace-case-fix) ✔ exercism workspace
/Users/andrewlouis/Exercism
```

Workspace directory was actually created with lowercase though (if this is not the expectation - possible because this was the case in an earlier version?):

```
➜  ~ ls | grep exercism
exercism
```

And so when trying to submit from my workspace, I'd repeatedly see: 

image

cd-ing into `Exercism` addressed the problem, so I figured this is since no normalization is occurring before validating that the user is in the workspace. The test in the first commit verifies this. 

Please take a look, let me know what y'all think.

### DIFF
--- a/workspace/workspace_test.go
+++ b/workspace/workspace_test.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -103,6 +104,10 @@ func TestSolutionDir(t *testing.T) {
 			ok:   true,
 		},
 		{
+			path: strings.ToUpper(filepath.Join(ws.Dir, "exercise")),
+			ok:   true,
+		},
+		{
 			path: filepath.Join(ws.Dir, "exercise", "file.txt"),
 			ok:   true,
 		},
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks for tackling this! The related issue is in https://github.com/exercism/cli/issues/637

The problem here is that while Windows and Mac are both case insensitive, linux is not, so we're going to need separate behavior for the different operating systems, as described in that issue.

andrewlouis93:
Thanks for the insight @kytrinyx, let me read up on the issue and try to move a fix along :)

kytrinyx:
@andrewlouis93 I'm just seeing this now, as I didn't see a notification that this was ready for review. In the future, would you please make sure to add a comment when you've made changes that have addressed any concerns that were brought up?

kytrinyx:
I took a look at this and I'm not sure I understand how it addresses the issue. @andrewlouis93 when you have a moment would you let me know what the state of this is, and how you're tackling the change?

kytrinyx:
As it's been a couple of months since we've heard from @andrewlouis93 on this, I'm going to close the PR. @andrewlouis93 if you do want to continue working on it with the alternate direction discussed above, please let us know.

PR #739, opened 2018-09-19
Update max allowed file size error to include filename

nywilken:
Currently when submitting a file that is larger than the max allowed size the user receives an error indicating that a submitted file is larger than the max allowed file size, but it does not include the actual name of the file.

This is not really an issue when submitting a single file, but when a user submits two or more files the error message is a bit of a guessing game.


nywilken:
Thanks for the review.

PR #744, opened 2018-09-27
[WIP] Add doctor cmd - migrate legacy exercises

### LINKED ISSUE #639: Implement behavior to migrate legacy exercises on the filesystem
Is atool as described in the subject planned, available or otherwise easily possible or not wanted at all?

### LINKED ISSUE #699: Implement behavior to clean up numeric suffixes in exercise directories
When we launched the v2 site, the CLI had some funky behavior that sometimes adds a numeric suffix to the exercise directory. This behavior is going away.

The `exercism doctor` command should clean any of these up.

We can rely on the `workspace.PotentialExercises()` functionality to do this, and then group them into each group of related exercises. E.g.

- `ruby/hello-world`
- `ruby/hello-world-2`
- `ruby/hello-world-3`

These would all be in one group.

For each group where there are any exercises with a suffix we should have an interactive prompt. Something like:

```
Which one would you like to keep? Select your choice by typing the number.
1. hello-world
2. hello-world-2
3. hello-world-3
4. Skip this one. Do nothing and move on.
```

If they select one of the exercises (not _skip_), then:

- delete all but the selected one
- rename the selected one to remove the suffix (if it has one)

## Additional error handling

In the `download` command we should bail with an error if trying to download to an exercise directory with a numeric suffix. 

In the `submit` command we should bail with an error if trying to submit an exercise that is in a directory with a numeric suffix.

The error messages should suggest running `exercism doctor`.

jdsutherland:
Closes #639
~~Closes 699~~ - decided [clean up numeric suffixes](https://github.com/exercism/cli/issues/699) should be split into separate PR.

Still a rough WIP.

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -1,8 +1,14 @@
 package cmd
 
 import (
+	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
+	"path/filepath"
+	"strings"
 
+	"github.com/exercism/cli/api"
 	"github.com/exercism/cli/config"
 	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
@@ -37,15 +43,31 @@ var doctorCmd = &cobra.Command{
 		// Ignore error. If the file doesn't exist, that is fine.
 		_ = v.ReadInConfig()
 
-		return runDoctor(cfg, cmd.Flags(), args)
+		return runDoctor(cfg, cmd.Flags())
 	},
 }
 
-func runDoctor(cfg config.Config, flags *pflag.FlagSet, args []string) error {
+func runDoctor(cfg config.Config, flags *pflag.FlagSet) error {
 	usrCfg := cfg.UserViperConfig
 	if usrCfg.GetString("workspace") == "" {
 		return fmt.Errorf(msgRerunConfigure, BinaryName)
 	}
+
+	fixup, err := flags.GetBool("fixup")
+	if err != nil {
+		return err
+	}
+	if fixup {
+		if err := runFixup(cfg); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func runFixup(cfg config.Config) error {
+	usrCfg := cfg.UserViperConfig
 
 	ws, err := workspace.New(usrCfg.GetString("workspace"))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/doctor.go
```diff
@@ -0,0 +1,191 @@
...
+	if err := res.Body.Close(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func setupDoctorFlags(flags *pflag.FlagSet) {
+	flags.BoolP("fixup", "f", false, "run tasks")
+	// TODO: --dry-run for fixup
```
kytrinyx: I think we don't need both `--dry-run` and `--fixup`. Originally I was thinking that running `doctor` bare would fix things, and in that case we'd need `--dry-run` to just do reporting. With this approach we're reversing the logic, so that `doctor` bare does the reporting, which means that we need a flag (`--fixup`) to make the actual changes.

cmd/doctor.go
```diff
@@ -0,0 +1,191 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// TODO: extract this and cmd/download into download service?
+// copied from download#runDownload
+func downloadMetadata(exercise workspace.Exercise, cfg config.Config) error {
```
kytrinyx: unexported functions are available in the entire `cmd` package. We don't need to copy it, just extract it.

cmd/doctor.go
```diff
@@ -0,0 +1,191 @@
...
+func runDoctor(cfg config.Config, flags *pflag.FlagSet) error {
+	usrCfg := cfg.UserViperConfig
+	if usrCfg.GetString("workspace") == "" {
+		return fmt.Errorf(msgRerunConfigure, BinaryName)
+	}
+
+	fixup, err := flags.GetBool("fixup")
+	if err != nil {
+		return err
+	}
```
kytrinyx: In terms of design, what I think would help make it easy to do both modes (reporting and fixing), would be to have a type for each type of thing we're fixing. We can initialize the value, and have three different steps:

1. Investigate.
2. Report.
3. Fix.

The investigation would store whatever it needs to make it easy to both report and fix.

cmd/doctor.go
jdsutherland: Right. I wasn't suggesting that duplicating the same function. I was trying to point out that [runDownload](https://github.com/exercism/cli/blob/0cd96e0c880e5df9f9a484c85f75512e20c8f7a0/cmd/download.go#L51-L169) currently has duplicate behavior that we're going to want to use in `doctor` but we can't use the existing function. It seems that the logic for getting a `downloadPayload` should be extracted outside of the `cmd` package (given that multiple cmd's use it) with separate functions for writing metadata and solution files. I'm wondering if https://github.com/exercism/cli/issues/719 is related. I think this needs understood before proceeding with this PR.

kytrinyx: I'm not sure that it should live outside the cmd package, if only commands need it.

The code in api is very generic, whereas this is pretty specific.

I'd stick the extracted logic in `cmd/cmd.go` for now.

jdsutherland: Only the commands need it but it seems more like a service than an actual command. Everything in the `cmd` package is currently a command, so either the download payload stuff goes in `cmd/cmd` as a tacit home for shared command functionality or it could be put into a new 'helper/util' file. Maybe that's fine. That was my thinking anyway.

jdsutherland:
I don't anticipate completing this so I'm going to close it.

kytrinyx:
Ok, thank you @jdsutherland.

PR #750, opened 2018-10-19
Bump Travis versions

HaraldNordgren:


.travis.yml
```diff
@@ -5,7 +5,8 @@ sudo: false
 go:
   - "1.9"
```
nywilken: When updating supported versions on Travis we tend to drop support for the second oldest version (i.e 1.9). 
```suggestion
```

HaraldNordgren:
@nywilken I would rather test a wider array of Go versions to maintain backward compatibility as long as it is viable. Could you explain you reasoning a bit further? 🙂

kytrinyx:
@HaraldNordgren This only affects people contributing to the package, not people using the CLI. We find that overall Go developers upgrade within the year, and if they don't we feel it's ok if they get slower feedback by seeing failing tests in CI.

HaraldNordgren:
@nywilken @kytrinyx Understood. I made sure to change `CONTRIBUTING.md` as well since the one change doesn't make much sense without the other, and vice versa.

nywilken:
@HaraldNordgren sorry for the slow response on this. Thank you for making the updates.

Issue #680, opened 2018-07-25
Link to upgrade FAQs upon error 


nywilken:
Currently, if an error in encountered when trying to upgrade the Exercism CLI via the upgrade command an error is provided, but no additional guidance on how to resolve is provided to the user. 

This request is to update the CLI upgrade command to provide a link to common upgrade issues so that the user can try to resolve the issue on their own. 

The messaging should be something like the following. 

```
We were not able to write your file because we had an error:

{{raw error}}

Please check the FAQ on upgrading problems.

https://exercism.io/faq#cli-upgrade
```

nywilken:
https://github.com/exercism/website-copy/issues/261

kytrinyx:
@nywilken that's the correct repo!

nywilken:
Thanks @kytrinyx

nywilken:
Closed with PR #752

Smarticles101:
@exercism/cli doesn't look like this was ever closed...

nywilken:
Thanks @Smarticles101

PR #752, opened 2018-10-24
Add Descriptive Error Message to Upgrade Command

farisj:
Hi there! I'm an exercism user but first time contributor. Happy Hacktoberfest! 😄 🎃 

## What This Does

Solves https://github.com/exercism/cli/issues/680

This PR adds a wrapper around the `exercism upgrade` command which provides a more human-readable error message along with instructions to visit the FAQ page.

Currently, the FAQ page does not yet have upgrade debugging solutions, so perhaps it would be best to wait until https://github.com/exercism/website-copy/pull/417 is merged as well.

Thanks for taking the time to read this PR!

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -23,7 +23,14 @@ You can always delete this file.
 	`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		c := cli.New(Version)
-		return updateCLI(c)
+		err := updateCLI(c)
+		if err != nil {
+			err = fmt.Errorf(`We were not able to upgrade the cli because we encountered an error: %s
+Please check the FAQ for solutions to common upgrading issues.
+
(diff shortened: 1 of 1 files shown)


cmd/upgrade.go
```diff
@@ -23,7 +23,14 @@ You can always delete this file.
...
 		c := cli.New(Version)
-		return updateCLI(c)
+		err := updateCLI(c)
+		if err != nil {
+			err = fmt.Errorf(`We were not able to upgrade the cli because we encountered an error: %s
+Please check the FAQ for solutions to common upgrading issues.
+
+https://exercism.io/faqs`, err)
+		}
+		return err
```
nywilken: Seeing as we have the guard clause above lets return fmt.Errorf(...) within the condition and just `return nil` here.

PR #755, opened 2018-10-26
Refactor extract download logic

jdsutherland:
This change was motivated by #744, where the new `doctor` command needs to download missing metadata for exercises in the workspace. `cmd/download` contained the logic for this but to be usable it needed refactoring. Due to the overlapping need for downloading metadata, it makes sense to extract it outside of an individual command and split the download logic into separate functions.

Previously, cmd/download runDownload contained all the logic for getting
a downloadPayload, writing metadata, and writing solution files.

These changes separate and extract this logic outside
cmd/download for use by future commands.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,5 +1,22 @@
 package cmd
 
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	netURL "net/url"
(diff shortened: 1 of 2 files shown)


jdsutherland:
I think the diffs are a bit hard to read so I decided to reorganize the commits using a new branch. I'm closing this and making another PR using the new target.

Issue #718, opened 2018-08-25
Download all items for given track


Jarvvski:
Did some quick searching and I see that the ability to download all items for a given track isn't currently in the CLI.

I'd really find this item useful, as it's great to have access to items to work on during a long flight etc.

In the mean time (because I've been searching for it in the issues here) I made a tiny script to be able to download all items for a given track: https://github.com/Jarvvski/exercism-downloader

kytrinyx:
I think this would be a nice addition.

A few notes:

- It would have to be "all unlocked", not "all"
- In "independent mode", "all unlocked" would be equivalent to "all"

We would need some design/implementation on the backend API in https://github.com/exercism/website before we can start addressing this

raymondchen625:
Before this task is completed, we can use this shell command to pull the list of go:
`curl -LS https://exercism.io/tracks/go/exercises | grep "/tracks/go/exercises/" | awk '{print $3}' | cut -d/ -f5 | cut -d\" -f1`

tsoernes:
```
$ curl -LS https://exercism.io/tracks/go/exercises | grep "/tracks/go/exercises/" | awk '{print $3}' | cut -d/ -f5 | cut -d\" -f1 > exercies.txt
$ while read in; do exercism download --exercise="$in" --track=rust; done < exercises.txt
```

cocoonkid:
I also would massively appreciate a way to download full tracks.

agentofuser:
If you're doing multiple tracks in parallel this might be helpful:

```bash
# "exercism download all" for specific tracks

function eda() {
  for trackname in "$@"; do
    echo https://exercism.io/tracks/$trackname/exercises
    curl -LS https://exercism.io/tracks/$trackname/exercises | grep "/tracks/$trackname/exercises/" | awk '{print $3}' | cut -d/ -f5 | cut -d\" -f1 > exercises.txt
    while read in; do exercism download --exercise="$in" --track=$trackname; done < exercises.txt
    rm exercises.txt
  done
}

# example usage:
# ❯ eda typescript reasonml ocaml racket nim
```

Thank you @tsoernes for doing the hard part. This is just a for loop around his code.

farv:
I'm having issues using that @tsoernes code from a computer using WSL2 and redid some part of it:

[...]

Issue #697, opened 2018-08-05
Update metadata file when downloading a solution


kytrinyx:
If the metadata file has changed (e.g. people have switched from mentored to independent mode, or back) then we'll want to include this data in the solution file.

The download command should be a bit smarter with regards to the metadata file:

- write it if it's not already there
- write it if the uuid matches
- bail with an error if the uuid is different

Aliciawyse:
@kytrinyx Is this still worth solving? I'd be happy to take up!

PR #756, opened 2018-10-26
Refactor download cmd - compose & extract shared behavior for new doctor cmd

jdsutherland:
These changes are motivated by #744, in which the new `doctor` command needs to download missing metadata for exercises in the workspace. `cmd/download` contained the logic for this but to be usable it needed refactoring; due to the overlapping need for downloading metadata, it makes sense to extract it outside of an individual command and split the download logic into separate functions.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -95,8 +95,8 @@ type download struct {
 
 // newDownloadFromFlags initiates a download from flags.
 // This is the primary interaction for downloading from the Exercism API.
-func newDownloadFromFlags(usrCfg *viper.Viper, flags *pflag.FlagSet) (*download, error) {
-	downloadParams, err := newDownloadParamsFromFlags(usrCfg, flags)
+func newDownloadFromFlags(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
+	downloadParams, err := newDownloadParamsFromFlags(flags, usrCfg)
 	if err != nil {
 		return nil, err
 	}
@@ -106,8 +106,8 @@ func newDownloadFromFlags(usrCfg *viper.Viper, flags *pflag.FlagSet) (*download,
 // newDownloadFromExercise initiates a download from an exercise.
 // This is used to get metadata and isn't the primary interaction for downloading.
 // Only allows writing metadata, not exercise files.
-func newDownloadFromExercise(usrCfg *viper.Viper, exercise ws.Exercise) (*download, error) {
-	downloadParams, err := newDownloadParamsFromExercise(usrCfg, exercise)
+func newDownloadFromExercise(exercise ws.Exercise, usrCfg *viper.Viper) (*download, error) {
+	downloadParams, err := newDownloadParamsFromExercise(exercise, usrCfg)
 	if err != nil {
 		return nil, err
 	}
@@ -391,7 +391,7 @@ type downloadParams struct {
 }
 
 // newDownloadParamsFromExercise creates a new downloadParams given an exercise.
-func newDownloadParamsFromExercise(usrCfg *viper.Viper, exercise ws.Exercise) (*downloadParams, error) {
+func newDownloadParamsFromExercise(exercise ws.Exercise, usrCfg *viper.Viper) (*downloadParams, error) {
 	d := &downloadParams{
 		slug:             exercise.Slug,
 		track:            exercise.Track,
@@ -401,7 +401,7 @@ func newDownloadParamsFromExercise(usrCfg *viper.Viper, exercise ws.Exercise) (*
 }
 
 // newDownloadParamsFromFlags creates a new downloadParams given flags.
-func newDownloadParamsFromFlags(usrCfg *viper.Viper, flags *pflag.FlagSet) (*downloadParams, error) {
+func newDownloadParamsFromFlags(flags *pflag.FlagSet, usrCfg *viper.Viper) (*downloadParams, error) {
 	d := &downloadParams{downloadableFrom: downloadableFromFlags{}}
 	var err error
 	d.uuid, err = flags.GetString("uuid")
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -180,11 +180,11 @@ func (d *download) setPayload() error {
 		return err
 	}
 
-	req, err := client.NewRequest("GET", d.requestURL(), nil)
+	req, err := client.NewRequest("GET", d.payloadURL(), nil)
 	if err != nil {
 		return err
 	}
-	d.buildQuery(req.URL)
+	d.buildPayloadQueryParams(req.URL)
 
 	res, err := client.Do(req)
 	if err != nil {
@@ -217,7 +217,7 @@ func (d *download) setPayload() error {
 	return nil
 }
 
-func (d download) requestURL() string {
+func (d download) payloadURL() string {
 	id := "latest"
 	if d.params.uuid != "" {
 		id = d.params.uuid
@@ -225,7 +225,8 @@ func (d download) requestURL() string {
 	return fmt.Sprintf("%s/solutions/%s", d.params.apibaseurl, id)
 }
 
-func (d download) buildQuery(url *netURL.URL) {
+// buildPayloadQueryParams adds optional query parameters to the URL.
+func (d download) buildPayloadQueryParams(url *netURL.URL) {
 	query := url.Query()
 	if d.params.slug != "" {
 		query.Add("exercise_id", d.params.slug)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -321,7 +321,7 @@ func (w fileDownloadWriter) writeMetadata() error {
 // An HTTP request is made using each filename and failed responses are swallowed.
 // All successful file responses are written except when 0 Content-Length.
 func (w fileDownloadWriter) writeSolutionFiles() error {
-	if err := w.download.params.ensureWritable(); err != nil {
+	if err := w.download.params.ensureExerciseFilesWritable(); err != nil {
 		return err
 	}
 	for _, filename := range w.download.payload.Solution.Files {
@@ -443,9 +443,10 @@ func (d *downloadParams) validate() error {
 	return nil
 }
 
-func (d downloadParams) ensureWritable() error {
+// ensureExerciseFilesWritable checks permission for writing exercise files.
+func (d downloadParams) ensureExerciseFilesWritable() error {
 	if !d.downloadableFrom.writeExerciseFilesPermitted() {
-		return errors.New("writing exercise files not permitted")
+		return errors.New("writing exercise files not permitted when downloading from this type")
 	}
 	return nil
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -122,11 +122,9 @@ func newDownload(params *downloadParams, writer downloadWriter) (*download, erro
 	}
 
 	d := &download{params: params}
-
 	d.payload, err = d.requestPayload()
 	if err != nil {
 		return nil, err
-
 	}
 
 	writer.init(d)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -373,7 +373,7 @@ func newDownloadParamsFromExercise(exercise ws.Exercise, usrCfg *viper.Viper) (*
 		track:            exercise.Track,
 		downloadableFrom: downloadableFromExercise{},
 	}
-	return d.newDownloadParams(usrCfg)
+	return d.build(usrCfg)
 }
 
 // newDownloadParamsFromFlags creates a new downloadParams given flags.
@@ -396,11 +396,11 @@ func newDownloadParamsFromFlags(flags *pflag.FlagSet, usrCfg *viper.Viper) (*dow
 	if err != nil {
 		return nil, err
 	}
-	return d.newDownloadParams(usrCfg)
+	return d.build(usrCfg)
 }
 
-// newDownloadParams contains the common creation logic for creating downloadParams.
-func (d *downloadParams) newDownloadParams(usrCfg *viper.Viper) (*downloadParams, error) {
+// build contains the common creation logic for creating downloadParams.
+func (d *downloadParams) build(usrCfg *viper.Viper) (*downloadParams, error) {
 	d.token = usrCfg.GetString("token")
 	d.apibaseurl = usrCfg.GetString("apibaseurl")
 	d.workspace = usrCfg.GetString("workspace")
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -321,32 +321,37 @@ func (d fileDownloadWriter) writeSolutionFiles() error {
 		return err
 	}
 	for _, filename := range d.Solution.Files {
-		res, err := d.requestSolutionFile(filename)
-		if err != nil {
-			return err
-		}
-		if res == nil {
-			continue
-		}
-		defer res.Body.Close()
+		d.writeSolutionFile(filename)
+	}
+	return nil
+}
 
-		// TODO: if there's a collision, interactively resolve (show diff, ask if overwrite).
-		// TODO: handle --force flag to overwrite without asking.
+func (d fileDownloadWriter) writeSolutionFile(filename string) error {
+	res, err := d.requester.requestSolutionFile(filename)
+	if err != nil {
+		return err
+	}
+	if res == nil {
+		return nil
+	}
+	defer res.Body.Close()
 
-		destination := filepath.Join(
-			d.destination(),
-			sanitizeLegacyFilepath(filename, d.exercise().Slug))
-		if err = os.MkdirAll(filepath.Dir(destination), os.FileMode(0755)); err != nil {
-			return err
-		}
-		f, err := os.Create(destination)
-		if err != nil {
-			return err
-		}
-		defer f.Close()
-		if _, err := io.Copy(f, res.Body); err != nil {
-			return err
-		}
+	// TODO: if there's a collision, interactively resolve (show diff, ask if overwrite).
+	// TODO: handle --force flag to overwrite without asking.
+
+	destination := filepath.Join(
+		d.destination(),
+		sanitizeLegacyFilepath(filename, d.exercise().Slug))
+	if err = os.MkdirAll(filepath.Dir(destination), os.FileMode(0755)); err != nil {
+		return err
+	}
+	f, err := os.Create(destination)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	if _, err := io.Copy(f, res.Body); err != nil {
+		return err
 	}
 	return nil
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -103,18 +103,21 @@ type download struct {
 	*downloadWriter
 }
 
-// newDownloadFromExercise is a convenience wrapper for creating a new download.
-func newDownloadFromExercise(usrCfg *viper.Viper, exercise ws.Exercise) (*download, error) {
-	downloadParams, err := newDownloadParamsFromExercise(usrCfg, exercise)
+// newDownloadFromFlags initiates a download from flags.
+// This is the primary interaction for downloading from the Exercism API.
+func newDownloadFromFlags(usrCfg *viper.Viper, flags *pflag.FlagSet) (*download, error) {
+	downloadParams, err := newDownloadParamsFromFlags(usrCfg, flags)
 	if err != nil {
 		return nil, err
 	}
 	return newDownload(downloadParams)
 }
 
-// newDownloadFromFlags is a convenience wrapper for creating a new download.
-func newDownloadFromFlags(usrCfg *viper.Viper, flags *pflag.FlagSet) (*download, error) {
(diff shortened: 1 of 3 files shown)


jdsutherland:
I extracted the common download stuff to `cmd/cmd` initially as an intended shared home for commands. It seems odd to leave `cmd/cmd` containing a bunch of download logic though. What do you think about moving it to a new file that indicates it contains shared download functionality (perhaps `cmd/downloadHelper`)?

jdsutherland:
Questionable points:
* The filename `download_helper` doesn't feel quite right.
* The name of the wrapping struct `downloadPayloadContext`. I came up with 'context' for lack of better term.
* Current test coverage is from `download_test` - depending on where this ends up living, should we consider adding/rearranging coverage?

cmd/download_helper.go
```diff
@@ -0,0 +1,244 @@
...
+		q := req.URL.Query()
+		q.Add("exercise_id", params.slug)
+		if params.track != "" {
+			q.Add("track_id", params.track)
+		}
+		if params.team != "" {
+			q.Add("team_id", params.team)
+		}
+		req.URL.RawQuery = q.Encode()
+	}
```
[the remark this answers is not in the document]
jdsutherland: Yeah, that's a good suggestion. I implemented that here https://github.com/exercism/cli/pull/756/commits/269c35805bee5499e4d88d53a516a6daaba94e14. 

> The request params could post back a map or some other type that we could range over, which if there is nothing to do could be just an empty one.

I'm not sure if I'm missing something. Are these changes what you had in mind?

cmd/download_helper.go
```diff
@@ -0,0 +1,244 @@
...
+
+type downloadParams struct {
+	usrCfg *viper.Viper
+	uuid   string
+	slug   string
+	track  string
+	team   string
+}
+
+type downloadPayloadContext struct {
```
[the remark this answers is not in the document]
jdsutherland: I did this here https://github.com/exercism/cli/pull/756/commits/02cbe9bf73f417ad7f53f69c4e944c68b3a3e659. Unless I'm missing something, I don't think there is an additional benefit of merging `downloadParams` into `downloadContext`. The decision rests on if you think that having a separate parameter object is worthwhile. I think it might be worth having for separation of concerns. What do you think?

cmd/download_helper.go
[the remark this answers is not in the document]
kytrinyx: I think this works as an implementation (I was thinking something slightly different, but my idea wasn't any better :-)).

cmd/download_helper.go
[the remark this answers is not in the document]
kytrinyx: Yepp, I like it!

cmd/cmd.go
```diff
@@ -33,3 +51,221 @@ const msgMissingMetadata = `
...
+type downloadContext struct {
+	usrCfg  *viper.Viper
+	uuid    string
+	slug    string
+	track   string
+	team    string
+	payload *downloadPayload
+}
+
+func newDownload(ctx *downloadContext) error {
```
kytrinyx: I'm not sure `newDownload` is the right name here. Generally if I see a function `newDownload` I would expect it to return a `Download` value or a pointer to a download value. In this case it _performs_ a download, right?

cmd/cmd.go
```diff
@@ -33,3 +51,221 @@ const msgMissingMetadata = `
...
+		return err
+	}
+
+	req, err := client.NewRequest("GET", ctx.requestURL(), nil)
+	if err != nil {
+		return err
+	}
+
+	query := req.URL.Query()
+	ctx.buildQuery(query)
```
kytrinyx: Should the `buildQuery` initialize the query itself as well? Something like:

```
query := ctx.buildQuery()
```

cmd/cmd.go
```diff
@@ -33,3 +51,221 @@ const msgMissingMetadata = `
...
+		case "track_ambiguous":
+			return fmt.Errorf("%s: %s", ctx.payload.Error.Message, strings.Join(ctx.payload.Error.PossibleTrackIDs, ", "))
+		default:
+			return errors.New(ctx.payload.Error.Message)
+		}
+	}
+	return nil
+}
+
+func (d *downloadContext) writeMetadata() error {
```
kytrinyx: Sometimes we've called this `d` other times we've called it `ctx`. I think it would be worth being consistent.

cmd/cmd.go
```diff
@@ -33,3 +51,221 @@ const msgMissingMetadata = `
...
+		}
+		defer f.Close()
+		if _, err := io.Copy(f, res.Body); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *downloadContext) getExercise() workspace.Exercise {
```
kytrinyx: Style-wise in the CLI we typically don't use `get` prefixes for getters, we'll usually name the method after what it returns (so `exercise()`).

cmd/cmd.go
```diff
@@ -33,3 +51,221 @@ const msgMissingMetadata = `
...
+		root = filepath.Join(root, "users", d.payload.Solution.User.Handle)
+	}
+	return workspace.Exercise{
+		Root:  root,
+		Track: d.payload.Solution.Exercise.Track.ID,
+		Slug:  d.payload.Solution.Exercise.ID,
+	}
+}
+
+func (d *downloadContext) getMetadata() workspace.ExerciseMetadata {
```
kytrinyx: Ditto here, this should probably be `metadata()`.

cmd/cmd.go
jdsutherland: I don't see a better way of doing it. Initializing requires a reference to the `*http.Request`, right? I don't see how we can avoid having to pass that in.

jdsutherland:
Thanks for the feedback.

Do we want to make any changes to the tests? I think we might be forced into making a move since there will be shared behavior behavior on `download` and `doctor` (#697 - unless we think it's not worth explicitly testing).

I think it's also worth noting that there are a few other methods that live in `download_test` and `submit_test` that `doctor_test` probably wants to share: `fakeDownloadServer` and `fakeSubmitServer`. Do we want to extract those to `cmd_test`?

kytrinyx:
> should they be extracted to `cmd_test`

Yeah, I think that would be a good move. It means that if we don't find the helper method in the test file we're in, we know to look for it there.

jdsutherland:
I've iterated on this quite a bit and I think it's (finally) reached a steady state. In the process of working on this, I started refactoring [submit](https://github.com/exercism/cli/pull/775) as well and somewhat of a common design emerged, so it might be worth considering both. This PR sort of grew into a monster so apologies for any hardships I may have caused in the process reviewing this.

jdsutherland:
I've iterated on this quite a bit and I think it's (finally) reached a steady state. In the process of working on this, I started refactoring [submit](https://github.com/exercism/cli/pull/775) as well and somewhat of a common design emerged, so it might be worth taking both into consideration. This PR sort of grew into a monster so apologies for any hardships I may cause as a result of reviewing this.

jdsutherland:
I've iterated on this quite a bit and I think it's (finally) reached a steady state. In the process of working on this, I started refactoring [submit](https://github.com/exercism/cli/pull/775) as well and somewhat of a common design emerged, so it might be worth taking both into consideration. This PR sort of grew into a monster so apologies for any hardships I may cause as a result of reviewing this.

jdsutherland:
I've iterated on this quite a bit and I think it's (finally) reached a steady state (I said this several times before making further changes). In the process of working on this, I started refactoring [submit](https://github.com/exercism/cli/pull/775) as well and somewhat of a common design emerged, so it might be worth taking both into consideration. This PR sort of grew into a monster so apologies for any hardships I may cause as a result of reviewing this.

Previously, there was reservation around the idea of extracting this into a new service package. I understand if we don't want to go there, but I couldn't help myself - I was curious to see how it would be like (regardless of if it gets merged) so I made a new branch which can be seen here https://github.com/jdsutherland/cli/commit/83aa979d0d030ba3b21bde158b081892abb126d2.

cmd/cmd.go
```diff
@@ -70,3 +80,323 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+	return newDownload(downloadParams)
+}
+
+// newDownload initiates a download by requesting a downloadPayload from the Exercism API.
+func newDownload(params *downloadParams) (*download, error) {
+	if err := params.validate(); err != nil {
+		return nil, err
+	}
+	d := &download{downloadParams: params}
+	d.downloadWriter = &downloadWriter{download: d}
```
jdsutherland: Having circular dependencies seems weird (`Download` and `downloadWriter` both hold direct references to each other) but maybe this is ok? It reads better to me to have the separate types but maybe doing so is too hacky. The thinking is that `Download` mainly contains data and queries whereas `downloadWriter` distinguishes commands (the write methods). This might offer flexibility as it could be easily modified to be an interface if needed (though maybe I'm missing YAGNI).

jdsutherland:
As a learning exercise, I tried extracting this to a service package along with adding isolated unit tests. This can be seen here https://github.com/jdsutherland/cli/commit/4e904d3e9ee9945c3696282823197aed952676c6. I see the tradeoffs being:

pro
* Info hiding - exported methods enforce API
* Clean architecture - without extracting, cmd/cmd becomes mostly related to Download when we might expect it to be more of a generally shared home for commands
* Test isolation - future consumer's tests can focus on higher level features
* Reduced coupling - avoids potential coupling between cmd/download's tests and other consumers

con
* tests increase codebase by 500 LOC (though perhaps cmd/download_test may be trimmed, reducing this)
* potential duplication/sync issues between service/download_test and cmd/download_test might be costly to resolve
* increased code review burden
* if it seems exceedingly unlikely that there will be new service/download consumers in the future this becomes hard to justify

I suspect there might not be much reason to think that there will be additional consumers and the cost seems fairly high so I'm inclined to think it's not worth it. I thought it might be worth mentioning anyway.

jdsutherland:
Based on https://github.com/exercism/cli/issues/639#issuecomment-410533610, is it correct that `doctor` shouldn't ever be writing exercise files? https://github.com/exercism/cli/pull/756/commits/b5527f52f96d85ca9de052688e44d81f1a253d77 throws an error when attempting to call write from a download init via an exercise (i.e. via doctor). Is this desired behavior?

cmd/cmd.go
```diff
@@ -70,3 +80,327 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+
+func (d *download) requestURL() string {
+	id := "latest"
+	if d.uuid != "" {
+		id = d.uuid
+	}
+	return fmt.Sprintf("%s/solutions/%s", d.usrCfg.GetString("apibaseurl"), id)
+}
+
+func (d *download) buildQuery(url *netURL.URL) {
```
jdsutherland: AFAICT, this is the only place where `track` or `team` flags are used. It seems that using `--track` or `--team` with `--uuid` renders track or team impotent. If this is indeed an invalid combination of arguments, it seems worth making explicit. An added benefit is clarification of what team/track are used for in `downloadParams`. Rather than commit this now (it might change based on the rest of the review direction), it will be noted here as a potential TODO. Something like: 

```go
func (d *downloadParams) givenUUIDWithTrackOrTeam() error {
	if d.uuid!= "" && (d.team != "" || d.track != "") {
		return errors.New("--team or --track needs --exercise (not --uuid)")
	}
	return nil
}
```

cmd/cmd.go
```diff
@@ -70,3 +80,327 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+
+// validate validates the presence of required downloadParams.
+func (d *downloadParams) validate() error {
+	if d.slug != "" && d.uuid != "" || d.uuid == d.slug {
+		if d.fromFlags {
+			return errors.New("need an --exercise name or a solution --uuid")
+		}
+		return errors.New("need a 'slug' or a 'uuid'")
+	}
+	if err := validateUserConfig(d.usrCfg); err != nil {
```
jdsutherland: `submit` and `download` require the same config values. I'm wondering if it'd be better to use config.Config (instead of viper.Viper) and move the validation directly to the config object. This strikes me as a case of Tell, Don't Ask.

cmd/cmd.go
```diff
@@ -70,3 +80,338 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+// downloadParams is required to create a download.
+type downloadParams struct {
+	usrCfg *viper.Viper
+	uuid   string
+	slug   string
+	track  string
+	team   string
+
+	// config
+	token, apibaseurl, workspace string
```
jdsutherland: I'm not sure it's worth putting the config values as fields here. The reasoning is that rather than have repeated lookups for config values, e.g. `d.usrCfg.GetString("token")`, we can store them in `downloadParams` so that `Download` has the values it needs directly without depending on the config structure. Encapsulating these values in fields has the benefit of preventing errors that the compiler can't catch, misspelling a key, say, `d.usrCfg.GetString("tokn")`.

If this seems worthwhile, `downloadParams` shouldn't need to store *viper.Viper as a field. Currently it remains because it's needed in `validate()` but that can change depending on if we want to move validation to the config object as mentioned in a further comment below. If we don't want to store it on the config object, we can add the validation logic directly to `downloadParams`.

cmd/cmd.go
```diff
@@ -70,3 +80,394 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+	metadata := d.metadata()
+	return metadata.Write(d.destination())
+}
+
+// writeSolutionFiles attempts to write each exercise file that is part of the downloaded Solution.
+// An HTTP request is made using each filename and failed responses are swallowed.
+// All successful file responses are written except when 0 Content-Length.
+func (d downloadWriter) writeSolutionFiles() error {
+	if d.params.fromExercise {
+		return errors.New("existing exercise files should not be overwritten")
```
jdsutherland: This probably isn't quite what this error should say. But if `doctor` shouldn't be allowed to write exercise files, it seems like worth throwing an error here.

cmd/cmd.go
```diff
@@ -70,3 +80,503 @@ func validateUserConfig(cfg *viper.Viper) error {
...
+
+// errGivenTrackOrTeamMissingSlug is a template pattern default.
+func (d downloadParams) errGivenTrackOrTeamMissingSlug() error {
+	return errors.New("track or team requires slug (not uuid)")
+}
+
+// downloadableFrom is the interface to use the template pattern when creating downloadParams from different types.
+// Clients can embed downloadParams to delegate to the default implementation.
+// This allows fine-grained specializations without having to define the entire interface.
+type downloadableFrom interface {
```
jdsutherland: I'm not sure how likely it is that there will be new data sources initiating a download. If it seems very unlikely, this is probably overkill. At present, it seems clear that we want slightly different behavior when downloading from Flags vs an Exercise.

If we imagine a future change allowing a download from ExerciseMetadata, it seems that the behavior and error messages might be the same as when downloading from an Exercise. If this sounds like a reasonable assumption then this interface might be simplified into a bool to signal creation from flags (as this is the only varying behavior). 

If we want to leave open the ability to add custom behavior when initiating a download from different types, then having an explicit duck type to easily hook into seems useful though.

jdsutherland:
This has been sitting for awhile and I don't plan on making further changes. This message will serve as a status of present concerns:

I don't think `cmd/cmd` is a good home for the download related stuff. Currently, download specific stuff takes up ~80% of the file where the remainder seems like appropriate shared cmd helpers. It seems that download is clearly a separate concept when compared to the rest of the file and therefore belongs elsewhere.

If we move it to a new file in the `cmd` package, it's hard to think of a good name (we already have `cmd/download`). It seems appropriate to pull this out into a separate package. This seems preferable to an awkwardly named file in the `cmd` package when this isn't really a command but rather a service consumed by commands.

I previously expressed some concerns about testing but I don't think it's worth worrying about.

kytrinyx:
Thanks @jdsutherland. I'm going to close this, but it has been a really useful discussion. The download command is a bit of a monster and I honestly don't quite know where to go with it.

jdsutherland:
I'm a bit surprised by the jump to closing this without discussion and I don't really have a sense as to why. In the interest of feedback, it'd be useful to understand the thought process here. Was this just so far off the mark that it wasn't worth iterating on?

kytrinyx:
@jdsutherland You said you weren't going to work on it anymore. Did I misunderstand?

jdsutherland:
@kytrinyx I see how what I said could be misunderstood.

The previous message was just an attempt to give an up to date status - I had posted several comments in the past that I now deem outdated based on iterating on this so I was trying to say: this seems stable and I don't plan on changing this until I get further feedback.

Sorry for the confusion.

kytrinyx:
Oh my goodness I'm so sorry. I totally see how that can be read both ways, now that I know what you intended!

So. Since I thought you had given up on this one, I basically took the direction that I liked about it, and extracted a bunch of little bits and pieces into separate PRs (I used the `co-authored-by` tag to give you credit) and merged them into master. Then I opened this as a suggested next step: https://github.com/exercism/cli/pull/820

The main idea is based on what you did here, but it has fewer abstractions.

Would you pull that down and see if it would solve the "extract so we can re-use" problem for you?
Would you take a look at it in terms of readability and see if you think we should take it in a different direction?

jdsutherland:
No worries.

I'm not sure if it's best to put things here or in #820, but I'll start here because this PR is relevant in reference to what I'm saying (I'll move it if we decide otherwise).

I think this PR (#756) may be overly abstract or complex than is needed but I think we're going to want more pulled out than is currently in #820.

I envision the part of doctor that migrates legacy metadata will look like this:
```go
exercises, err := ws.PotentialExercises()
if err != nil {
	return err
}

for _, exercise := range exercises {
	if _, err := exercise.MigrateLegacyMetadataFile(); err != nil {
		return err
	}
	if ok, _ := exercise.HasMetadata(); !ok {
		download, err := newDownloadFromExercise(exercise, cfg.UserViperConfig)
		if err != nil {
			return err
		}

		writer, err := newDownloadWriter(download)
		if err != nil {
			return err
		}

		if err := writer.writeMetadata(); err != nil {
			return err
		}
	}
}
```

If you think about implementing this from #820, there's a lot of potential duplication among shared behavior. `doctor` will have much more in common with `download` than is currently pulled out. For example:
* validation
* handling the download response
* writing metadata (including destination)

It doesn't seem that really any of the behavior is going to differ between them so it seems appropriate to abstract it out and encapsulate it.

I think there's good potential future reasons for pulling it out as well (https://github.com/exercism/cli/issues/718 is one example).

kytrinyx:
Yepp, that makes perfect sense to me.

I think I'd prefer to do the extractions in multiple PRs, though, rather than in one big one, since it is so much harder to review a big PR (which means I end up putting it off much longer).

Shall we see if we can get #820 into a reasonable state for what it's currently doing, and then tackle the next extractions in a series of smaller PRs, discussing how far to take things each time?

jdsutherland:
I'm not sure how I feel about moving forward with #820. I've put a good amount of work into this PR and it seems to faciliate what's needed for the doctor command. I'm happy to adjust things based on review but throwing most of it out without feedback and starting from another point doesn't make much sense to me. If there's a good reason for doing so, I'm interested to understand why but I currently don't see it.

kytrinyx:
Ok, sure. Let's move forward with this one. I'll close #820.

However, would you be willing to break it up into smaller pieces, one at a time? It's really challenging to review a change that is this big. If you have an idea of where you're going, could we take it in smaller steps?

jdsutherland:
Sure. Given how many garbage commits this has (and it has conflicts that need resolved anyway), it might make more sense to start from master.

Given that I can't unsee this, it's probably going to be hard to me to gauge what comprehensible steps are going to look like. Should we proceed by me guessing and go from there or should we try to clear certain things up first?

For example, is it possible to look at it now and say that there are concerns about the main abstractions?

kytrinyx:
> Since that I can't unsee this, it's probably going to be hard to me to gauge what comprehensible steps look like.

Yeah, I know the feeling :)

> it might make more sense to start from master.

I think that makes sense.

My main concern is that there are so many abstractions I think that having fewer abstractions, at least to start, would make more sense. Let me go through the PR and see if I can suggest some first steps.

kytrinyx:
Ok, what do you think about starting off with a very simple `download` type similar to what you have, and creating it from flags? I'd like to avoid creating a separate params type and validator type, and have that behavior on the download type.

In this first step, I'd like to avoid thinking about what the `doctor` command needs just yet, and just work towards a minimal change that is better than what we currently have.

kytrinyx:
Looking at master and at your code again, I think that we'll end up pulling all the flag stuff out of `addQueryToDownloadRequest` and `downloadIdentifier` straight into your abstraction, and do what you did with the params and validation: pull stuff out of the flags so you can deal with the errors up front, and then validate the logic without having to deal with the errors in the same flow.

jdsutherland:
https://github.com/exercism/cli/pull/821 is an attempt at following what you've said. My concern is that this can seem bad in the context of intermediate steps but in the finished state, it makes more sense.

It's possible that the types I have aren't necessary but I think that download may end up being large and having multiple responsibilities whereas I was viewing it as a simple value object that just holds data.

jdsutherland:
This is going in a different direction so it seems fit to close this.

Issue #760, opened 2018-11-05
Fail the travis build if `fmt` has not been run


kytrinyx:
We have seen a few pull requests recently where we've manually asked authors to run `fmt`. Since we always want this to happen, let's make the Travis build fail with a helpful message if `fmt` would make changes.

PR #761, opened 2018-11-08
Require gofmt on Travis build

jdsutherland:
Closes https://github.com/exercism/cli/issues/760

### DIFF
--- a/.travis.gofmt.sh
+++ b/.travis.gofmt.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+cd "$(dirname $0)"
+if [ -n "$(go fmt ./...)" ]; then
+    echo "Go code is not formatted, run 'go fmt github.com/exercism/cli/...'" >&2
+    exit 1
+fi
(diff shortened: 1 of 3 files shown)


.travis.gofmt.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cd "$(dirname $0)"
+if [ -n "$(go fmt ./...)" ]; then
```
jdsutherland: I yanked [google/gopacket](https://github.com/google/gopacket/blob/f9d6df4cf07e221be4221834a9f903ee7688ba7a/.travis.gofmt.sh)'s approach.

To clarify how this works: `if [ -n "$(go fmt ./...)" ]` writes the fixes and outputs the files (making this conditional true thus failing Travis). So the committer likely doesn't need to manually run gofmt again - Travis failure output provides an explanation and the comitter can simply verify/commit the formatting fixes.

kytrinyx: Lovely!

Issue #758, opened 2018-10-30
Add completion for fish shell


hellow554:
I'd like to see a completion for the [fish shell](https://fishshell.com/)

kytrinyx:
Agreed, that would be lovely.
See https://github.com/exercism/cli/issues/666 for reference to the Cobra documentation for completion scripts.

olingern:
@kytrinyx I think this can be closed a la #763

kytrinyx:
Agreed! Thanks so much :-)

PR #763, opened 2018-11-14
Fish tab completions

John-Goff:
https://github.com/exercism/cli/issues/758

I tried to copy the zsh and bash completions as closely as I could, but I'd appreciate if someone else could review these. Preferably someone who's more familiar with how the bash and zsh completions should preform than I am!


shell/exercism.fish
```diff
@@ -0,0 +1,49 @@
...
+complete -f -c exercism -n "__fish_use_subcommand" -a "configure" -d "Writes config values to a JSON file."
+complete -f -c exercism -n "__fish_seen_subcommand_from configure" -s t -l token -d "Set token"
+complete -f -c exercism -n "__fish_seen_subcommand_from configure" -s w -l workspace -d "Set workspace"
+complete -f -c exercism -n "__fish_seen_subcommand_from configure" -s a -l api -d "set API base url"
+complete -f -c exercism -n "__fish_seen_subcommand_from configure" -s s -l show -d "show settings"
+
+# Download
+complete -f -c exercism -n "__fish_use_subcommand" -a "download" -d "Downloads and saves a specified submission into the local system"
+complete -f -c exercism -n "__fish_seen_subcommand_from download" -s e -l exercise -d "the exercise slug"
+complete -f -c exercism -n "__fish_seen_subcommand_from download" -s h -l help -d "help for download"
```
nywilken: We can probably do without tab completion for the help flag, as it will repeat the information you have already added for each option. Thoughts?

shell/exercism.fish
```diff
@@ -0,0 +1,49 @@
...
+complete -f -c exercism -n "__fish_seen_subcommand_from download" -s T -l team -d "the team slug"
+complete -f -c exercism -n "__fish_seen_subcommand_from download" -s t -l track -d "the track ID"
+complete -f -c exercism -n "__fish_seen_subcommand_from download" -s u -l uuid -d "the solution UUID"
+
+# Help
+complete -f -c exercism -n "__fish_use_subcommand" -a "help" -d "Shows a list of commands or help for one command"
+complete -f -c exercism -n "__fish_seen_subcommand_from help" -a "configure download help open submit troubleshoot upgrade version workspace"
```
nywilken: ditto

shell/README.md
```diff
@@ -31,3 +31,10 @@ the following snippet
 #### Oh my Zsh
 
 If you are using the popular [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) framework to manage your zsh plugins, move the file `exercism_completion.zsh` into `~/.oh-my-zsh/custom`.
+
+### Fish
+
+Completions must go in the user defined `$fish_complete_path`. By default, this is `~/.config/fish/completions`
+
```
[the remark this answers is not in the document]
John-Goff: We could do that, but the final file has to have the same name as the executable. I figured it would be easier to have it named that way from the start, rather than requiring users to rename it themselves. I foresee at least a few people dragging and dropping rather than using the command line.

shell/exercism.fish
John-Goff: these completions fire when a user types `exercism <command> -<tab>` (note the dash). It's only there for completeness, I created these mostly by copying the output of `exercism help <command>`.

John-Goff:
Has anyone been able to test these? If I need to make any changes, please let me know.

nywilken:
> Has anyone been able to test these? If I need to make any changes, please let me know.

@John-Goff I had an opportunity to test the first time around, but didn't test entirely. I will take another look. Thank you for the detailed responses.

shell/exercism.fish
```diff
@@ -0,0 +1,49 @@
...
+complete -f -c exercism -n "__fish_seen_subcommand_from version" -s l -l latest -d "check latest available version"
+complete -f -c exercism -n "__fish_seen_subcommand_from version" -s h -l help -d "help for version"
+
+# Workspace
+complete -f -c exercism -n "__fish_use_subcommand" -a "workspace" -d "Outputs the root directory for Exercism exercises."
+complete -f -c exercism -n "__fish_seen_subcommand_from workspace" -s h -l help -d "help for workspace"
+
+# Options
+complete -f -c exercism -s h -l help -d "show help"
+complete -f -c exercism -l timeout -a "(seq 0 100 100000)" -d "override default HTTP timeout"
```
nywilken: The tab completion for possible timeout values is cool. But I found the possible options a bit overwhelming - it filled my terminal. Is this a common practice you’ve seen in fish? If so, I recommend shortening the possible value range. 

Removing the range entirely has a funky behavior where if I tab after the timeout flag it tab completes to some random file. Possibly why you added the sequence?

John-Goff:
@nywilken No problem and thanks for your review!

shell/exercism.fish
John-Goff: I can definitely shorten the range. I agree there's potentially too many options to tab through. Your analysis is spot on, I added it to prevent the files autocompleting there. Files autocompleting is fish's default behaviour, and most programs do this if there's no available options to add. I thought some range of numbers to be better than nothing, if for no other reason than to show that it's possible. I'll shorten the range.

NobbZ: Is it possible in `fish` to document a value as it is in `zsh`? If so wouldn't it make sense to only have most commonly used timouts of 10, 30, 60 seconds, 5 and 10 minutes? 

In `zsh` such a completion list would roughly look like:

```
10 -- 10 seconds
30 -- 30 seconds
60 -- 1 minute
300 -- 5 minutes
600 -- 10 minutes
```

I consider such a completion much better than a plain list of values that no-one knows what they do without canceling the completion and looking them up in the documentation first.

NobbZ:
I've left another comment on the timeout range inline.

shell/exercism.fish
John-Goff: This is indeed possible and I have updated the completions to include this. Thanks for the suggestion, I was really unsure what to put for timeout values.

John-Goff:
Apologies I didn't see this update before now. Thank you for your review, I have updated the completions to incorporate your suggestions.

shell/exercism.fish
John-Goff: In fish this looks like
```
➤ exercism --timeout=
…-timeout=10  (10 seconds)  …-timeout=30  (30 seconds)  …-timeout=60  (1 minute)  …-timeout=300  (5 minutes)  …-timeout=600  (10 minutes)
```

kytrinyx:
Thank you @John-Goff! I'm going to pull this in, and if we find any other improvements we can do them in a second round ✨

hellow554:
@John-Goff  please use one of the [github keywords](https://help.github.com/articles/closing-issues-using-keywords/) to automatically close a referenced issue when the PR has been merged so it's easier :) thanks for the work!

PR #769, opened 2018-11-20
Remove dead code from configure_test

jdsutherland:
I was looking at `configure_test` and noticed that several tests have similar setup for spying on stdout/err but in few places it's not actually being cashed out. These changes attempt to remove the dead code.


kytrinyx:
Nice! Thanks so much for catching this. We've done a bunch of refactoring in the past few months, and stuff like this can be easy to miss in the heat of the moment :)

PR #770, opened 2018-11-20
Submit prints out api error messages

### LINKED ISSUE #768: submit: If I just submitted an identical iteration, tell me that
Let's say I've forgotten whether I've submitted my most recent changes or not. Maybe I stepped outside for a break, and forgot what I did right before stepping out. I'd like to know.

The API server currently tells this. If you changed something, it gives 201 Created and body `{}`. If you didn't change anything, it gives 400 Bad Request and body `{"error":{"type":"duplicate_iteration","message":"No files you submitted have changed since your last iteration"}}` . Making use of that information could help.

Smarticles101:
Closes #768 

I mostly copied this after the code similar to it in `cmd/download`. It should print out any error message the api returns.

I haven't implemented tests, but I have some thoughts on it. `fakeSubmitServer` should test for some case where the api should return an error message (like when a duplicate submission happens) and then the test should check that `runSubmit` returns an error. Not sure how to do this though. I'm a go noob 😛

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"mime/multipart"
+	"net/http"
 	"os"
 	"path/filepath"
 
@@ -249,6 +251,15 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode == http.StatusBadRequest {
+		var jsonErrBody apiErrorMessage
+		if err := json.NewDecoder(resp.Body).Decode(&jsonErrBody); err != nil {
+			return fmt.Errorf("failed to parse error response - %s", err)
+		}
+
+		return fmt.Errorf(jsonErrBody.Error.Message)
+	}
+
 	bb := &bytes.Buffer{}
 	_, err = bb.ReadFrom(resp.Body)
 	if err != nil {
@@ -271,4 +282,11 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
 func init() {
 	RootCmd.AddCommand(submitCmd)
+}
+
+type apiErrorMessage struct {
+	Error struct {
+		Type    string `json:"type"`
+		Message string `json:"message"`
+	} `json:"error,omitempty"`
 }
--- a/cmd/submit_test.go
+++ b/cmd/submit_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
@@ -504,6 +505,8 @@ func fakeSubmitServer(t *testing.T, submittedFiles map[string]string) *httptest.
 			}
 			submittedFiles[fileHeader.Filename] = string(body)
 		}
+
+		fmt.Fprint(w, "{}")
 	})
 	return httptest.NewServer(handler)
 }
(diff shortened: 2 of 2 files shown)


Smarticles101:
@jdsutherland yeah, just a bit busy recently.

Thanks for the comments @kytrinyx, I should have some time today or tomorrow to try and throw a test together

Smarticles101:
I implemented a test that whatever error message is returned by the server gets printed.

I struggled with some `unsupported protocol scheme` error for a while that was caused by not having this `v.Set("apibaseurl", errts.URL)`.

For right now is it necessary that we even use `fakeSubmitServer`? The server I implemented, simple as it is, is unique to the one test. Would refactoring add value for other tests?

Smarticles101:
I implemented a test that whatever error message is returned by the server gets printed.

I struggled with some `unsupported protocol scheme` error for a while that was caused by not having this `v.Set("apibaseurl", errts.URL)`.

For right now is it necessary that we even use `fakeSubmitServer`? The server I implemented, simple as it is, is unique to the one test. Would refactoring add value for other tests?

jdsutherland:
Yeah, `v.Set("apibaseurl", errts.URL)` is required to make `runSubmit` hit the httptest.Server.

I don't think `fakeSubmitServer` is necessary because the CLI doesn't need to concern itself with the API implementation details - the CLI just needs to deal with the response.

kytrinyx:
> I don't think `fakeSubmitServer` is necessary because the CLI doesn't need to concern itself with the API implementation details - the CLI just needs to deal with the response.

Agreed.

My comment above was more to make sure that you didn't try to shoehorn your thing into `fakeSubmitServer` than to suggest that eventually we _should_ use it. Sorry about the confusion!

cmd/submit.go
```diff
@@ -240,6 +242,15 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	}
 	defer resp.Body.Close()
 
+	var jsonErrBody apiErrorMessage
+	if err := json.NewDecoder(resp.Body).Decode(&jsonErrBody); err != nil {
+		return fmt.Errorf("failed to parse response - %s", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
jdsutherland: We probably want to check the entire range of success (2xx). We currently PATCH and expect a 201 response, so this will fail as is.

cmd/submit.go
```diff
@@ -240,6 +242,15 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	}
 	defer resp.Body.Close()
 
+	var jsonErrBody apiErrorMessage
```
jdsutherland: It probably makes sense to put this in the `if` block since we only need to do this work when there is an error response.

jdsutherland:
I've been refactoring submit and noticed something I think might be worth pointing out. In thinking about validation for the `ID` used in the request, if the ID (and resulting URL) 404's, with these changes, we'd get an error trying to parse the response.

We might want to explicitly handle 404's to help signal that the problem lies with the ID.

cmd/submit.go
Smarticles101: If we expect a 201 response wouldn't it be better to use `http.StatusCreated`?

cmd/submit.go
Smarticles101: 👍

cmd/submit.go
jdsutherland: Given that we're decoding an explicit error response, this would be brittle. The API might change to return something other than 201 and this would signal false errors.

I don't want to overstep any boundaries here. Definitely take what I say with a grain of salt until a committer chimes in.

Smarticles101: 👍

kytrinyx:
> if the ID (and resulting URL) is bad and 404's, in `master` we get output saying submit was successful (even though 404) and with these changes we'd get an error trying to decode the response.

Ouch, yes this is not good at all. @jdsutherland would you mind opening an issue for this?

cmd/submit_test.go
```diff
@@ -534,6 +537,46 @@ func TestSubmitRelativePath(t *testing.T) {
 	assert.Equal(t, "This is a file.", submittedFiles["file.txt"])
 }
 
+func TestSubmitServerErr(t *testing.T) {
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, `{"error": {"type": "error", "message": "test error"}}`)
+	})
+
+	errts := httptest.NewServer(handler)
```
kytrinyx: We only have one test server in this test, right? If so, can we name the variable `ts` for consistency with the other tests? By giving it a different name I find myself reading everything more carefully to see in what way it is special or what the other servers in the test are called to require it to be special.

cmd/submit_test.go
```diff
@@ -534,6 +537,46 @@ func TestSubmitRelativePath(t *testing.T) {
...
+	defer errts.Close()
+
+	tmpDir, err := ioutil.TempDir("", "secret-temp-dir")
```
kytrinyx: Let's give this a prefix that reflects the purpose of the test. Maybe something like `submit-error`

Smarticles101:
@kytrinyx thanks for reviewing this and getting some of the other pull requests closed out.

 I’ll commit those changes when I get a chance in the upcoming days

Smarticles101:
This should be ready for review now, I've resolved your comments @kytrinyx, resolved merge conflicts and then squashed the changes into one commit. :)

PR #774, opened 2018-12-08
Exercise dirname matches metadata exercise slug

### LINKED ISSUE #773: Ensure exercise directory name matches exercise slug in metadata
Currently, when someone has an exercise directory that is different from the exercise slug, the CLI submits the absolute path on the user's filesystem as the absolute path.

```
# Example 1
/home/username/exercism/go/tournament-2/tournament.go

# Example 2
/Users/username/Exercism/go/__crypto-square/crypto_square.go
```

The API is unable to handle paths with leading slashes, so when the CLI requests these files we get a 404 response from the API, even though the API is the one who told us in the first place that this is the correct filepath to request.

Even if the API were able to return the file, we'd end up with a very strange file tree for the exercise.

So! The correct solution is to not let this happen in the first place.

We should ensure that the directory that the exercise lives in matches the name of the exercise slug, which can be found inside of the `metadataFilepath`.

See https://github.com/exercism/exercism/issues/4503 for background and the details of the investigation into this bug.

jdsutherland:
Closes #773

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -143,7 +143,16 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
 	if exercise.Slug != metadata.Exercise {
 		// TODO: error msg should suggest running future doctor command
-		return errors.New("exercise directory does not match exercise slug in metadata")
+		msg := `
+
+	The exercise directory does not match exercise slug in metadata:
+
+		expected '%[1]s' but got '%[2]s'
+
+	Please rename the directory '%[1]s' to '%[2]s' and try again.
+
+		`
+		return fmt.Errorf(msg, exercise.Slug, metadata.Exercise)
 	}
 
 	if !metadata.IsRequester {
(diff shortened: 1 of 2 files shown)


cmd/submit.go
```diff
@@ -141,6 +141,11 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
+	if exercise.Slug != metadata.Exercise {
+		// TODO: error msg should suggest running future doctor command
+		return errors.New("exercise directory does not match exercise slug in metadata")
```
nywilken: Maybe we should indicate what was found, what was expected, and how to fix. When users open issues pertaining to path errors the messaging doesn't contain path information, which makes the troubleshooting a little harder than it needs to be. 

`exercise directory does not match slug: expected %s but found %s. Please rename the directory and try again. `

nywilken:
@jdsutherland thanks for making the updates. @Smarticles101 thanks for reviewing as well.

PR #782, opened 2018-12-14
Fix Travis failing golang/dep install

### LINKED ISSUE #781: Travis fails to install dep
Travis is failing to install `dep`. See [Travis job](https://travis-ci.org/exercism/cli/jobs/467827450).
```
$ go get -u github.com/golang/dep/...
package github.com/golang/dep/...: github.com/golang/dep/...: invalid import path: malformed import path "github.com/golang/dep/...": double dot
The command "go get -u github.com/golang/dep/..." failed and exited with 1 during .
```

jdsutherland:
Closes #781 

It seems there are a couple of potential fixes involving changing how `dep` gets installed:

1. `go get -u github.com/golang/dep/cmd/dep` - see https://github.com/exercism/cli/pull/782/commits/5b8988f6aacc39e8c236f5cbb3da708313ea91a3 

2. Install via github releases via golang/dep [install script](https://github.com/golang/dep/blob/master/install.sh) - see https://github.com/exercism/cli/pull/782/commits/9b1321104c8117879195a233562ac5a75926b3a3

https://golang.github.io/dep/docs/installation.html recommends using releases for stability but I'm not sure how workable that is with the AppVeyor install. It seems like a bad idea to have `dep` install versions differ for Travis vs AppVeyor; on the hand, it seems that `dep` might be more prone to similar failures in the future by not using a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,9 @@ go:
   - tip
 
 install:
-  - go get -u github.com/golang/dep/...
+  - go get -u github.com/golang/dep/cmd/dep
   - dep ensure
 
 script:
   - ./.travis.gofmt.sh
   - go test -cover ./...
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,10 @@ go:
   - tip
 
 install:
-  - go get -u github.com/golang/dep/cmd/dep
+  - go get -u github.com/golang/dep/...
   - dep ensure
 
 script:
   - ./.travis.gofmt.sh
   - go test -cover ./...
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ go:
   - tip
 
 install:
-  - go get -u github.com/golang/dep/cmd/dep
+  # install dep
+  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
   - dep ensure
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ go:
   - tip
 
 install:
-  # install dep
-  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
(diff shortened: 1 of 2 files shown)


jdsutherland:
Reverting prior to the fixes, it now looks like Travis doesn't fail (╯°□°）╯︵ ┻━┻. (https://github.com/exercism/cli/pull/782/commits/2dbcef24d0a4ec4ebf23906c6826609c204cd989 shows that Travis was failing without any changes though). https://github.com/golang/dep/commits/master as of now shows shows the most recent commit being prior to this PR so it seems golang/dep repo isn't the cause here. Maybe there is some detail between `go get` and `git clone` I don't understand. Otherwise this is hard to make sense of.

Given that https://github.com/exercism/cli/pull/782/commits/2dbcef24d0a4ec4ebf23906c6826609c204cd989 proves there was a failure without changing anything, this shows that it could happen again so it still seems worth considering if the install can be made more stable (e.g. using releases).

kytrinyx:
@jdsutherland if I recall correctly, the `dep` command relies on https://github.com/exercism/cli/blob/master/Gopkg.lock -- not sure if that helps here though, because that would mean that the version is stable, and it sounds like it's not.

kytrinyx:
@jdsutherland I'm not sure what the next step is here. Rereading I think I misunderstood what "stable" we were referring to above. Using godep means that all of our other dependencies are locked to a stable version, but this doesn't actually mean that godep itself is locked to a version.

Are we OK with leaving things as they are for now?

(I agree that we probably shouldn't have different versions of godep on AppVeyor and Travis).

jdsutherland:
If you accept https://github.com/exercism/cli/pull/812 (switch to go modules) that will deprecate this, right?

kytrinyx:
Oh! Yes, that's exactly right. I completely missed that connection :-)

I intend to accept the gomodules one, once the documentation is sorted.

Shall we just close this?

PR #785, opened 2018-12-26
Silence noisy output in cmd tests

kytrinyx:
If you run the tests for just the `cmd` package from within the `cmd` directory on master, the output is very noisy:

```
You have configured the Exercism command-line client:

Config dir:                       
Token:         (-t, --token)      existing-token
Workspace:     (-w, --workspace)  
API Base URL:  (-a, --api)        http://127.0.0.1:57279


You have configured the Exercism command-line client:

Config dir:                       
Token:         (-t, --token)      a-token
Workspace:     (-w, --workspace)  
API Base URL:  (-a, --api)        http://127.0.0.1:57279


You have configured the Exercism command-line client:

Config dir:                       
Token:         (-t, --token)      replacement-token
Workspace:     (-w, --workspace)  
API Base URL:  (-a, --api)        http://127.0.0.1:57279

    http://example.com/bogus-url

PASS
ok  	github.com/exercism/cli/cmd	0.189s
```

This is because we typically write to both the `Out` and the `Err` stream, but often were only overwriting one of them in the tests.

This sets up a helper struct to make it easier to setup and reset the overrides correctly, and then uses it across the board, which silences the test output.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -20,8 +20,6 @@ var (
 	Out io.Writer
 	// Err is used to write errors.
 	Err io.Writer
-	// In is used to provide mocked test input (i.e. for prompts).
-	In io.Reader
 )
 
 const msgWelcomePleaseConfigure = `
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -3,8 +3,25 @@ package cmd
 import (
 	"fmt"
 
+	"io"
+
 	"github.com/exercism/cli/config"
 	"github.com/spf13/viper"
+)
+
+var (
+	// BinaryName is the name of the app.
+	// By default this is exercism, but people
+	// are free to name this however they want.
+	// The usage examples and help strings should reflect
+	// the actual name of the binary.
+	BinaryName string
+	// Out is used to write to information.
+	Out io.Writer
+	// Err is used to write errors.
+	Err io.Writer
+	// In is used to provide mocked test input (i.e. for prompts).
+	In io.Reader
 )
 
 const msgWelcomePleaseConfigure = `
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
 	"testing"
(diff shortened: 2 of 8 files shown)


cmd/cmd.go
```diff
@@ -1,5 +1,20 @@
...
+	// are free to name this however they want.
+	// The usage examples and help strings should reflect
+	// the actual name of the binary.
+	BinaryName string
+	// Out is used to write to information.
+	Out io.Writer
+	// Err is used to write errors.
+	Err io.Writer
+)
+
```
kytrinyx: I moved these into the `cmd.go` file from the `root.go` file, since we're collecting shared logic here. I also deleted the unused `In` variable.

cmd/cmd_test.go
```diff
@@ -90,3 +91,31 @@ func (test *CommandTest) Teardown(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// terminalOverride lets us more easily redirect streams in the tests.
+type terminalOverride struct {
+	oldOut, oldErr, newOut, newErr io.Writer
+}
```
kytrinyx: This is the new helper that I created, which lets us store off the old `Out` and `Err`, as well as the new ones that we want to override with.

cmd/cmd_test.go
```diff
@@ -90,3 +91,30 @@ func (test *CommandTest) Teardown(t *testing.T) {
...
+
+// newTerminalOverride creates a new value to override the streams.
+func newTerminalOverride() terminalOverride {
+	return terminalOverride{
+		oldOut: Out,
+		oldErr: Err,
+		newOut: ioutil.Discard,
+		newErr: ioutil.Discard,
+	}
+}
```
kytrinyx: Nearly all of the tests discard the output, so that is the default behavior that I chose.

cmd/cmd_test.go
```diff
@@ -90,3 +91,30 @@ func (test *CommandTest) Teardown(t *testing.T) {
...
+
+// override sets the package variables to the fake streams.
+func (to terminalOverride) override() {
+	Out = to.newOut
+	Err = to.newErr
+}
```
kytrinyx: I didn't keep this in the `newTerminalOverride` function for two reasons. First, sometimes we want different behavior than an `ioutil.Discard`, so that lets us set that up manually before we actually trigger the override. Second, I don't like having side effects in a `new` function, especially side effects that affect the entire package.

cmd/configure_test.go
```diff
@@ -82,6 +80,10 @@ func TestConfigureShow(t *testing.T) {
 }
 
 func TestConfigureToken(t *testing.T) {
+	to := newTerminalOverride()
+	to.override()
+	defer to.reset()
+
```
kytrinyx: I moved this one to the top since it feels cleaner to keep that setup/teardown consistent.

kytrinyx:
I've rebased onto master and resolved the conflicts with the changes that got merged in today.

cmd/cmd.go
nywilken: Good call on the consolidation.

cmd/cmd_test.go
nywilken: This is a good addition for sure. 👍🏽 

 I find the name a bit confusing as it deals more with output redirection that it does terminal overrides. Have you considered going with something along the lines of `CapturedOutput` or `OutputRedirect`?

nywilken:
@kytrinyx I like the approach here and think this is a nice addition to help clean up the test code.  I left a comment about the name `terminalOverride` but I think I might be nitpicking with the name. Looking forward to getting your thoughts.  None the less this is good to go.

cmd/cmd_test.go
nywilken: > I don’t like having side effects in a new function, especially side effects that affect the entire package.

That’s a good point. I like the approach! Thanks for calling out the rationale.

cmd/cmd_test.go
kytrinyx: I was really struggling to find a good name. I like both of your suggestions; I think maybe `capturedOutput` is the most expressive.

PR #787, opened 2018-12-30
Change !os.IsNotExist to os.IsExist

Smarticles101:


Smarticles101:
Okay... So `os.IsExist(err) != !os.IsNotExist(err)`. I thought the contrary but looking into it, I found [this gist](https://gist.github.com/mastef/05f46d3ab2f5ed6a6787) which explains it pretty well. Basically, the methods are just error checkers, they check for errors that are expected if a file exists or doesn't. 

Since we query the file with [Lstat](https://golang.org/pkg/os/#Lstat), we expect the file to exist already, so it wouldn't throw an error if the file exists.

My fault for not testing it before I pushed it, but cool to learn that.

cmd/configure.go
```diff
@@ -164,7 +164,7 @@ func runConfigure(configuration config.Config, flags *pflag.FlagSet) error {
 		workspace = config.DefaultWorkspaceDir(configuration)
 
 		// If it already exists don't clobber it with the default.
-		if _, err := os.Lstat(workspace); !os.IsNotExist(err) {
+		if _, err := os.Lstat(workspace); os.IsExist(err) {
```
nywilken: `os.Lstat` will return nil  if the path exist. `os.IsExist` requires error to be ErrExist which is why file existence is being checked with `!os.IsNotExist`; Lstat will always return ErrNotExist  if the path is not found otherwise nil (if there are no other errors).

Smarticles101: Yeah, I found that out lol. Thanks!

Issue #460, opened 2017-08-18
nextercism: Consider a --remote flag for open


kytrinyx:
Open has some nice functionality that lets you specify just the name of an exercise, and it will find all the possible matches that you have locally, and ask which one you want to open.

Many people work on multiple computers, or might not have all their Exercism work downloaded to their computer, which means that they might actually want to open something that they don't have locally.

In this case, it would be nice to provide a `--remote` flag which would ask the API what potential exercises we are talking about.

This will require defining in detail what the API needs to return and how the CLI will display this.

My gut feeling is that it should display the list exactly as though it were local ones.

```
We found more than one. Which one did you mean?
Type the number of the one you want to select it.

  [1] python/clock
  [2] go/clock
  [3] ruby/clock
  [4] elixir/clock
  [5] haskell/clock

> 
```

When implementing this, go ahead and write a test with a test server to fake out the API, until you get it the way you want it. A good starting point is probably:

```
{
	"solutions": [{
		"id": "abc-123",
		"url": "https://v2.exercism.io/my/solutions/abc-123",
		"track": {
			"track": "cpp",
			"language": "C++"
		},
		"exercise": {
			"id": "clock"
		}
	}]
}
```

palash25:
I'm interested in this issue. Can you direct me to the files that need to be changed?

kytrinyx:
Hi @palash25, that's great.

This needs to be based off of the `nextercism` https://github.com/exercism/cli/tree/nextercism branch, and the command in question is the `open` command, which can be found in this file:
https://github.com/exercism/cli/blob/nextercism/cmd/open.go

You would need to create an `open_test.go` file in the `cmd` directory. You can see how we've added tests to https://github.com/exercism/cli/blob/nextercism/cmd/configure_test.go (or other tests in that directory).

We have an abstraction that helps us do some of the setup here: https://github.com/exercism/cli/blob/nextercism/cmd/cmd_test.go

PR #788, opened 2019-01-07
improve open cmd

Smarticles101:
This pull request aims to improve the open command.

I start by solving #460, while diving into the open command, I found that it could be improved in a few various ways.


cmd/open.go
```diff
@@ -17,15 +27,95 @@ Pass the path to the directory that contains the solution you want to see on the
...
+		defer res.Body.Close()
+		var payload openPayload
+		if err := json.NewDecoder(res.Body).Decode(&payload); err != nil {
+			return fmt.Errorf("unable to parse API response - %s", err)
+		}
+
+		if res.StatusCode != http.StatusOK {
+			switch payload.Error.Type {
+			case "track_ambiguous":
+				return fmt.Errorf("%s: %s", payload.Error.Message, strings.Join(payload.Error.PossibleTrackIDs, ", "))
```
Smarticles101: #460 suggests a prompt such as this:
```
We found more than one. Which one did you mean?
Type the number of the one you want to select it.

  [1] python/clock
  [2] go/clock
  [3] ruby/clock
  [4] elixir/clock
  [5] haskell/clock

> 
```

Rather than following this approach, I decided to stay consistent with the way `download` handles the `track_ambiguous` error and that is just by returning the error.

Smarticles101: #460 suggests a prompt such as this:
```
We found more than one. Which one did you mean?
Type the number of the one you want to select it.

  [1] python/clock
  [2] go/clock
  [3] ruby/clock
  [4] elixir/clock
  [5] haskell/clock

> 
```

Rather than following this approach, I decided to stay consistent with the way `download` handles the `track_ambiguous` error and that is just by returning the error.

cmd/open.go
```diff
@@ -101,35 +105,41 @@ func runOpen(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+
+			if meta.Exercise != exerciseID {
+				continue
+			}
+
+			if trackID != "" && meta.Track != trackID {
+				continue
+			}
+
+			if meta.Team != teamID {
```
Smarticles101: Initially, I tried `teamID != "" && meta.Team != teamID`, but the current boolean statement is more in line with how the api handles team exercises.

cmd/open_test.go
```diff
@@ -0,0 +1 @@
+package cmd
```
Smarticles101: I'd like to add tests after this is reviewed, to check that functionality is how we want it

workspace/workspace.go
```diff
@@ -57,6 +57,16 @@ func (ws Workspace) PotentialExercises() ([]Exercise, error) {
 			continue
 		}
 
+		if topInfo.Name() == "teams" {
```
Smarticles101: I needed to add support for teams to `PotentialExercises` as before this change, it doesn't find them

cmd/open.go
kytrinyx: I agree. I need to go through the old issues and get rid of all the references to the interactive stuff. It's too complicated and makes the whole thing harder to use.

kytrinyx:
I completely failed to keep up with my Exercism notifications, and I'm inexcusably late to the party. To add to this, my laptop crashed and is now in for repair, and as of today I have a loaner that I've not quite gotten set up to be usable. So. My apologies (first) and (second)... my goal is to get through all the open PRs in the CLI track over the course of the next week. So sorry to leave you hanging!

Smarticles101:
> I completely failed to keep up with my Exercism notifications, and I'm inexcusably late to the party. To add to this, my laptop crashed and is now in for repair, and as of today I have a loaner that I've not quite gotten set up to be usable. So. My apologies (first) and (second)... my goal is to get through all the open PRs in the CLI track over the course of the next week. So sorry to leave you hanging!

Don't worry about it too much! Things happen and I'm pretty sure most people here are guilty of ignoring notifications for a long time. Completely understandable

kytrinyx:
Hey @Smarticles101 -- overall it looks like what you have here is much better.

In order to help me review this properly, would you summarize exactly what changed, and how this will change the usage of the command?

Smarticles101:
The old open command worked in a way similar to this:
```
exercism open ./path/to/exercise/folder
```

The new command, rather than using an exercise path, follows a style similar to the other commands. Now instead of passing an exercise directory, you pass a slug.
```
exercism open --exercise slug
```

I also added support for `--team`, which works locally or remotely, and `--remote`, which opens exercises you've submitted to the site, but may not have locally.

kytrinyx:
Thank you, that's really helpful!

In order to not break people's work flow we'll need to support the original usage as well, though... at least until we do a major version bump.

Smarticles101:
@kytrinyx That sounds good, I can add that back in when I have free time tomorrow

Smarticles101:
@kytrinyx original usage is supported now :)

Smarticles101:
No idea why the build is failing, I haven't changed anything that should make it fail, and I've rebased against master and compiled it locally and it seems to work?

kytrinyx:
> original usage is supported now :)

w000p!!!!

> No idea why the build is failing

I'll take a look.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+
+		matchingExerciseMeta := make([]*workspace.ExerciseMetadata, 0, len(exercises))
+		for _, exercise := range exercises {
+			metaDir := exercise.MetadataDir()
+			meta, err := workspace.NewExerciseMetadata(metaDir)
+			if err != nil {
+				return err
+			}
+
+			if meta.Exercise != exerciseID {
```
kytrinyx: This needs to be `meta.ExerciseSlug` to pass the tests (see https://github.com/exercism/cli/pull/815)

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+		return runOpen(cfg, cmd.Flags(), args)
+	},
+}
+
+func runOpen(cfg config.Config, flags *pflag.FlagSet, args []string) error {
+	var url string
+	usrCfg := cfg.UserViperConfig
+	trackID, _ := flags.GetString("track")
+	exerciseID, _ := flags.GetString("exercise")
+	teamID, _ := flags.GetString("team")
```
kytrinyx: I know error checking seems tedious, but I'd rather we did it for completion. If it gets really bad we can extract it into a validator type, but for now let's just do it inline.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+
+		return runOpen(cfg, cmd.Flags(), args)
+	},
+}
+
+func runOpen(cfg config.Config, flags *pflag.FlagSet, args []string) error {
+	var url string
+	usrCfg := cfg.UserViperConfig
+	trackID, _ := flags.GetString("track")
+	exerciseID, _ := flags.GetString("exercise")
```
kytrinyx: Let's name this `exerciseSlug` for consistency with the rest of the codebase (we don't currently call it exerciseID anywhere).

cmd/open.go
kytrinyx: Cat we go with just `team` as the variable name? We don't call this teamID anywhere else.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+	if exerciseID == "" {
+		// if no --exercise is given, use original functionality
 		metadata, err := workspace.NewExerciseMetadata(args[0])
 		if err != nil {
 			return err
 		}
 		browser.Open(metadata.URL)
 		return nil
-	},
+		//return fmt.Errorf("Must provide an `--exercise`")
```
kytrinyx: Let's ✂️ the commented out stuff.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+func runOpen(cfg config.Config, flags *pflag.FlagSet, args []string) error {
+	var url string
+	usrCfg := cfg.UserViperConfig
+	trackID, _ := flags.GetString("track")
+	exerciseID, _ := flags.GetString("exercise")
+	teamID, _ := flags.GetString("team")
+
+	if exerciseID == "" {
+		// if no --exercise is given, use original functionality
 		metadata, err := workspace.NewExerciseMetadata(args[0])
```
kytrinyx: How about extracting `args[0]` into a variable `path` up at the start where we're also extracting from flags?

```go
var path string
if args.length > 0 {
  path = args[0]
}
```

That way down here we can say `if path != "" {`

If we want to be really explicit, we could also have a validation on `path != "" and exerciseSlug != ""` and explain that you need one or the other.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
 		if err != nil {
 			return err
 		}
 		browser.Open(metadata.URL)
 		return nil
-	},
+		//return fmt.Errorf("Must provide an `--exercise`")
+	}
+
+	if remote, _ := flags.GetBool("remote"); remote {
```
kytrinyx: Let's extract the remote flag up top as well, and check the error.

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+		res, err := client.Do(req)
+		if err != nil {
+			return err
+		}
+
+		defer res.Body.Close()
+		var payload openPayload
+		if err := json.NewDecoder(res.Body).Decode(&payload); err != nil {
+			return fmt.Errorf("unable to parse API response - %s", err)
+		}
```
kytrinyx: Would it make sense to use the api error decoder helper?
https://github.com/exercism/cli/blob/c3525cf31135f24a84f8127f553e90b539b0298d/cmd/cmd.go#L78

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+
+		if res.StatusCode != http.StatusOK {
+			switch payload.Error.Type {
+			case "track_ambiguous":
```
kytrinyx: I wonder if we can make the error decoder helper smarter and add this logic into it.
https://github.com/exercism/cli/blob/c3525cf31135f24a84f8127f553e90b539b0298d/cmd/cmd.go#L78

(If so, let's do that in a separate PR and get that into master quickly so we can use it elsewhere).

cmd/open.go
```diff
@@ -15,17 +25,155 @@ var openCmd = &cobra.Command{
...
+				return fmt.Errorf("%s: %s", payload.Error.Message, strings.Join(payload.Error.PossibleTrackIDs, ", "))
+			default:
+				return errors.New(payload.Error.Message)
+			}
+		}
+
+		url = payload.Solution.URL
```
kytrinyx: Let's open the URL and return here so that we can out-dent the `else`.

workspace/workspace.go
kytrinyx: Would you mind extracting this change into a separate PR? (We should probably include tests for the change).

Smarticles101:
I made a mess of my git tree today and ended up having to cherry-pick my commits out.

I updated it to use `ExerciseSlug` now so the build passes. I'll see when I have time to look into your other comments. I should get some time early next week

kytrinyx:
@Smarticles101 just a quick note to say that I *think* the next step here is to rebase onto master before I do another review. If I'm wrong about that please let me know :-)

cmd/open.go
Smarticles101: I need to decode the json anyways for the solution url. The helper only returns the error. Would it be beneficial to introduce a helper that decodes the entire api response?

kytrinyx: Hm. Maybe, but let's hold off on that for now.

kytrinyx:
@Smarticles101 Is this ready for another round of review?

cmd/open.go
```diff
@@ -15,17 +25,182 @@ var openCmd = &cobra.Command{
...
+	if len(args) > 0 {
+		path = args[0]
+	}
+
+	if exerciseSlug == "" {
+		if path == "" {
+			return fmt.Errorf("must provide an --exercise slug or an exercise path")
+		}
+		// if no --exercise is given, use original functionality
+		metadata, err := workspace.NewExerciseMetadata(path)
```
kytrinyx: I think that this would be easier to understand if we did something like this:

```
if exerciseSlug == "" && path == "" {
  return fmt.Errorf...
}

if path != "" {
  // get metadata and open the url
}

// etc
```

Smarticles101:
I know it's been nearly two years, but I thought I would get this finished up. I've addressed the last changes requested. If everything looks good I will write some tests.

Smarticles101:
@ekingery if you'd like to take a look at this and let me know if you see anything that's off, I'd appreciate it :)

ekingery:
> @ekingery if you'd like to take a look at this and let me know if you see anything that's off, I'd appreciate it :)

Thanks @Smarticles101! We're fully focused on getting configlet updated for V3 right now, but this change looks like something we should be able to get merged as soon as we can bring some focus back to the CLI and get a new version out. It looks like the build is broken at the moment, any ideas on that?

Smarticles101:
@ekingery yes, like in #957, it fails on older versions of go presumably due to outdated modules. It builds correctly on the latest version

ErikSchierboom:
This is quite an old PR and has some merge conflicts. @Smarticles101 do you still want to work on this?

Smarticles101:
@ErikSchierboom thanks for checking on this. I've been out of the loop of exercism for a few years and recently started my first full-time job out of college so I don't have much time to pick this up again. 

I'll go ahead and close this. If anyone comes across this in the future, feel free to keep working on it :)

PR #793, opened 2019-01-11
Submit handles non 2xx responses

### LINKED ISSUE #790: Submit only checks 400 when printing success (404 etc. unhandled)
Currently, submit only checks for 400 responses before printing a success message. 404s are silent and the resulting output misleadingly indicates that the submission was successful.

Mentioned in https://github.com/exercism/cli/pull/770#issuecomment-445709576.

jdsutherland:
Previously, 404s silently continue to misleading success output. This change throws an error for all non 2xx responses.

Closes #790.

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
-	"net/http"
 	"os"
 	"path/filepath"
 
@@ -268,12 +267,8 @@ func (s *submitCmdContext) submit(metadata *workspace.ExerciseMetadata, docs []w
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == http.StatusBadRequest {
-		return decodedAPIError(resp)
-	}
-
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return fmt.Errorf("submission unsuccessful (%s)", resp.Status)
+		return decodedAPIError(resp)
 	}
 
 	bb := &bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


cmd/submit.go
```diff
@@ -261,6 +261,10 @@ func (s *submitCmdContext) submit(metadata *workspace.ExerciseMetadata, docs []w
 		return fmt.Errorf(jsonErrBody.Error.Message)
 	}
 
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return fmt.Errorf("submission unsuccessful (%s)", resp.Status)
```
jdsutherland: I'm guessing we might want to say more here. What do we want this to say?

kytrinyx: The API should respond with `{"error": {"type": "some_value", "message": "A human-readable message"}}` in the case of an error. Let's use the `message` if it has something available in it, otherwise let's use the default that you have here.

kytrinyx: Actually, you wrote code that fixed it. I've extracted that here:
https://github.com/exercism/cli/pull/814

kytrinyx:
@jdsutherland Now that we've gotten some of the helpers into master, would it make sense to rework this to use the helpers?

jdsutherland:
This should be up to date now.

cmd/submit_test.go
```diff
@@ -611,6 +611,44 @@ func TestSubmitServerErr(t *testing.T) {
 	assert.Regexp(t, "test error", err.Error())
 }
 
+func TestHandleErrorResponse(t *testing.T) {
```
kytrinyx: Thank you for adding this test!

PR #797, opened 2019-01-21
Fix panic when submit not given args

jdsutherland:
Currently, submit panics when run without args (i.e., `exercism submit`) due to attempting to index
an empty slice:

    exercise, err := ctx.exercise(submitPaths[0])


kytrinyx:
@jdsutherland Life intruded, and I have failed to check in on all the CLI work! I'm so sorry to have left you hanging. I'm going to go through all the open work and see where we are, and see if I can get back on track.

jdsutherland:
@kytrinyx no problem! I don't think any sane person should expect this to be a priority over life stuff. Take your time.

PR #806, opened 2019-02-10
Make zsh completion work on $fpath.

QuLogic:
This allows the file to be placed, e.g., in a system-wide `site-functions` directory and just work automatically.

Now, I don't use zsh personally, but I tried to copy what the fd, ripgrep, and restic packages do, and also the [beginning of these zsh-completion instructions](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command). I tested it out both system-wide, and user-only (with the instructions here), and it works. I don't know if the oh-my-zsh instructions are correct though.

### DIFF
--- a/shell/README.md
+++ b/shell/README.md
@@ -17,15 +17,17 @@ adding the following snippet:
 
 ### Zsh
 
+Load up the completion by placing the `exercism_completion.zsh` somewhere on
+your `$fpath` as `_exercism`. For example:
+
     mkdir -p ~/.config/exercism
-    mv ../shell/exercism_completion.zsh ~/.config/exercism/exercism_completion.zsh
+    mv ../shell/exercism_completion.zsh ~/.config/exercism/_exercism
 
-Load up the completion in your `.zshrc`, `.zsh_profile` or `.profile` by adding
-the following snippet
+and then add the directory to your `$fpath` in your `.zshrc`, `.zsh_profile` or
+`.profile` before running `compinit`:
 
-    if [ -f ~/.config/exercism/exercism_completion.zsh ]; then
-      source ~/.config/exercism/exercism_completion.zsh
-    fi
+    export fpath=(~/.config/exercism $fpath)
+    autoload -U compinit && compinit
 
 
 #### Oh my Zsh
--- a/shell/README.md
+++ b/shell/README.md
@@ -20,13 +20,13 @@ adding the following snippet:
 Load up the completion by placing the `exercism_completion.zsh` somewhere on
 your `$fpath` as `_exercism`. For example:
 
-    mkdir -p ~/.config/exercism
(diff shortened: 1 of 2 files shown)


shell/exercism_completion.zsh
```diff
@@ -33,4 +35,4 @@ _exercism() {
     esac
 }
 
-compdef '_exercism' exercism
+_exercism "$@"
```
NobbZ: This is the wrong way.

If this file is put into the `fpath`, then its already the function `_exercism`, then you define another internal scoped funtion with the same name and unconditionally call that.

This gives us an extra level of calling.

The proper way is to remove the enclosing function definition and remove this last line.

Thats the theory at least.

Problem with this though, as far as I remember, these files are autogenerated by the CLI-Framework.

shell/README.md
```diff
@@ -17,15 +17,17 @@ adding the following snippet:
 
 ### Zsh
 
+Load up the completion by placing the `exercism_completion.zsh` somewhere on
+your `$fpath` as `_exercism`. For example:
+
     mkdir -p ~/.config/exercism
-    mv ../shell/exercism_completion.zsh ~/.config/exercism/exercism_completion.zsh
+    mv ../shell/exercism_completion.zsh ~/.config/exercism/_exercism
```
NobbZ: I do not like those living in `~/.config/exercism`, this is not a configuration file!

A more idiomatic way is to use `~/.zsh`, `~/.zsh/functions`, or `~/.zsh/completions`. Some folks might even use something like `~/.config/zsh/…`, but I haven't seen this very often in the dotfiles I studied so far.

QuLogic: I don't really like it either, but I was trying to leave the change as minimal as possible.

shell/exercism_completion.zsh
QuLogic: Of the three that I looked at, two used a function, but you're right that the majority of zsh completions I have installed do not have the extra function.

If the file isn't autogenerated, then I can certainly go that way.

shell/README.md
kytrinyx: @QuLogic I would be totally fine with having a more intrusive change, if it makes it better. Basically the shell completion scripts have not had nearly as much attention as the other parts of the code, so if this is something that you have skill and knowledge to contribute to, that would be very welcome, indeed!

I have never used zsh, so I will defer to @NobbZ with respect to the discussion of what is right and wrong.

QuLogic: If I were using zsh, I'd prefer a more XDG-compliant directory like `~/.local/share/zsh/functions`, but since I don't, I'll go with your suggestion of `~/.zsh/functions`.

QuLogic:
@NobbZ OK, hopefully that takes case of the comments you made.

PR #814, opened 2019-02-26
Extract API error logic to helper

kytrinyx:
The submit command handles API errors correctly.

We need to be able to share this logic so we can stop swallowing errors in other commands.

This adds a default error message that we can fall back to if the API doesn't return a `message`.

This change was made by @jdsutherland as part of https://github.com/exercism/cli/pull/756.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


kytrinyx:
I'm going to go ahead and merge this when the build goes green, so that we can use it in other PRs.

PR #818, opened 2019-02-26
Extract logic for the URI variable identifying the exercise to download

kytrinyx:
There is very likely a better abstraction hiding close by.
This is a minimal, cohesive change, which I think is fairly uncontroversial.
    
If we do find the better abstraction, it's likely that we'll be able to move this one to it wholesale.
    
I've left a bit of duplication in place for now, rather than make a big guess about the abstraction.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -54,18 +54,6 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Oops. Totally missed a spot here. Ignore this.

kytrinyx:
Ok, tried again. I'll merge this when it goes green so that I can keep chipping away at the download command refactor.

PR #820, opened 2019-02-27
Extract download client from download command

kytrinyx:
This consolidates the logic related to making the first HTTP request
to the API in the download command.

This should let us reuse it in the doctor command.

The abstraction has a `Do()` method, which makes it feel like the api.Client,
but it doesn't actually embed the *api.Client, as that felt like overkill.

The first commit creates the abstraction, the second uses it in the download.
It will probably be easiest to view each commit separately.

This is very much inspired by the work that @jdsutherland did in https://github.com/exercism/cli/pull/756, so I've added him as co-author on the commit.

@jdsutherland I'd love your thoughts on this variation of your approach.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2,12 +2,15 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 
 	"io"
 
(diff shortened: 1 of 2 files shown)


kytrinyx:
Closing in favor of #756

PR #821, opened 2019-02-28
Extract download type from download command

jdsutherland:
This is an attempt at a minimal first change in refactoring the download command. This change adds a download type.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -54,66 +54,29 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
-	identifier, err := downloadIdentifier(flags)
-	if err != nil {
-		return err
-	}
-	url := fmt.Sprintf("%s/solutions/%s", usrCfg.GetString("apibaseurl"), identifier)
-
-	client, err := api.NewClient(usrCfg.GetString("token"), usrCfg.GetString("apibaseurl"))
+	download, err := newDownload(flags, usrCfg)
 	if err != nil {
 		return err
 	}
 
-	req, err := client.NewRequest("GET", url, nil)
-	if err != nil {
-		return err
-	}
-
-	req, err = addQueryToDownloadRequest(flags, req)
-	if err != nil {
-		return err
-	}
-
-	res, err := client.Do(req)
-	if err != nil {
-		return err
-	}
-
-	var payload downloadPayload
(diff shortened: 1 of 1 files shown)


cmd/download.go
```diff
@@ -176,6 +139,161 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+	// user config
+	token, apibaseurl, workspace string
+
+	// optional
+	track, team string
+
+	payload *downloadPayload
+}
+
+func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
```
jdsutherland: This is currently longer than ideal. I think it makes sense to have the logic to populate the payload in the constructor though. Clients can't do anything with a download without the payload being populated and extracting it to a method doesn't seem appropriate because clients should never need to call it - it just needs to happen on creation.

cmd/download.go
```diff
@@ -176,6 +139,161 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	return nil
 }
 
+type download struct {
+	// either/or
+	slug, uuid string
+
+	// user config
+	token, apibaseurl, workspace string
```
jdsutherland: I'm not sure if most would agree to save these an individual fields rather than the config object. Calling `GetString` throughout the system seems vulnerable prone to typo's where using fields has the benefit of compiler catching it.

cmd/download.go
```diff
@@ -176,6 +139,161 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+// needsSlugXorUUID checks the presence of slug XOR uuid.
+func (d download) needsSlugXorUUID() error {
+	if d.slug != "" && d.uuid != "" || d.uuid == d.slug {
+		return errors.New("need an --exercise name or a solution --uuid")
+	}
+	return nil
+}
+
+// needsUserConfigValues checks the presence of required values from the user config.
+func (d download) needsUserConfigValues() error {
```
jdsutherland: In `cmd/cmd` we have `func validateUserConfig(cfg *viper.Viper)` but coupling to an outside validation seems bad so that's why it's duplicated here.

kytrinyx:
Sweet! I'll look at this today. This looks like a very manageable size to review, I appreciate that very much 💙 💚 💛

PR #823, opened 2019-03-01
Download response uses shared error parser

jdsutherland:
Continuing from https://github.com/exercism/cli/pull/821.

This simplifies handling the download response by using the shared API error function.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -204,26 +204,7 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	defer res.Body.Close()
 
 	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/download.go
```diff
@@ -204,26 +204,7 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
...
-
-	if res.StatusCode == http.StatusUnauthorized {
-		return nil, fmt.Errorf(
-			"unauthorized request. Please run the configure command. You can find your API token at %s/my/settings",
-			config.InferSiteURL(d.apibaseurl),
-		)
-	}
-	if res.StatusCode != http.StatusOK {
-		switch d.payload.Error.Type {
-		case "track_ambiguous":
```
jdsutherland: Can we live without this or do we need to incorporate it into these changes?

PR #826, opened 2019-03-01
Extract downloadWriter from download command

jdsutherland:
These changes create a downloadWriter type that handles writing solution files and metadata.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -197,8 +197,8 @@ func (d download) needsSlugWhenGivenTrackOrTeam() error {
 	return nil
 }
 
-func (d download) metadata() ws.ExerciseMetadata {
-	return ws.ExerciseMetadata{
+func (d download) metadata() *ws.ExerciseMetadata {
+	return &ws.ExerciseMetadata{
 		AutoApprove:  d.Solution.Exercise.AutoApprove,
 		Track:        d.Solution.Exercise.Track.ID,
 		Team:         d.Solution.Team.Slug,
@@ -211,8 +211,7 @@ func (d download) metadata() ws.ExerciseMetadata {
 }
 
 func (d download) exercise() ws.Exercise {
-	metadata := d.metadata()
-	return metadata.Exercise(d.workspace)
+	return d.metadata().Exercise(d.workspace)
 }
 
 // requestFile requests a Solution file from the API, returning an HTTP response.
@@ -257,8 +256,7 @@ func newDownloadWriter(dl *download) *downloadWriter {
 
 // writeMetadata writes the exercise metadata.
 func (w downloadWriter) writeMetadata() error {
-	metadata := w.download.metadata()
-	return metadata.Write(w.destination())
+	return w.download.metadata().Write(w.destination())
 }
 
 // writeSolutionFiles attempts to write each file from the downloaded solution.
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/exercism/cli/api"
 	"github.com/exercism/cli/config"
-	"github.com/exercism/cli/workspace"
+	ws "github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


kytrinyx:
This is a really big jump, and I'm not convinced that this is the right abstraction.

A lot of this change has to do with the solution files specifically. Would it make sense to explore something that only encapsulates solution-file related logic first, and see where that goes?

jdsutherland:
It doesn't seem like there are that many potential ways for this to end up. Can you give more of a sense the issue you have with it?

kytrinyx:
I feel like it has several jumps that make it hard to reason about the code. It hides information that seems necessary to understand what is going on.

Let me sit down on Sunday and mess around with it for a while so I can get a sense of what the options are.

jdsutherland:
I see. To me it looks pretty similar to the existing runDownload loop with the difference being that separate responsibilities have been extracted and the long reaching variables are done away with.

I started by pulling out the loop that writes solution files. Having done that, it seems like there are a few distinguishable responsibilities: `requestFile` and `sanitizeLegacyFilepath`. I also extracted `destination` as it is used a few places and seems worth following Demeter.

To my eye, this seems like a reasonable split of responsibilities. I tried starting over again but it didn't change anything for me; maybe my eyes are no good because I've looked at it too long. I'd be interested to see what you would suggest here.

kytrinyx:
Cool, yeah you might be right that this is the best split of responsibilities! I'm now very curious to see where this goes :-) I have time scheduled tomorrow to play with it.

cmd/download.go
```diff
@@ -264,6 +197,119 @@ func (d download) needsSlugWhenGivenTrackOrTeam() error {
 	return nil
 }
 
+func (d download) metadata() *ws.ExerciseMetadata {
```
nywilken: I think this method might be going beyond the scope of the download type. Which If I understand correctly is a custom type used for parsing flags and setting up the download command. Having actual exercise metadata and solutionFile data :arrow_down: tied to it doesn't seem correct.

cmd/download.go
```diff
@@ -264,6 +197,119 @@ func (d download) needsSlugWhenGivenTrackOrTeam() error {
...
+	}
+}
+
+func (d download) exercise() ws.Exercise {
+	return d.metadata().Exercise(d.workspace)
+}
+
+// requestFile requests a Solution file from the API, returning an HTTP response.
+// 0 Content-Length responses are swallowed, returning nil.
+func (d download) requestFile(filename string) (*http.Response, error) {
```
nywilken: I could see this being its own function not tied to download. Maybe this is something that should live on the Exercism Client API with a single argument being the `parsedURL.string()` or if usign @kytrinyx Solution File rerfactor `sf.url()`.

cmd/download.go
```diff
@@ -264,6 +197,119 @@ func (d download) needsSlugWhenGivenTrackOrTeam() error {
...
+	return res, nil
+}
+
+// downloadWriter writes download contents to the file system.
+type downloadWriter struct {
+	download *download
+}
+
+// newDownloadWriter creates a downloadWriter.
+func newDownloadWriter(dl *download) *downloadWriter {
```
nywilken: I think we could get a way with just initializing a literal downloaderWriter in `runDownload` and dropping the constructor seeing as it is not doing anything special. With that being said, is there a reason for returning a pointer to a downloadWriter?

jdsutherland:
@nywilken thanks for your review. I would respond to the review but I was under the impression that there isn't interest in pursuing this line based on @kytrinyx response. I was expecting to close this.

cmd/download.go
kytrinyx: I like the idea of having a helper on the client API with a single URL argument. That feels like a perfectly reasonable (and cohesive!) bit of behavior for the CLI.

nywilken:
@jdsutherland I think there was some good discovery here. I’ll go ahead and close this PR and hope that we can use the discussion as a point of reference for future PRs. Feel free to open a new PR or issue(s) for the items discussed.

PR #829, opened 2019-03-03
Enhance workspace command

larson004:
I have also reverted the change made in this PR #827
It added team exercises to Potential Exercises in workspace.

I think instead of polluting root workspace's PotentialExercises with team's exercises.
A better way would be to think of each Team directory as a seperate workspace. Since they essentially mirror the structure of the root workspace.

```
$WORKSPACE
├── go
│   └── hello-world
├── ruby
│   └── hello-world
└── teams
    └── funfun
        └── go
            └── gigasecond
    └── foobar
```

So the workspace for team funfun would be `workspace.new("/teams/funfun")`. And it's potential exercises can be retrieved just like any other workspace. I believe this is easier to wrap one's head around and lends itself better to changes. 

Thoughts?

### DIFF
--- a/cmd/workspace.go
+++ b/cmd/workspace.go
@@ -1,10 +1,15 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/exercism/cli/config"
+	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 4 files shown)


kytrinyx:
@larson004 I don't understand why you reverted the previous change. Would you please provide background for this PR?

kytrinyx:
@larson004 While I understand the direction of your change, this PR does a lot of things at once.

I think potentially the change to the workspace that you suggest makes sense. If you would like to pursue that change, please submit a PR that only makes that change to the workspace, and does not make any changes to the commands, other than to adjust for the changed workspace package.

larson004:
Alright. I'll make the change in two separate PRs.

PR #830, opened 2019-03-03
Isolate file logic from http and filesystem logic in download

kytrinyx:
This creates a new type upon which we can hang logic specific to the solution file in the big for loop in download.

@jdsutherland this is an example of what I mean by a smaller step. What I like about this abstraction is that it doesn't know anything about the filesystem or the API, it just hides some of the noisy cruft that made the for loop bigger and more unwieldy. I don't think this is the end of the refactoring, there's more to do, but I think it's a cohesive change that doesn't make guesses or hide necessary information.

@nywilken if you have time to look at this one, I'd love your input!

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -75,15 +75,11 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
-	for _, file := range download.payload.Solution.Files {
-		unparsedURL := fmt.Sprintf("%s%s", download.payload.Solution.FileDownloadBaseURL, file)
-		parsedURL, err := netURL.ParseRequestURI(unparsedURL)
-
+	for _, sf := range download.payload.files() {
+		url, err := sf.url()
 		if err != nil {
 			return err
 		}
-
-		url := parsedURL.String()
 
 		req, err := client.NewRequest("GET", url, nil)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


nywilken:
@kytrinyx @jdsutherland seeing as this change is related to the PR #826 I will comment there as well.

> What I like about this abstraction is that it doesn't know anything about the filesystem or the API, it just hides some of the noisy cruft that made the for loop bigger and more unwieldy

I agree with this statement. The for loop has a lot going on already so breaking out the details around the construction of Solution File paths and URLs helps to make the for loop grokkable. I particularly like how you can test Solution File logic directly now.

cmd/download.go
```diff
@@ -311,6 +293,51 @@ func (dp downloadPayload) metadata() workspace.ExerciseMetadata {
 	}
 }
 
+type solutionFile struct {
+	path, baseURL, slug string
+}
+
+func (dp downloadPayload) files() []solutionFile {
```
nywilken: I'm wrapping my head around the other types defined here, but for now I would recommend moving up this block so that it is closer to the `downloadPayload` type.

kytrinyx: Yepp, that's a good call!

I suppose eventually we might want to find a better home for these things, but for now since they're only used by the download command I'm just keeping them in place.

PR #834, opened 2019-03-08
Add test for Solution File Type

larson004:


cmd/download_test.go
```diff
@@ -77,6 +77,71 @@ func TestDownloadWithoutFlags(t *testing.T) {
...
+			if sf.relativePath() != expectedPath {
+				t.Fatalf("Expected path '%s', got '%s'", expectedPath, sf.relativePath())
+			}
+
+			url, err := sf.url()
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			expectedURL := sf.baseURL + tc.expectedURL
```
kytrinyx: Ah. Hm. I notice now that `tc.expectedURL` isn't really a URL, but rather a path. But we already have a `tc.expectedPath`.

I'm not sure what the best way to clarify this so we're not confused in 3 months when we've forgotten the genesis of this. I can think of a couple possible approaches that would keep us accurate:

1. ensure that `tc.expectedURL` is actually the URL, by defining it with the baseURL already in there
2. rename the fields to be `tc.expectedFilePath` and `tc.expectedURLPath` to reflect what they actually are right now

... or something else that I can't think of.

larson004:
@kytrinyx I have made the changes you suggested.

cmd/download_test.go
```diff
@@ -77,6 +77,65 @@ func TestDownloadWithoutFlags(t *testing.T) {
...
+		{
+			name:         "filename with special character",
+			file:         "special-char-filename#.txt",
+			expectedPath: "special-char-filename#.txt",
+			expectedURL:  "http://www.example.com/special-char-filename%23.txt",
+		},
+		{
+			name:         "filename with leading slash",
+			file:         "/with-leading-slash.txt",
+			expectedPath: filepath.Join("/", "with-leading-slash.txt"),
```
kytrinyx: On Linux and Mac, `filepath.Join("/", "with-leading-slash.txt")` returns `/with-leading-slash.txt`.
On Windows, `filepath.Join("/", "with-leading-slash.txt")` returns `\with-leading-slash.txt`.

I think this would be less confusing, if we said:

```
fmt.Sprintf("%s%s", os.PathSeparator, "with-leading-slash.txt")
```

Likewise in the other tests, I think we should be very explicit about where we expect a forward slash backward slash or encoded slash or an os-specific filepath separator.

larson004: How about `filepath.Join(string(os.PathSeparator), "with-leading-slash.txt")`?

kytrinyx: `filepath.Join` inserts the `os.PathSeparator`. In this case, I don't think using `filepath.Join` expresses clearly enough what would actually be the expected path, because it's not clear when it will translate a slash, when it will add another slash, etc. Looking at your suggested line of code I would wonder if it ended up with two slashes, for example.

larson004: @kytrinyx I have used fmt.Sprintf instead of filrpath.Join. Does this look better?

cmd/download_test.go
```diff
@@ -77,6 +77,65 @@ func TestDownloadWithoutFlags(t *testing.T) {
...
+		{
+			name:         "filename with special character",
+			file:         "special-char-filename#.txt",
+			expectedPath: "special-char-filename#.txt",
+			expectedURL:  "http://www.example.com/special-char-filename%23.txt",
+		},
+		{
+			name:         "filename with leading slash",
+			file:         "/with-leading-slash.txt",
+			expectedPath: fmt.Sprintf("%c%s", os.PathSeparator, "with-leading-slash.txt"),
```
kytrinyx: For readability, I'd like the hard coded parts to be part of the template:

```
fmt.Sprintf("%cwith-leading-slash.txt", os.PathSeparator)
```

larson004: Done

PR #837, opened 2019-03-22
Bring Go current in Travis

gliptak:


gliptak:
Could you review the Travis settings? It seems that ```master``` is not built for new commits, only for tags?

https://travis-ci.org/exercism/cli/builds

kytrinyx:
@gliptak Yes, this is on purpose. We build only pushes to pull requests, not pushes to branches. The only direct pushes to master are when we cut a new release, which only changes the version number and the changelog. All other changes (including by maintainers) are made via pull requests.

Exercism is a very big project, with lots of activity. If we build for every push to a branch _and_ every pull request, it means we are building twice for every push. We do not get a whole lot of free concurrent resources on Travis, so by turning off pushes to branches we are able to build a lot more commits concurrently.

(Thanks for asking... this reminded me to go in and check the settings on all the Exercism repositories, and it turned out there were a bunch that had the "build pushes to branches" setting turned on).

gliptak:
This is different than the "build flow" of other projects are interacted with, but thank you for offering the rationale.

kytrinyx:
> This is different than the "build flow" of other projects are interacted with

I don't understand what you mean. Would you please elaborate?

gliptak:
@kytrinyx I would recommend always building ```master```.

Here is a scenario, where where master build might start failing:
- ```PR1``` opened and PR build green
- ```PR2``` opened and PR build green
- ```PR1``` approved and committed
- At this point ```PR2``` no longer builds against current ```HEAD```
- ```PR2``` approved and committed and now ```master``` is broken

Always building ```master``` allows rolling back PR2 and asking for rebase/corrections. Instead, all subsequent PRs might start breaking in your current model.

kytrinyx:
Yeah, we've actually had this happen, now that I think about it.

I'm going to turn on the feature that requires that a PR be up-to-date with master before merging. That will ensure that we're protected against this case.

PR #865, opened 2019-07-31
Fix Oh my Zsh correct fpath

grisotto:
Every time I open the terminal show this:
```/.oh-my-zsh/custom/exercism_completion.zsh:local:6: options: can't change type of autoloaded parameter```
You need to put the file in any `$fpath`, but `~/.oh-my-zsh/custom` is not one of them.

Look this: https://github.com/beetbox/beets/issues/1731

### DIFF
--- a/shell/README.md
+++ b/shell/README.md
@@ -32,7 +32,11 @@ and then add the directory to your `$fpath` in your `.zshrc`, `.zsh_profile` or
 
 #### Oh my Zsh
 
-If you are using the popular [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) framework to manage your zsh plugins, move the file `exercism_completion.zsh` into `~/.oh-my-zsh/custom`.
+If you are using the popular [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) framework to manage your zsh plugins, follow these steps:
+
+    mkdir -p ~/.oh-my-zsh/completions
+    mv ../shell/exercism_completion.zsh ~/.oh-my-zsh/completions
+
 
 ### Fish
 


NobbZ:
I'm wondering why or how this works…

I'm not using `omz` anymore, but when I used it, it was discouraged to put any user provided additions outside of `custom`, as this could interfere with the way how `omz` updates itself via `git`.

It really should be in `~/.oh-my-zsh/custom`.

kytrinyx:
@NobbZ So you're saying we need to revert this?

Would someone please reference some official docs to get a canonical answer?

NobbZ:
Sadly there is not much documenation about "custom"…

https://github.com/robbyrussell/oh-my-zsh#custom-plugins-and-themes

But as far as I can tell, the exercism completion is not a "plugin" in the way as `omz` defines its plugins. While I'm not sure what would make a plugin a plugin rather than a bare script.

But, from the sources, the `fpath` is set to contain the folder mentioned by the OP in the PR when looking at the code:

https://github.com/robbyrussell/oh-my-zsh/blob/master/oh-my-zsh.sh#L20

Therefore, I conclude that `omz` lacks documentation about those folders and should also properly advertise them as they do with `custom` while ignoring their contents in `gitignore`.

@grisotto would you mind opening a ticket there and linking back?

kytrinyx:
> Therefore, I conclude that `omz` lacks documentation about those folders and should also properly advertise them as they do with `custom` while ignoring their contents in `gitignore`.

That would be really helpful!

> @grisotto would you mind opening a ticket there and linking back?

❤️ This would be a great help :-)

grisotto:
Hi, I think you can revert this PR.
I don't want open an issue in oh-my-zsh. There are a lot of issues and I think this was asked before.

kytrinyx:
Ok, I'll do that. Thanks.

PR #866, opened 2019-08-02
The API token outputted during verbose will now be masked by default

### LINKED ISSUE #860: `--verbose` switch for `submit` leaks users token (perhaps other subcommands as well)
Sometimes when debugging problems, we ask to run `exercism submit --verbose …` and post the full output into chat/github.

Today I realised that this does leak the literal user token.

Please filter/blank the token from the output generated by the `--verbose` flag, also please check for any other output if it leaks the token and filter/blank it there as well.

We should consider to invalidate all current tokens once this is fixed and deployed.

Jrank2013:
Will fix #860. The token is now masked by default and the unmask-token command will unmask it now


cmd/root.go
```diff
@@ -46,4 +49,5 @@ func init() {
 	api.UserAgent = fmt.Sprintf("github.com/exercism/cli v%s (%s/%s)", Version, runtime.GOOS, runtime.GOARCH)
 	RootCmd.PersistentFlags().BoolP("verbose", "v", false, "verbose output")
 	RootCmd.PersistentFlags().IntP("timeout", "", 0, "override the default HTTP timeout (seconds)")
+	RootCmd.PersistentFlags().BoolP("unmask-token", "", false, "will unmask the API duruing a request/response dump")
```
kotp: Typographical error.

```suggestion
	RootCmd.PersistentFlags().BoolP("unmask-token", "", false, "will unmask the API during a request/response dump")
```

debug/debug.go
```diff
@@ -41,6 +45,14 @@ func DumpRequest(req *http.Request) {
 	body := io.TeeReader(req.Body, &bodyCopy)
 	req.Body = ioutil.NopCloser(body)
 
+	temp := req.Header.Get("Authorization")
+
+	if !UnmaskAPIKey {
+		if token := strings.Split(temp, " ")[1]; token != "" {
+			req.Header.Set("Authorization", "Bearer "+utils.Redact(token))
+		}
+	}
```
kytrinyx: This is really nice! Great catch.

utils/utils.go
```diff
@@ -0,0 +1,12 @@
+package utils
+
+import "strings"
+
+/*
+Redact masks the given token by replacing part of the string with *
+*/
```
kytrinyx: For consistency with the rest of the codebase, would you mind using `//`-style comments?

debug/debug.go
Jrank2013: Thanks!

Jrank2013:
What suggestions did you have for Redact?

kytrinyx:
Shoot, I wonder why that didn't get posted. I would like to avoid having a `utils` package. Can we keep `Redact` in the `debug` package?

Jrank2013:
Updated

PR #871, opened 2019-08-13
Error message returned if the track is locked

### LINKED ISSUE #870: Exercism 3.0.12 no longer reports errors when trying to download a locked exercise
Steps to reproduce:

1. Try to download an exercise that you have not yet unlocked
2. Notice the output "Downloaded to (exercism-workspace-dir)/users"
3. Check the users directory as well as the track directory and notice no folders have been downloaded

This was working properly before the recent update.

Jrank2013:
This fixes #870.

If the user has not unlocked the track that they have requested to download, the cli will return an error.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -188,6 +188,9 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
 		return nil, decodedAPIError(res)
 	}
+	if d.payload.Error.Message != "" {
+		return nil, errors.New(d.payload.Error.Message)
+	}
 
 	return d, nil
 }
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -1,10 +1,12 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	netURL "net/url"
 	"os"
@@ -185,11 +187,11 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	}
 	defer res.Body.Close()
 
-	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
+	body, err := ioutil.ReadAll(res.Body)
+	res.Body = ioutil.NopCloser(bytes.NewReader(body))
+
+	if err := json.Unmarshal(body, &d.payload); err != nil || res.StatusCode < 200 || res.StatusCode > 299 {
 		return nil, decodedAPIError(res)
-	}
-	if d.payload.Error.Message != "" {
-		return nil, errors.New(d.payload.Error.Message)
 	}
 
 	return d, nil
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -187,10 +187,14 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	}
 	defer res.Body.Close()
 
-	body, err := ioutil.ReadAll(res.Body)
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		return nil, decodedAPIError(res)
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/download.go
```diff
@@ -188,6 +188,9 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
 		return nil, decodedAPIError(res)
 	}
+	if d.payload.Error.Message != "" {
+		return nil, errors.New(d.payload.Error.Message)
+	}
```
kytrinyx: On looking at this I'm wondering why the `decodedAPIError` above doesn't return the error. It seems like it should, right?

https://github.com/exercism/cli/blob/4ad01b1562f320ac6c75b9463827b5d5857f4f03/cmd/cmd.go#L98

Jrank2013: You were right. There wasn't a check for the http status code which was causing the issue

cmd/download.go
```diff
@@ -185,7 +187,10 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	}
 	defer res.Body.Close()
 
-	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
+	body, err := ioutil.ReadAll(res.Body)
+	res.Body = ioutil.NopCloser(bytes.NewReader(body))
+
+	if err := json.Unmarshal(body, &d.payload); err != nil || res.StatusCode < 200 || res.StatusCode > 299 {
```
kytrinyx: The `decodedAPIError` already unmarshals the error, right? Can we move all the error handling logic into that method?

cmd/download_test.go
```diff
@@ -272,6 +272,42 @@ func assertDownloadedCorrectFiles(t *testing.T, targetDir string) {
 	assert.True(t, os.IsNotExist(err), "It should not write the file if empty.")
 }
 
+func TestDownloadError(t *testing.T) {
```
kytrinyx: 😍 Thank you!

cmd/download_test.go
```diff
@@ -272,6 +272,42 @@ func assertDownloadedCorrectFiles(t *testing.T, targetDir string) {
...
+		DefaultBaseURL:  "http://example.com",
+	}
+
+	flags := pflag.NewFlagSet("fake", pflag.PanicOnError)
+	setupDownloadFlags(flags)
+	flags.Set("uuid", "value")
+
+	err = runDownload(cfg, flags, []string{})
+
+	fmt.Println(err)
```
kytrinyx: Would you mind removing the debugging output here?

cmd/download_test.go
```diff
@@ -272,6 +272,42 @@ func assertDownloadedCorrectFiles(t *testing.T, targetDir string) {
...
+
+	assert.Regexp(t, "test error", err.Error())
```
kytrinyx: Does this need to be a regexp? The error string should be just "test error", right?

cmd/download.go
Jrank2013: The response body can only be read once. If we unmarshal or decode the body before saving a copy we wouldn't have a way to decode the JSON into the error struct.

cmd/download.go
```diff
@@ -185,7 +187,14 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	}
 	defer res.Body.Close()
 
-	if err := json.NewDecoder(res.Body).Decode(&d.payload); err != nil {
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		return nil, decodedAPIError(res)
```
kytrinyx: Now we have the `decodedAPIError` twice, which seems like it shouldn't be necessary.

I'm sorry if I'm being dumb, here. I'll take some time after work to pull this down and play around with it to see if I understand it better.

kytrinyx: Ok, I played with this and I totally see what's going on.

Sorry for being slow!

Jrank2013: No problem, just glad I can help!

PR #880, opened 2019-09-15
download: Infer track and exercise slug from CWD

petertseng:
I think this PR is highly unlikely to be merged.
But I will submit it anyway in case I am wrong.

This PR materially improves my quality of life because I make use of both forms introduced by this PR during my interaction with Exercism.
However, I think it is unlikely to be merged because my feeling is that with the increased number of options for download (`exercism download` allows `--uuid` and `--team` compared to `exercism fetch` which only had exercise slug and track), it is desirable to be more explicit rather than implicit as this PR does. This PR makes too many implicit assumptions that are correct for one person's use cases, but are not necessarily correct for any other person's.

You need not feel guilt for closing this PR; it will not harm my workflow (because I will simply run it on my copy of the cli).

Original commit message below:

---

If I'm in the `exercism/c` directory and invoke:

    exercism download --exercise hello-world

I would like `--track c` to be inferred.

If I'm in the `exercism/c/hello-world` directory and invoke without
arguments:

    exercism download

I would like both `--track c` and `--exercise hello-world` to be
inferred.


kytrinyx:
I'm ok with using inference as long as the context is explicit. If the context cannot be guessed, then I'll want to provide an error message.

I will review the changes in the next few days (sorry it's taken so long to get to it)

petertseng:
Let me just pre-empt myself a bit here. So I realised I need to exclude `users` and `teams` from being considered track names. Just like `https://github.com/exercism/cli/blob/master/workspace/workspace.go#L44-L60` does. So I should probably add a test for those special dirs too.

This is part of what makes this PR so difficult to craft and (I am sure) review - the `download` command has at least three different modes of operation (download mine, download someone else's, download a team's), and I am adding functionality to just the first modes and until just now was completely unaware of the other two. I need to make sure I do not stomp on the other modes.

petertseng:
EBWODP: proper link: https://github.com/exercism/cli/blob/master/workspace/workspace.go#L44-L60

petertseng:
I had to make some significant changes to make it not do the wrong thing on `users` and `teams`. 

Main problem I ran into was that for `users/whatever-user/whatever-track/whatever-slug`, the previous code would set the slug to `whatever-user`, even if I made it not set track to `users`. To fix this problem, I had to have the inferring-the-slug part also check not only the second path component but also the first.

cmd/download.go
```diff
@@ -162,6 +162,19 @@ func newDownload(flags *pflag.FlagSet, usrCfg *viper.Viper) (*download, error) {
 	d.apibaseurl = usrCfg.GetString("apibaseurl")
 	d.workspace = usrCfg.GetString("workspace")
 
+	if d.uuid == "" {
+		if d.slug == "" {
+			if _, slug, ok := trackAndSlugFromCwd(d.workspace); ok && slug != "" {
```
petertseng: oh, the `!= ""` check is, strictly speaking unnecessary. If we have gotten to this point, then `d.slug` is already `""` and setting it from `""` to `""` does nothing.

cmd/download.go
```diff
@@ -349,6 +362,43 @@ func (sf solutionFile) relativePath() string {
 	return filepath.FromSlash(file)
 }
 
+func trackAndSlugFromCwd(workspace string) (string, string, bool) {
```
petertseng: surely, it would be possible to call this `flagsFromCwd` and have it automatically detect the `team` if inside a teams directory (from CWD) and the UUID if inside `users` (but this requires reading a file in order to get the UUID). Since I don't use either of these pieces of functionality, and the latter needs more thought, I would not like to take on such a task in this PR.

ErikSchierboom:
> If I'm in the `exercism/c` directory and invoke:
>
>    exercism download --exercise hello-world
>
> I would like `--track c` to be inferred.

👍 I quite like this. It would be really useful.

> If I'm in the `exercism/c/hello-world` directory and invoke without
> arguments:
> 
>     exercism download
> 
> I would like both `--track c` and `--exercise hello-world` to be
> inferred.

I'm not certain how useful this would be. I'd be more than happy to try and get this PR merged with just the first option.

@petertseng are you still interested in working on this PR? If not, no worries.

PR #886, opened 2019-10-08
Added GoReleaser config, updated docs, made archive naming adjustments

### LINKED ISSUE #700: Release tarballs are named ambiguously
From @vbuell in https://github.com/exercism/interactive-cli-walkthrough/issues/58:

> The sentence "32-bit vs 64-bit vs ARM" is confusing since ARM has 64 extension as well. Would be cleaner if it is replaced with something like "x86 (32-bit) vs x86_64 vs ARM". The same issue with the naming of released tarballs - the name should contain full qualified architecture name, not just a width of the cpu registers.

ekingery:
This PR adds support for using [GoReleaser](https://goreleaser.com) and closes #700. Additional information on the transition to GoReleaser and adding Snap support for Linux is available here: https://gist.github.com/ekingery/4732af2a828d5992b1cc69d07c87778d

For reference, you can find a test release on my fork: https://github.com/ekingery/cli/releases/tag/v3.0.38


ekingery:
> This looks good! Thank you so much for taking the time to work through this whole process.
> 
> Would you mind rebasing onto master and editing the commit to run gofmt on it?

Absolutely. This is done and the build is passing. The rebase got a little messy but I think everything looks OK now.

PR #892, opened 2019-10-25
improvement to bash snippet in RELEASE doc: don't parse ls

glennj:
Fixing up the shell snippet: [Don't parse `ls`](https://mywiki.wooledge.org/ParsingLs)


kotp:
Thanks @glennj 

I can't merge this in, but I can approve it.

Issue #824, opened 2019-03-01
Feature: infer the file to be submitted


dogweather:
I'd like it if, as a student, I could simply run: 

```bash
~/Exercism/python/two-fer $ exercism submit
```

...and it'd infer that the argument is `two_fer.py`. I think this would be big UX win. I don't know, though, if any exercises are so complex or oddly named that the solution filename couldn't be inferred. (?) If this is a problem, we could introduce an exercise metadata file. This could also help #230:

```json
"meta_data": {
    "solution_file": "two_fer.py",
    "test_file": "two_fer_test.py"
}
```

Open to a PR for this?

kytrinyx:
The solutions do have a metadata file, but it doesn't say what the solution would be named.

kytrinyx:
Sorry, accidentally submitted before I was done typing.

The long and short of it is that with over 120 possible exercises in over 50 different programming languages, we can't currently infer the exercise name. It would require an extensive change in over 60 different repositories to accomplish this, and at the moment I think that the effort of implementing and maintaining it would be more than the perceived gains in terms of UX.

I'm going to leave this open for a bit in case someone can think of how to do this more easily.

dogweather:
Oh! I didn't think to check for a metadata file. And it has a unique identifier for the solution:

```
{
  "track":"python",
  "exercise":"pangram",
  "id":"3f6863eba984439fafcb595be59f7d7e",
  # ...
}
```

So one solution could be a new server endpoint which returns the required exercise filename, given a solution id. And then `submit` would first call this to discover which file to submit.

I'm guessing that the server (1) can retrieve and exercise given a solution, and (2) has the knowledge of which

kytrinyx:
The API currently does not have any knowledge about the expected names of solution files. Most language tracks follow a convention, but it's a different convention from programming language to programming language, and we have no way of guessing. It's not currently written down in a way that is programmatically accessible.

Some language tracks require submitting multiple files.

I don't see a way of doing this that doesn't require quite a lot of work and make things quite significantly more complex.

petertseng:
My half-baked ideas of how to do this were at https://github.com/exercism/cli/issues/370#issue-207011775. I ended up going with a local list of rules hard-coded per language. It has served me well enough. https://gist.github.com/petertseng/e3e88bf1c383865ff67f4095413993b2. But I did not submit it as a PR for this repo because 1. It is written in the wrong language for this repo 2. Bad UX for languages it doesn't know about.

[...]

PR #929, opened 2020-04-24
Allow to submit without specifying solution file

chocopowwwa:
This pr allows submitting exercism solution without specifying the actual solution file, I tried to implement this as suggested in https://github.com/exercism/cli/issues/824#issuecomment-563903452 and the previous comment.

Note:
I haven't try all possible languages, I tried manually for bash and c, and my test cases only check for bash, c, cpp, crystal, dart, elixir, go, and haskell. Since this PR require some labor work (50+ languages that are active) to make sure all languages are supported, first I'd like to see some review of the code, if the code looks good then I'll continue to complete the test case.G


arlm:
Isn't this already happening? I think it could be closed.

ErikSchierboom:
This has been implemented.

PR #932, opened 2020-04-27
fix #931 set DualStack on default api client

### LINKED ISSUE #931: Dial failure on dual-stack networks without ipv6 route to internet
When on a dual-stack system with an incomplete or non-routing IPv6 stack, the first records returned are `AAAA`, forcing the client to use the ipv6 stack first. This causes a failure in the client like so:

```
$ exercism configure --token=...

Error: The base API URL 'https://api.exercism.io/v1' cannot be reached.
Get https://api.exercism.io/v1/ping: dial tcp: lookup api.exercism.io on [::1]:53: read udp [::1]:38502->[::1]:53: read: connection refused
```

This is due to the `Ping` function using an unconfigured HTTP Client (https://github.com/exercism/cli/blob/master/api/client.go#L20).

To fix this, add the [DualStack](https://golang.org/pkg/net/#Dialer) field to the DialContext in the Transport of the HTTP Client.

tydavis:
When running on broken or incomplete DualStack networks (e.g. ipv6 and
ipv4 addresses are available, but only ipv4 routes to the internet),
enable HappyEyeballs (DualStack) fast-fallback in the Dial context
settings to ensure successful connections.

RFC-6555: https://tools.ietf.org/html/rfc6555

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tydavis:
Pushed a fix to remove the unnecessary `HTTP2` change

tydavis:
DualStack is now enabled by default on newer compilers.

PR #951, opened 2020-08-10
Add command to register a custom URI handler

SaschaMann:
**This is a proof of concept, nothing more.**

---

In https://github.com/exercism/v3/discussions/2164#discussioncomment-49086 I suggested it would be nice to have a button that automatically downloads a solution or exercise instead of having to copy the command into the terminal; similar to a `spotify:track:2pnZVpeKqLrEto2X9vYF6s` URI opening the Spotify app.

This PR contains a proof of concept that adds such a handler on Windows, based on the protocol outlined on [MSDN](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa767914(v=vs.85)?redirectedfrom=MSDN) (note that the docs refer to Internet Explorer, but these handlers are system-wide). The primary purpose is to demonstrate that this is easily possible without modifying the CLI itself.

### What would be required to actually add this feature?

- Support for it in the UI (v3)
    - to be clear: I don't think this should replace the copy buttons, it should be an addition. Maybe with a dropdown button or a user setting to configure the default.
- Figure out how to test this properly. Since it modifies the registry, we need to be careful to not destroy anything.
- Figure out macOS support (https://superuser.com/questions/548119/how-do-i-configure-custom-url-handlers-on-os-x as a starting point)
- People who use Linux are proficient enough to figure this out on their own if they want to use it ¯\\\_(ツ)\_/¯
- Ideally add a command that adds the cli to PATH. Otherwise allow people to specify which path it is in.
- Add unregister command.
- Instead of launching `exercism download --uuid=...`, we should define a command `exercism uri-handler-entrypoint %1` that parses the custom URI and then executes the commands. This would allow us to run different commands, e.g. we could have `exercism:download:mentor:<uuid>` that runs  `exercism download --uuid=<uuid>` and `exercism:download:julia:leap` that runs `exercism download --exercise=leap --track=julia`. Everything that right now requires copying a command could be run through this.
- MSDN lists security precautions. I don't really see any attack vectors through the CLI, but it's something that needs to be considered.

Alternatively, but I like this option far less, we could wrap the CLI in a standard installer which can probably set this up quite easily.

@ErikSchierboom Do you know how to help with the windows API side of this? Primarily, how should one test changes to the registry etc?

### DIFF
--- a/cmd/handler-entrypoint.go
+++ b/cmd/handler-entrypoint.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 
@@ -19,13 +21,24 @@ var handlerEntrypointCmd = &cobra.Command{
 	},
 }
 
+func exit(exitCode int) {
+	fmt.Println("\nPress ENTER to exit.")
+	_, err := fmt.Scanln()
+	if err != nil {
+		log.Fatal(err)
+	}
+	os.Exit(exitCode)
+}
+
 func dispatchURICommand(cmd *cobra.Command, args []string) error {
 	if runtime.GOOS != "windows" {
-		return errors.New("only supported on Windows")
+		fmt.Println("only supported on Windows")
+		exit(1)
 	}
 
 	if len(args) < 1 {
-		return errors.New("no arguments given")
+		fmt.Println("no arguments given")
+		exit(1)
 	}
 
 	params := strings.Split(args[0], ":")
@@ -34,22 +47,28 @@ func dispatchURICommand(cmd *cobra.Command, args []string) error {
 	switch params[0] {
 	case "download":
 		uuid := params[1]
-		fmt.Printf("Downloading %s...", uuid)
-		err := downloadCmd.RunE(cmd, []string{fmt.Sprintf("--uuid=%s", uuid)})
+		_cmd := exec.Command("exercism", "download", fmt.Sprintf("--uuid=%s", uuid))
+		_cmd.Stderr = os.Stderr
+		_cmd.Stdout = os.Stdout
+		err := _cmd.Run()
 		if err != nil {
-			log.Fatalf("download failed: %v", err)
+			exit(1)
 		}
-		return nil
+		exit(0)
 	case "configure":
-		if params[2] == "token" {
-			token := params[3]
-			fmt.Printf("Configuring token=%s...", token)
-			err := configureCmd.RunE(cmd, []string{fmt.Sprintf("--token=%s", token)})
+		if params[1] == "token" {
+			token := params[2]
+			//err := configureCmd.RunE(cmd, []string{fmt.Sprintf("--token=%s", token)})
+			_cmd := exec.Command("exercism", "configure", fmt.Sprintf("--token=%s", token))
+			_cmd.Stderr = os.Stderr
+			_cmd.Stdout = os.Stdout
+			err := _cmd.Run()
 			if err != nil {
-				log.Fatalf("configuring token failed: %v", err)
+				exit(1)
 			}
+			exit(0)
 		}
-		return nil
+		exit(1)
 	}
 
 	return errors.New("invalid command")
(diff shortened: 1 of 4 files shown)


iHiD:
@SaschaMann Thank you for both the detailed issue and the POC 🙂 

> to be clear: I don't think this should replace the copy buttons, it should be an addition. Maybe with a dropdown button or a user setting to configure the default.

Could you maybe outline where the advantage is to having this URI over the command to paste?
- Would it be clickable in the UI to trigger the download via this URL?
- If not, is the advantage that you can paste it into the window you're in (e.g. firefox, rather than having to switch to a terminal etc)?

Thanks! :)

SaschaMann:
> Would it be clickable in the UI to trigger the download via this URL?

Yes. You'd click on a button and it would launch the exercism cli and run the command for you. You don't have to copy and paste anything into your terminal. I can share a screen recording of the PoC tomorrow.

> If not, is the advantage that you can paste it into the window you're in (e.g. firefox, rather than having to switch to a terminal etc)?

You could do that, but I don't think it's a big advantage.

iHiD:
> Yes. You'd click on a button and it would launch the exercism cli and run the command for you. 

Excellent. Thanks for clarifying :)

This seems like a great idea to me.

SaschaMann:
Short comment on the review requests: No point in reviewing the actual code yet, I know it's trash, but I'd appreciate some input on the general idea and the points I outlined above.

ErikSchierboom:
I really like this idea. It is also something that Slack does for example, and it is really helpful. 

> Support for it in the UI (v3)
> to be clear: I don't think this should replace the copy buttons, it should be an addition. Maybe with a dropdown button or a user setting to configure the default.

Yep, this would be nice.

> Figure out how to test this properly. Since it modifies the registry, we need to be careful to not destroy anything.

Yes, this is always problematic with the registry, although usually the problematic bit is doing the correct cleanup when removing an application. A proper integration test would involve modifying the registry, although we might be able to get away with  just unit-testing the custom entry point at first.

> Figure out macOS support (https://superuser.com/questions/548119/how-do-i-configure-custom-url-handlers-on-os-x as a starting point)
> People who use Linux are proficient enough to figure this out on their own if they want to use it ¯\_(ツ)_/¯

Agreed

> Ideally add a command that adds the cli to PATH. Otherwise allow people to specify which path it is in.

Do you mean something like `exercism register` that would add it to the path? 

> Add unregister command.

See comment on testing with registry

> Instead of launching exercism download --uuid=..., we should define a command exercism uri-handler-entrypoint %1 that parses the custom URI and then executes the commands. This would allow us to run different commands, e.g. we could have exercism:download:mentor:<uuid> that runs exercism download --uuid=<uuid> and exercism:download:julia:leap that runs exercism download --exercise=leap --track=julia. Everything that right now requires copying a command could be run through this.

This sounds good. Clearly we'd have to be careful to only handle valid input, but this is totally doable of course.

> MSDN lists security precautions. I don't really see any attack vectors through the CLI, but it's something that needs to be considered.

Did they list any particular attack vectors? As the CLI is based on the API, a lot of the security stuff is automatically handled. It would also help if we would only do read actions through this (as in only downloading things, not uploading things).

> Alternatively, but I like this option far less, we could wrap the CLI in a standard installer which can probably set this up quite easily.

What in particular do you dislike about this?

SaschaMann:
> Do you mean something like exercism register that would add it to the path?

Yes.

> What in particular do you dislike about this?

Just seems like overkill to wrap a standalone binary in an installer.

SaschaMann:
> Did they list any particular attack vectors? 

Primarily malicious inputs:

> Security Warning: Applications that handle URI schemes must consider how to respond to malicious data. Because handler applications can receive data from untrusted sources, the URI and other parameter values passed to the application may contain malicious data that attempts to exploit the handling application.

ErikSchierboom:
> Just seems like overkill to wrap a standalone binary in an installer.

Right. I can understand that :)

ErikSchierboom:
Having given this some thought, I think that for now we won't be proceeding with this PR. We might reconsider later on.

PR #952, opened 2020-08-11
Avoid using defer in loop by extracting a function

suzaku:


ErikSchierboom:
What is the benefit of not having `defer` in the loop? (I don't know much Go)

Issue #945, opened 2020-06-19
Exercism can't use mkdir


Llcoolsouder:
OS: Ubuntu

Running `exercism configure ...` gives error mkdir /home/$USER/.config/exercism permission denied
Even after running `chmod 777 ~/.config` this error still occurs. Running with sudo alleviates the problem.

But then after configuring, running `exercism download ...` I get another permission denied error for mkdir n the workspace directory even with sudo

NobbZ:
What method of installation did you choose? Who owns `~/.config`? Do you have perhaps SElinux or similar in place (`sestatus` and `getenforce` will check for SElinux availability and its current enforcement policy)?

My guess: `~/.config` is owned by you, you installed via `snap` and SElinux is currently in `Enforced` mode…

Llcoolsouder:
I installed via snap. I own `~/.config`. I don't believe I have SElinux in place. `sestatus` returns disabled.

Llcoolsouder:
Ok, so I got it working (I probably broke some things along the way, but I probably wouldn't have had any issue at all if this machine wasn't already screwed up from younger me :shrug: )

So I got all of the permission set up, but exercism still couldn't open `~/.config/exercism/user.json` because it didn't exist, so I created that file, and THEN `exercism configure <options>` worked.

Then when I ran `exercism download <options>` I was getting this message.
```
Error: 

    Welcome to Exercism!

    To get started, you need to configure the tool with your API token.
    Find your token at

        https://exercism.io/my/settings

    Then run the configure command:

        /snap/exercism/5/exercism configure --token=YOUR_TOKEN
```
So I set the symbolic link which was previously `/snap/bin/exercism -> /usr/bin/snap` to `/snap/bin/exercism -> /snap/exercism/5/exercism`.

So far everything is working now. I just think it's odd that it was an issue to begin with.

richardneish:
I'm also seeing this issue on Ubuntu 20.04, installed from snap.
Running `exercism configure --token <TOKEN>` results in the error `Error: open /home/$USER/.config/exercism/user.json: permission denied`
If I run `/snap/exercism/current/exercism configure --token <TOKEN>` it seems to work properly.

Is this a general problem with snap?

NobbZ:
There are known and reoccuring but irreproducible problems coming up with snap on a regular basis. Mostly solved by either circumventing snap sandbox or removing the snap in favor of a manual installation.

Therefore I have created #973 right now, which would basically solve all of the issues we have with snap.

[...]

PR #972, opened 2020-11-24
Allow read-write access to configuration and download directories when installed as a Snap

richardneish:
We need write access to the $HOME/.config/exercism/ directory for `exercism config` and $HOME/exercism is the default download directory so we need write access there as well.

This should resolve issue #945

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -97,4 +97,9 @@ snapcrafts:
       386: i386
     apps:
       exercism:
-        plugs: ["home", "network", "removable-media"]
+        plugs: ["home", "network", "removable-media","personal-files"]
+    plugs:
+      personal-files:
+        write:
+        - $HOME/.config/exercism/
+        - $HOME/exercism/
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -101,5 +101,4 @@ snapcrafts:
     plugs:
       personal-files:
         write:
-        - $HOME/.config/exercism/
-        - $HOME/exercism/
+        - $HOME/


.goreleaser.yml
```diff
@@ -97,4 +97,9 @@ snapcrafts:
       386: i386
     apps:
       exercism:
-        plugs: ["home", "network", "removable-media"]
+        plugs: ["home", "network", "removable-media","personal-files"]
+    plugs:
+      personal-files:
+        write:
+        - $HOME/.config/exercism/
+        - $HOME/exercism/
```
kotp, line 105: This is only the default installation directory, and should not be hard coded, i think, but should gather that from the configuration file. Otherwise we are likely going to have intermittent permission problems if the user changes the workspace.

richardneish: We can allow write access to any subdirectory of $HOME - would that be better?

kotp: I suspect that is better.  Perhaps the configuration file is not yet written at this point in time, to be able to get the workspace from the configuration.

ekingery (approved):
It is possible that the ideal solution is for a tighter restriction than $HOME. However, given the installation troubles we've seen, I think it is worthwhile to merge this as-is.

ekingery:
Thanks for the contribution @richardneish! It should be reflected in the next release of the CLI, which should correspond with the Exercism V3 launch. The build should pass after getting merged with `main`.

ekingery:
In order to merge, the build needs to pass. In order for the build to pass, we need to merge the latest changes from `main`. There was no base branch to merge into, so I created a separate PR for getting this merged: https://github.com/exercism/cli/pull/980

PR #983, opened 2021-02-22
Add --postprocess configure option to designate a script to run after downloading an exercise.

joelnordell:
This adds a `--postprocess` option to `exercism configure` that, when set, causes a designated script to be executed after each download. Additionally, there is a `--no-postprocess` option to remove the postprocess configuration. The postprocess script runs with the current directory set to the newly downloaded exercise directory.

The use case I have in mind for this, and the reason I added it, is to allow me to provide a script that automatically sets up each new downloaded exercise as a git repository so I can track my changes to the exercise:

```
#!/usr/bin/env bash

git init .
git add .
git commit -m 'Initial commit from exercism'
```

I just put that into a script, then use `exercism configure --postprocess /path/to/above/script` to set up exercism to automatically run it after each download.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -132,10 +132,10 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		cmd := exec.Command(usrCfg.GetString("postprocess"))
 		cmd.Dir = metadata.Dir
 		err := cmd.Run()
-		if err != nil {
-			fmt.Fprintf(Err, "\nThere was an error running the postprocess script: %v\n", err)
-		} else {
+		if err == nil {
 			fmt.Fprintf(Err, "\nSuccessfully executed the postprocess script.\n")
+		} else {
+			fmt.Fprintf(Err, "\nThere was an error running the postprocess script: %v\n", err)
 		}
 	}
 	return nil
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	netURL "net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/download.go
```diff
@@ -125,6 +126,18 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+	// run postprocess script if one was configured
+	if usrCfg.GetString("postprocess") != "" {
+		cmd := exec.Command(usrCfg.GetString("postprocess"))
+		cmd.Dir = metadata.Dir
+		err := cmd.Run()
+		if err != nil {
+			fmt.Fprintf(Err, "\nThere was an error running the postprocess script: %v\n", err)
+		} else {
+			fmt.Fprintf(Err, "\nSuccessfully executed the postprocess script.\n")
+		}
```
kotp, line 139:
```suggestion
		if err == nil {
			fmt.Fprintf(Err, "\nSuccessfully executed the postprocess script.\n")
		} else {
			fmt.Fprintf(Err, "\nThere was an error running the postprocess script: %v\n", err)
		}
```

This focuses on the desired, and if that isn't the case, then it fails.  (Not tested).

joelnordell: I tested this change locally and it was fine.

arlm:
Quicl question, does it make sense to add a configuration flag that runs the post-process script relatively to the track directory so that people can have specialized scripts for each language/tooling instead of needing to add a lot if if-clauses and switch-clauses?

joelnordell:
> Quick question before the code review finishes, does it make sense to add a configuration flag that runs the post-process script relatively to the track directory so that people can have specialized scripts for each language/tooling instead of needing to add a lot if if-clauses and switch-clauses?

You might be able to do that with the existing implementation by specifying a relative path for the postprocess script, e.g. `../postprocess`, perhaps? Or if that doesn't work, using a small wrapper script which itself just does something like `exec ../postprocess` might do it.

Or maybe an enhancement here could be to pass some arguments to the script to make it easier to implement functionality like that in your own script (such as passing the track path etc.).

I haven't used exercism in a while (this branch is very old!) so my memory is fuzzy about how the directories are structured, and/or what may have changed in the past couple years.

arlm:
Basically, we have the track name and then the exercise name like: `BASE_PATH/track-name/exercise-name`. My suggestion was that either we have a parameter, like you said, stating which exercise and track is running, or that a script located inside the track folder runs using the exercise folder as current/working path. This way things can be prepared per language.

Lately I need to set files as `not-read-only` on some exercises when I update them, but I would like to run `npm install`, or other package manager to prepare for example, on some tracks. Also, when the exercise is updated and I need to force, I need to remove all over again the `ignore` commands/attributes on the test files, for all tracks, and a `sed` command could do that for me if I know the language ...

ErikSchierboom:
IMHO this can be done just as well with a custom (bash) script instead of baking it into the CLI. Therefore, I'm closing this PR.

PR #1009, opened 2021-10-03
Reduce the returns in Dir to 2 and make it shorter

RHL120:


kotp:
@RHL120 You closed it again, I am assuming for the same thought that it is not liked.  We just may not be ready for it to come in yet.  If you close it, then it is likely that this change may be forgotten about, regardless of its quality and if we like it.

(I know it can be irritating to have a PR open for a long time, seemingly ignored and neglected. But you took the time to write it, let us take the time to review it, where we can take the proper time it deserves.)

RHL120:
> @RHL120 You closed it again, I am assuming for the same thought that it is not liked.  We just may not be ready for it to come in yet.  If you close it, then it is likely that this change may be forgotten about, regardless of its quality and if we like it.
> 
> (I know it can be irritating to have a PR open for a long time, seemingly ignored and neglected. But you took the time to write it, let us take the time to review it, where we can take the proper time it deserves.)

Ok i am sorry if it was annoying to have someone opening and closjng Prs. I will reopen it.

ErikSchierboom:
@ekingery What do you think of this PR?

config/config.go
```diff
@@ -52,30 +52,22 @@ func SetDefaultDirName(binaryName string) {
...
-		}
-		dir = os.Getenv("XDG_CONFIG_HOME")
-		if dir == "" {
-			dir = filepath.Join(os.Getenv("HOME"), ".config")
-		}
-		if dir != "" {
-			return filepath.Join(dir, DefaultDirName)
+		if ev = os.Getenv("EXERCISM_CONFIG_HOME"); ev == "" {
+			if ev = os.Getenv("XDG_CONFIG_HOME"); ev == "" {
+				ev = os.Getenv("HOME")
```
ekingery, line 61: Do we not need to append ".config" here, as in the conditional on the left in L.68?

ekingery (commented):
The refactor looks good, and I _think_ this code is tested reasonably well (`gocov` is probably worth running to confirm?). However, I'm not sure about dropping the ".config" suffix for `XDG_CONFIG_HOME`? Probably worth holding until someone can confirm that is what we want to do (or maybe I'm misreading the code / intention).

config/config.go
RHL120: Hi sorry for the 1 month long delay. Yes I did append .config in the latest commit. I will mark as resolved if the result is what you expected.

config/config.go
```diff
@@ -52,30 +52,22 @@ func SetDefaultDirName(binaryName string) {
...
+		if ev = os.Getenv("EXERCISM_CONFIG_HOME"); ev == "" {
+			if ev = os.Getenv("XDG_CONFIG_HOME"); ev == "" {
+				ev = filepath.Join(os.Getenv("HOME"), ".config")
+			}
 		}
 	}
-	// If all else fails, use the current directory.
-	dir, _ = os.Getwd()
-	return dir
+	if ev == ".config" {
```
ekingery, line 65: I _think_ this would fix the change for `XDG_CONFIG_HOME`, however it now changes the behavior for windows, right? In order to approve, I think we would need to see a test case that documents the expected config file names for each OS and confirms the previous behavior matches the new behavior. Otherwise, I don't think the refactor is  worth the risk of breaking the configs for some subset of users. \cc @ErikSchierboom

ekingery (changes requested):
I'm not familiar enough with the current requirements for config file names across OS's, to be able to vouch for this code not breaking (windows currently, but any OS generally). In the comments I requested a test to be added in order to confirm the expected behavior in each case.

config/config.go
ErikSchierboom: > I think we would need to see a test case that documents the expected config file names for each OS and confirms the previous behavior matches the new behavior.

I really wouldn't want a refactoring to change behavior (which technically means it _isn't_ a refactoring).

ekingery (commented):
@rhl120 I appreciate your efforts in this PR. However, I am not convinced this refactor is A) a true refactor (tests confirming pre and post refactor functionality would be required to prove this) and B) that the modest readability improvement justifies the risk of changing this critical section of (what I think is) currently untested logic.

RHL120:
> @RHL120 I appreciate your efforts in this PR. However, I am not convinced this refactor is A) a true refactor (tests confirming pre and post refactor functionality would be required to prove this) and B) that the modest readability improvement justifies the risk of changing this critical section of (what I think is) currently untested logic.

Ok great! What should I do?

ekingery:
> > @RHL120 I appreciate your efforts in this PR. However, I am not convinced this refactor is A) a true refactor (tests confirming pre and post refactor functionality would be required to prove this) and B) that the modest readability improvement justifies the risk of changing this critical section of (what I think is) currently untested logic.
> 
> Ok great! What should I do?

You could add one or more unit tests documenting and validating the existing functionality in a separate PR. Once that is merged, if the tests still pass with the refactor, we can reconsider this PR. To be candid I'm not sure if it's worth the effort, as the readability is improved in this PR, but I would call it a modest improvement and probably not worth the risk of the change. I'd defer to @ErikSchierboom if it is worth continuing to pursue this PR assuming tests were added. The tests on the other hand, those are definitely valuable and would be a welcome addition to the codebase.

ErikSchierboom:
> I'd defer to @ErikSchierboom if it is worth continuing to pursue this PR assuming tests were added. 

I'm not a Go programmer and don't have a natural affinity with how Go code should be formatted, so I don't think I can have a well-based opinion.

> The tests on the other hand, those are definitely valuable and would be a welcome addition to the codebase.

Agreed. Tests would be very helpful.

ErikSchierboom:
Having looked at this PR, I'm not sure the refactoring actually improves readability much (which is subjective, I admit). Therefore, I'm closing this PR. Thanks for the PR!

PR #957, opened 2020-09-02
Removing team functionality

ekingery:
Resolves #956

PR #1013, opened 2021-10-29
Bump the printed version of Exercism to v3

siebenschlaefer:
This is an tiny patch that updates the updates the printed version of Exercism to "v3".

Closes https://github.com/exercism/exercism/issues/6078

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,7 +16,7 @@ import (
 var RootCmd = &cobra.Command{
 	Use:   BinaryName,
 	Short: "A friendly command-line interface to Exercism.",
-	Long: `A command-line interface for the v2 redesign of Exercism.
+	Long: `A command-line interface for the v3 redesign of Exercism.
 
 Download exercises and submit your solutions.`,
 	SilenceUsage: true,


ErikSchierboom (approved):

ErikSchierboom:
@ekingery I don't know how new releases are handled?

ekingery:
> @ekingery I don't know how new releases are handled. Could you enlighten me perhaps?

The [RELEASE.md](/exercism/cli/blob/main/RELEASE.md) should be more-or-less up to date. However, there are a bunch of PRs both merged and waiting for merge (e.g. #957) that are backed up waiting on V3 and various other decisions to be made. So it's probably not nearly as simple as following those instructions. Happy to help, slack is probably the best way to discuss.

cmd/root.go
```diff
@@ -16,7 +16,7 @@ import (
 var RootCmd = &cobra.Command{
 	Use:   BinaryName,
 	Short: "A friendly command-line interface to Exercism.",
-	Long: `A command-line interface for the v2 redesign of Exercism.
+	Long: `A command-line interface for the v3 redesign of Exercism.
```
ErikSchierboom, line 19: I'm wondering if we should even be mentioning the version here, but for now it's as least better than the current text :)

PR #1044, opened 2022-06-07
Submit without specifying files

### LINKED ISSUE #991: Allow submitting solution without specifying files
One of the most requested features for the CLI is to be able to submit a solution without having to explicitly specify the files to submit. With Exercism v3, each exercise will have a `.meta/config.json` file (see https://github.com/exercism/docs/blob/main/building/tracks/concept-exercises.md#file-metaconfigjson and https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-metaconfigjson). This file contains a `files` property with a nested a `solution` string array property. The values in the `solution` array are the names of the files that the student should work on. In other words: these file names can be used to determine the files that should be submitted, provided no files were specified by the student.

As an example, consider the following `.meta/config.json` file:

```json
{
  "files": {
    "solution": ["Lasagna.fs"],
    "test": ["LasagnaTests.fs"],
    "exemplar": [".meta/Exemplar.fs"]
  }
}
```

If the student runs `exercism submit` (no file names specified), the CLI should then interpret this as if the student ran `exercism submit Lasagna.fs`.

Note that if a student _does_ specify files, the CLI should use those and ignore the `.meta/config.json` file.

Also note that the `solution` property is an array, so it can contain multiple files (which should all be submitted).

andrerfcsantos:
Fixes #991

Adds the possibility of submitting without specifying files, with the "algorithm" specified in https://github.com/exercism/cli/issues/991#issuecomment-939768772

Other than the tests in the PR, I also tried to submit solutions using this version of the program, including messing up the exercise directory and everything went as expected :)

### DIFF
--- a/cmd/submit.go
+++ b/cmd/submit.go
@@ -26,7 +26,6 @@ var submitCmd = &cobra.Command{
 
     Call the command with the list of files you want to submit.
 `,
-	Args: cobra.MinimumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cfg := config.NewConfig()
 
@@ -44,6 +43,14 @@ var submitCmd = &cobra.Command{
 		v.SetConfigType("json")
 		// Ignore error. If the file doesn't exist, that is fine.
 		_ = v.ReadInConfig()
+
+		if len(args) == 0 {
+			files, err := getExerciseSolutionFiles(".")
+			if err != nil {
+				return err
+			}
+			args = files
+		}
 
 		return runSubmit(cfg, cmd.Flags(), args)
 	},
@@ -112,6 +119,23 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 
 	ctx.printResult(metadata)
 	return nil
+}
+
+func getExerciseSolutionFiles(baseDir string) ([]string, error) {
+	v := viper.New()
+	v.AddConfigPath(filepath.Join(baseDir, ".exercism"))
+	v.SetConfigName("config")
+	v.SetConfigType("json")
+	err := v.ReadInConfig()
+	if err != nil {
+		return nil, errors.New("no files to submit")
+	}
+	solutionFiles := v.GetStringSlice("files.solution")
+	if len(solutionFiles) == 0 {
+		return nil, errors.New("no files to submit")
+	}
+
+	return solutionFiles, nil
 }
 
 type submitCmdContext struct {
--- a/cmd/submit_test.go
+++ b/cmd/submit_test.go
@@ -108,6 +108,45 @@ func TestSubmitExerciseWithoutMetadataFile(t *testing.T) {
 	}
 }
 
+func TestGetExerciseSolutionFiles(t *testing.T) {
+
+	tmpDir, err := ioutil.TempDir("", "dir-with-no-metadata")
+	defer os.RemoveAll(tmpDir)
+	assert.NoError(t, err)
+
+	_, err = getExerciseSolutionFiles(tmpDir)
+	if assert.Error(t, err) {
+		assert.Regexp(t, "no files to submit", err.Error())
+	}
+
+	validTmpDir, err := ioutil.TempDir("", "dir-with-valid-metadata")
+	defer os.RemoveAll(validTmpDir)
+	assert.NoError(t, err)
+
+	metadataDir := filepath.Join(validTmpDir, ".exercism")
+	err = os.MkdirAll(metadataDir, os.FileMode(0755))
+	assert.NoError(t, err)
+
+	err = ioutil.WriteFile(
+		filepath.Join(metadataDir, "config.json"),
+		[]byte(`
+{
+	"files": {
+		"solution": [
+		  "expenses.go"
+		]
+  	}
+}
+`), os.FileMode(0755))
+	assert.NoError(t, err)
+
+	files, err := getExerciseSolutionFiles(validTmpDir)
+	assert.NoError(t, err)
+	if assert.Equal(t, len(files), 1) {
+		assert.Equal(t, files[0], "expenses.go")
+	}
+}
+
 func TestSubmitFilesAndDir(t *testing.T) {
 	tmpDir, err := ioutil.TempDir("", "submit-no-such-file")
 	defer os.RemoveAll(tmpDir)


asarkar:
This PR seems to assume the current directory is an exercise directory. As a student, I’ve build and submit my projects from the workspace root, not from individual exercises. I should be able to do `exercism submit <exercise_dir>` from anywhere, since the CLI already knows the workspace path.

andrerfcsantos:
This PR does assume that.

I sympathize with your use case, but I think allowing directories should probably be discussed better outside this PR. The idea here is to allow to submit files without specifying anything. Allowing directories should be another feature and a separate discussion as the CLI currently explicitly disallows them.

asarkar:
I don’t think it is “my use case” as much as it is convenience and common sense. If the point of this PR is to improve user experience, adding an additional step of going in the directory is anything but. If the goal of this PR is to reduce user error due to missing files, that is not achieved either because they may submit from a wrong working directory.
Without directory support, all this PR does is help close a ticket, not help a student.

cmd/submit.go
```diff
@@ -114,6 +121,23 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 	return nil
 }
 
+func getExerciseSolutionFiles(baseDir string) ([]string, error) {
+	v := viper.New()
+	v.AddConfigPath(filepath.Join(baseDir, ".exercism"))
```
coriolinus, line 126: Pretty sure it's now `.meta`, not `.exercism`.

andrerfcsantos: It's `.meta/config.json` in the tracks repositories, but gets downloaded to `.exercism/config.json` when the student downloads the exercise.

andrerfcsantos: It makes sense to have the `.meta` folder in the test runner because like Erik says there, it is a copy of the exercise from the repository and in the repository the folder is `.meta`.

But locally on the student's computer it is `.exercism`:

https://github.com/exercism/cli/blob/fbd552b334235d4537d55af45331107c3ad09c12/workspace/exercise_metadata.go#L15-L17

coriolinus: Pretty sure this convo can be marked as resolved; I just don't have any permissions in this repo.

cmd/submit.go
```diff
@@ -114,6 +121,23 @@ func runSubmit(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+	v.SetConfigName("config")
+	v.SetConfigType("json")
+	err := v.ReadInConfig()
+	if err != nil {
+		return nil, errors.New("no files to submit")
+	}
+	solutionFiles := v.GetStringSlice("files.solution")
+	if len(solutionFiles) == 0 {
+		return nil, errors.New("no files to submit")
+	}
```
coriolinus, line 136: Haven't written any go in a while, but I think that this is all we'd need to support globs nicely (depending on https://github.com/yargevad/filepathx):

```suggestion
	patterns := v.GetStringSlice("files.solution")
	if len(patterns) == 0 {
		return nil, errors.New("no files to submit")
	}
	
	solutionFiles := make([]String, len(patterns))
	for _, pattern := range patterns {
	    matches, err := filepathx.Glob(pattern)
	    if err != nil {
	        return nil, errors.New("invalid glob pattern") // TODO: include bad pattern in err
	    }
	    
	    solutionFiles = append(solutionFiles, matches...)
	}
```

In the event that the patterns are path literals, then this keeps current behavior. Otherwise, this gets all files which match any glob pattern in the list.

I'd personally prefer to see this included in this PR instead of needing to create a new issue and PR.

andrerfcsantos: It'd would be something like that, yes, but the logic would have to be inside the `runSubmit` function and not here. This function only runs when no paths were given and I see no point on trying to resolve globs from the paths from the `files.solution` key as I don't think it's even possible to have globs there (we don't have them in the Go track and quickly checking some exercises on the Rust track, I can't see exercises using globs in their `files.solution`).

I don't mind implementing globs and the support for specifying just a directory for upload. I'm kinda just waiting on greenlight for those features, because they weren't planned or discussed properly.

coriolinus: It's a chicken and egg problem. Right now, globs there are not supported, so we're not using them. Adding support in this tool isn't sufficient to call them Supported by Exercism, but it's necessary. 

What kind of discussion did you have in mind, to qualify as planned or discussed properly?

andrerfcsantos: > What kind of discussion did you have in mind, to qualify as planned or discussed properly?

Having more than one person weight on it at least? In particular I'd like the opinion of someone more familiar with the CLI.

I mean, this is my first PR for the CLI, I don't even know if the way I did the "no files" thing is the right way to do it given the current codebase and we're already talking about adding more features on top of it.

I'm not saying that these aren't features I would like to see, I like them, just don't see why we can't do it in baby steps and have more feedback on each of them.

coriolinus: Yes, it's only necessary if the tool and exercism must support globs. I believe that to be the case, but I also believe that merging some kind of submit-without-specifying is more important than the specific globs feature.

I don't intend to be a drag on the process here. I'm just convinced that:

- in several tracks, students can split up their work into modules which live in different files with arbitrary names
- splitting work into several files is often the right thing to do, and we should encourage it in those cases
- auto-submit without globs will always fail for a student who has done that

This implies that globs are a feature we'll want to support eventually.

ErikSchierboom:
> I sympathize with your use case, but I think allowing directories should probably be discussed better outside this PR. The idea here is to allow to submit files without specifying anything. Allowing directories should be another feature and a separate discussion as the CLI currently explicitly disallows them.

Like in [my comment above](https://github.com/exercism/cli/pull/1044#discussion_r892082066), the CLI has been carefully designed and has evolved over a long period of time. Therefore, new changes should be discussed before being implemented and carefully considered by the Exercism team.

> I don’t think it is “my use case” as much as it is convenience and common sense. 
> If the point of this PR is to improve user experience, adding an additional step of going in the directory is anything but. 

This is a personal opinion and not something that has any data to back it up with. Exercism documents and guides are all written from the viewpoint of the CLI being used from within the exercise's directory. This is also how the vast majority of users use the CLI. I also don't like the "common sense" remark, which suggests that arguing otherwise implies a lack of common sense. 

Note that we're not against any changes, but we do feel that changes to such a widely-used tool as the CLI should be carefully thought out and considered by the Exercism team. You clearly feel strongly about being able to submit a directory, so what you should do is open an issue where we can discuss its merits. 

> If the goal of this PR is to reduce user error due to missing files, that is not achieved either because they may submit from a wrong working directory.

I disagree with this statement. For the vast majority of the users, this PR will be extremely helpful. It will make using the CLI more ergonomic, greatly reduces the chance of students submitting the wrong files (this will only happen if they manually submit files) whilst still allowing for the current behavior. 

> Without directory support, all this PR does is help close a ticket, not help a student.

This is a rude statement and not constructive in any way. I'll kindly refer you to our [Code of Conduct](https://exercism.org/docs/using/legal/code-of-conduct) for guidance on how we expect our community members to interact with each other.

Side note: I think the feature this PR supports are possibly the number one most request feature.

cmd/submit.go
ErikSchierboom: I can definitely see the arguments in favor of globs, so let's discuss them in a separate issue.

ErikSchierboom (commented):
Thank you so much for doing this! This is probably our number one most wanted CLI feature :)

I don't know any Go though, so maybe @ekingery could take a look?

cmd/submit.go
coriolinus: Made the separate issue: https://github.com/exercism/exercism/issues/6399 . It's in `/exercism` instead of `/cli` because I'm pretty sure that support for that feature will require changes in at least two repos.

ErikSchierboom: Great, thanks!

iHiD:
@andrerfcsantos Thanks for doing this!

kytrinyx (approved):
This looks lovely. Clean, straight forward, and easy to follow. ❤️ Thank you for taking the time to implement this!

ekingery (approved):
Excellent, thank you @andrerfcsantos!

ErikSchierboom:
Merged! Thanks so much!

Next step: building a new release. @ekingery remind me what the steps are?

ekingery:
> > remind me what the steps are?
> 
> There is [RELEASE.md](https://github.com/exercism/cli/blob/8e73ec3131e8218c0ff31d5e0d1a050cabd6864d/RELEASE.md). The previous release was the first that was signed, with @ekingery's key. So I guess Eric has to be the person who releases, if we want to keep signing with the same key.
> 
> The process of verifying the signature is:
> 
> 1. Go to https://github.com/exercism/cli/releases
> 2. Download `exercism_checksums.txt` and `exercism_checksums.txt.sig` to the same directory
> 3. (Optionally, `gpg --recv-keys EAF483E69A029F9E89AA97CD92CA296D5C8C4CD1`, to import the key in a separate step)
> 4. Run `gpg --verify exercism_checksums.txt.sig`
> 
> I don't know whether the CLI is packaged anywhere that expects releases to be signed with that key. At least [the AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=exercism-bin) doesn't check the key. If it did, it would have this line:
> 
> ```
> validpgpkeys=('EAF483E69A029F9E89AA97CD92CA296D5C8C4CD1')
> ```
> 
> For later, should we consider having a dedicated key for signing CLI releases, which perhaps could live in this repo as a secret? See [GoReleaser docs on signing](https://goreleaser.com/customization/sign/).

Yes, the generated packages should not rely on my (or any) signature at this point, so changing the signing key for the next release will be fine. I agree, a dedicated signing key (or keys) would be best - I just signed it because it was faster to get done so it's probably also fine to have whoever is building the release sign it. Ideally we would continue to automate the release process by building and signing using Github Actions: https://goreleaser.com/ci/actions/?h=sign#signing

ErikSchierboom:
@ekingery I think getting the release process automated is quite important. Would you be interested in working on that?

ekingery:
> @ekingery I think getting the release process automated is quite important. Would you be interested in working on that?

I will spend a little time poking around - shouldn't be too much work to at least make some progress, although I can't promise anything due to a pretty tight schedule I'm working with these days. I am happy to work with anyone else who volunteers as well - anyone with some knowledge or interest in learning github actions and simple scripting tools would be able to contribute meaningfully.

ErikSchierboom:
@ekingery I'm happy to help with the GitHub Actions part, as I've worked with it quite a lot. Others can join too of course. Maybe someone from @exercism/github-actions?

ErikSchierboom:
Great! We have a [RELEASE.md document](https://github.com/exercism/cli/blob/8e73ec3131e8218c0ff31d5e0d1a050cabd6864d/RELEASE.md) that details how a release can be built using `GoReleaser` automatically.

Here is the page with details on integrating GoReleaser with GitHub actions: https://goreleaser.com/ci/actions/ Their example runs the script when a new tag is published, and I think that is a nice way to handle this.

andrerfcsantos:
@ErikSchierboom Will take a closer look at it later. However, at a quick glance, I see we'll need `secrets.GPG_PRIVATE_KEY`, `secrets.PASSPHRASE` and `secrets.GITHUB_TOKEN` defined. Do we have this already?

andrerfcsantos:
Cool!

Reading the release instructions more carefully, I realize we also need the credentials for snapcraft to publish the snap. [This post](https://forum.snapcraft.io/t/how-to-log-in-to-snapcraft-with-only-1-command/11871/5) mentions that it's possible to create a credentials file for snap. Putting this file as a secret will also be useful for CI to perform `snapcraft login` before publishing the snap.

ErikSchierboom:
Ah neat. I'll check that out.

kytrinyx:
@ekingery I'd also be happy to help work on the automation bit. Meanwhile... how hard would it be to do a manual release? I'd really like to get this change shipped so people can use it. I'd be happy to do the release if there's a way to do it without your keys, so long as I can holler if I get stuck working through RELEASE.md, since it's been a few years since I've touched this repo.

ekingery:
> @ekingery I'd also be happy to help work on the automation bit. Meanwhile... how hard would it be to do a manual release? I'd really like to get this change shipped so people can use it. I'd be happy to do the release if there's a way to do it without your keys, so long as I can holler if I get stuck working through RELEASE.md, since it's been a few years since I've touched this repo.

@kytrinyx, cutting a release manually would be a great way to prep for the automation. It should be much easier than it used to be 🤞. As you have probably seen it does require the new tooling as described in the RELEASE.md file. Happy to help with any issues you run into or questions you have! I am on the exercism slack, or we can hash things out here or via email - whatever works best for you.

kytrinyx:
Awesome, thank you @ekingery. I'll take a stab at this over the weekend.

SleeplessByte:
@kytrinyx when you eventually succeed, can you give me a ping with the version number so I can update the vscode extension to use the automated submission from that version forward?

kytrinyx:
Okay, I've cut a release but was unable to publish to snapchat (linux) and homebrew (mac). People should still be able to install the old version and then use the `exercism upgrade` command to get the newer version. Presumably. Maybe.

@SleeplessByte the new version is v3.1.0

andrerfcsantos:
@kytrinyx Can confirm upgrading with `exercism upgrade` works.

```console
$ exercism version
exercism version 3.0.13
$ exercism upgrade
$ exercism version
exercism version 3.1.0
```

kytrinyx:
@andrerfcsantos That's excellent news! Thank you.

PR #1051, opened 2022-08-17
Add basic Dockerfile to build the project.

CodeMan99:
Hello!

Would love to have the cli up on docker hub. First step is a sane `Dockerfile`. That's this PR.

In the future maybe also spin up some environment specific docker images and/or compose? Anyway, let's not get ahead of ourselves :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,4 +9,4 @@ RUN go mod download && go mod verify
 COPY ./ .
 RUN go build -v -o /usr/local/bin/exercism ./exercism
 
-CMD ["/usr/local/bin/exercism"]
+ENTRYPOINT ["exercism"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.19 as build
+
+ENV CGO_ENABLED=0
+WORKDIR /usr/src/app
+
+COPY go.mod go.sum .
+RUN go mod download && go mod verify
+
+COPY ./ .
+RUN go build -v -o /usr/local/bin/exercism ./exercism
+
+CMD ["/usr/local/bin/exercism"]


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.19 as build
+
+ENV CGO_ENABLED=0
```
CodeMan99, line 3: I saw this in the [.goreleaser](https://github.com/exercism/cli/blob/8e73ec3131e8218c0ff31d5e0d1a050cabd6864d/.goreleaser.yml#L6) file. Not sure if this is really necessary.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /usr/src/app
+
+COPY go.mod go.sum .
+RUN go mod download && go mod verify
+
+COPY ./ .
+RUN go build -v -o /usr/local/bin/exercism ./exercism
+
+CMD ["/usr/local/bin/exercism"]
```
CodeMan99, line 12: I'm not sure this is correct. I was expecting to be able to `docker run -v ~/exercism:/root exercism <SUBCOMMAND> <OPTIONS>` where `exercism` is the docker image name. Instead I had to say both the image name and the binary name.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.19 as build
```
CodeMan99, line 1: `as build` is here because I assume size will matter. This enables [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) to be use for such an optimization.

ErikSchierboom:
@CodeMan99 I haven't had the time to look at this, sorry. It is on my radar, but it'll probably be next week, sorry.

ErikSchierboom:
@CodeMan99 Thanks for the PR! What are you hoping a Dockerized version of the CLI will give you? By itself, the CLI doesn't really do anything, you need to also configure it with your specific settings. How would that work for the Dockerized version?

CodeMan99:
> Thanks for the PR! What are you hoping a Dockerized version of the CLI will give you? By itself, the CLI doesn't really do anything, you need to also configure it with your specific settings. How would that work for the Dockerized version?

A Dockerized version of a CLI gives me a standardized way of installing. It also enables me to use the image as base to build other docker environments or easily add it to a docker-compose file.

As for configuration, you either bind-mount your real file system, create a volume that you always mount, or just don't delete your container.

```bash
$ docker run --rm -v ~:/home/exercism exercism config --token="<token>"        # bind mount home directory
$ docker run --rm -v exercism:/home/exercism exercism config --token="<token>" # mount a volume named exercism
$ docker run exercism config --token="<token>"                                 # create the container with the token inside the container's filesystem
```

What this enables for you is that the user no longer needs to download anything. They can run any of those commands with no more steps since docker will download the image when not found locally.

This enables you to have a single command in your copy-command block in the UI. You could just have the `download` subcommand. The error message is already good enough to direct them to the token step. New users no longer need to wonder, "Where do I go to download that?"

--

Some example of other CLI tools I use exclusively in docker: [newman](https://hub.docker.com/r/postman/newman), [datalust/seqcli](https://hub.docker.com/r/datalust/seqcli), [sqitch](https://hub.docker.com/r/sqitch/sqitch) - though to be fair, this also uses a wrapping shell script you must download.

ErikSchierboom:
We've discussed this internally and while we really appreciate the effort, we've decided not to merge this. If you'd like to build this, feel free to fork the project to build your own Dockerized version of the CLI. Thanks again!

PR #1065, opened 2022-10-12
Fix submit command usage

### LINKED ISSUE #1063: help text for submit
The help text for submit is no longer accurate.
```sh
$ command exercism troubleshoot
...
Current: 3.1.0
...
```
```sh
$ command exercism submit --help
Submit your solution to an Exercism exercise.

    Call the command with the list of files you want to submit.

Usage:
   submit FILE1 [FILE2 ...] [flags]
...
```

The help text still indicates at least one file is required.

(I have an "exercism" shell function: "command" bypasses it)

andrerfcsantos:
Fixes #1063


andrerfcsantos:
Ready to review.

Just a side note about formatting: CI initially was complaining the code was not formatted with `go fmt`. I found this weird, because my editor runs `go fmt` to format the code. Turns out the problem seems to be that `go1.15 fmt ./...` and `go1.19 fmt ./...` don't agree on the formatting. I had to run `go1.15 fmt ./...` for CI to be happy. This meant `Use: "` having just 1 space between the `:` and `"`. On the other hand, `go1.19 fmt ./...` puts more spaces, in an attempt to align all the values of the struct vertically. I like the `go1.19` formatting more, but used the `go1.15` formatting for CI to pass.

cmd/submit.go
```diff
@@ -19,12 +19,14 @@ import (
...
 var submitCmd = &cobra.Command{
-	Use:     "submit FILE1 [FILE2 ...]",
+	Use: "submit [FILE1	 ...]",
 	Aliases: []string{"s"},
 	Short:   "Submit your solution to an exercise.",
 	Long: `Submit your solution to an Exercism exercise.
 
     Call the command with the list of files you want to submit.
+    If you omit the list of files, the CLI will attempt to submit
+    automatically the appropriate solution files for the exercise.
```
kotp, line 29:
```suggestion
    the appropriate solution files for the exercise.
```

It may not matter if it is automatically, programmatically, or some other way, so I think we can remove that word (and if we do it magically instead later, we will not feel like we have to change this information.)

kotp (approved):
Non-blocking suggestion for wording.

cmd/submit.go
```diff
@@ -19,12 +19,14 @@ import (
 
 // submitCmd lets people upload a solution to the website.
 var submitCmd = &cobra.Command{
-	Use:     "submit FILE1 [FILE2 ...]",
+	Use: "submit [FILE1	 ...]",
```
ErikSchierboom, line 22: I wonder if we could somehow make this a bit more clear, where previously the text implied multiple files could be submitted. Nested square brackets is weird.

What about:

```suggestion
	Use: "submit [FILE1, ...]",
```
or 
```suggestion
	Use: "submit [FILE(S)]",
```

cmd/submit.go
ErikSchierboom: I agree with this suggestion.

cmd/submit.go
kotp:
```
ls [OPTION]... [FILE]...
```

From the man page for `ls` to indicate optional options and files, as an example of documentation...

Git uses:

```
command [<file> ...]
```

as another example for when a file or files are optionally given, and `less` uses

```
[filename] ...
```

to indicate 0 or more file names.

ErikSchierboom: Maybe we use the git version?

```suggestion
	Use: "submit [<FILE> ...]",
```

ErikSchierboom: @andrerfcsantos what do you think about that?

kotp:
Should that merge have been a rebase on main instead of a merge of main into this branch?

cmd/submit.go
```diff
@@ -19,12 +19,14 @@ import (
...
 	Aliases: []string{"s"},
 	Short:   "Submit your solution to an exercise.",
 	Long: `Submit your solution to an Exercism exercise.
 
     Call the command with the list of files you want to submit.
+    If you omit the list of files, the CLI will attempt to submit
+    the appropriate solution files for the exercise.
```
ErikSchierboom, line 29: I think "attempt" sounds odd, like it might fail. Maybe:

```suggestion
    If you omit the list of files, the CLI will submit the
    default solution files for the exercise.
```

kotp: It might fail, but for various reasons not under its control.  Full disk, lack of network connectivity, DDOS attack on Exercism, etc.  So there are points at which it may attempt to submit and fail.

Maybe if it fails the first time it automatically retries until it succeeds, and so success is guaranteed?

PR #1084, opened 2023-02-12
Explain auto-close policy in ISSUE_TEMPLATE.md

NateEag:
Per discussion in this forum thread: http://forum.exercism.org/t/update-github-issue-template-to-warn-users-new-issues-are-being-auto-closed/3676/2

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,10 @@
 <!--
-If you're having trouble, please open an issue in
-http://github.com/exercism/exercism/issues.
+Hello. Thanks for opening an issue on Exercism - we really appreciate your interest in helping!
 
-If necessary, we will open actionable issues here
-about what needs to change.
+We are currently in a phase of our journey where we have paused community contributions to allow us to take a breather and redesign our community model. You can learn more in this blog post: https://exercism.org/blog/freeing-our-maintainers. **As such, all issues and PRs in this repository are being automatically closed.**
+
+That doesn’t mean we’re not interested in your ideas, or that if you’re stuck on something we don’t want to help. The best place to discuss things is with our community on the Exercism Community Forum: https://forum.exercism.org. If necessary, we will open actionable issues here about what needs to change.
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,7 +1,10 @@
 <!--
-If you're having trouble, please open an issue in
-http://github.com/exercism/exercism/issues.
+Hello. Thanks for opening an issue on Exercism - we really appreciate your interest in helping!
+
+We are currently in a phase of our journey where we have paused community contributions to allow us to take a breather and redesign our community model. You can learn more in this blog post: https://exercism.org/blog/freeing-our-maintainers. **As such, all issues and PRs in this repository are being automatically closed.**
+
+That doesn’t mean we’re not interested in your ideas, or that if you’re stuck on something we don’t want to help. The best place to discuss things is with our community on the Exercism Community Forum: https://forum.exercism.org. If necessary, we will open actionable issues here about what needs to change.
+
+If you do open an issue, the auto-close comment will include a link to create a discussion on the forum.
```
kotp, line 8: These lines do not match the (technical) style of the lines replaced.

iHiD, line 8:
```suggestion
Hello. Please do NOT open issues in this repository. Instead use the Exercism Community Forum: https://forum.exercism.org
```

iHiD:
@NateEag Thanks for opening this :) I've left a suggestion.

iHiD:
>We might also follow the Exercism recommendation for Markdown, here, to be consistent.

@kotp I actually think we don't want to do this here as this text is never **rendered** so the rules don't apply. Instead the text is shown as is, with normal paragraph formatting, is much more readable I think.

NateEag:
@kotp Thanks for the suggestion! I went with plain paragraph formatting for the reasons @iHiD outlined.

iHiD:
Thanks :)

kotp:
No problems.  It is not that important, this file does not churn like others can.  But the reasons are not only for rendering, but for "a sentence change should not need to show diff for the entire paragraph" so the reasoning is not only for rendering purposes, they are for maintenance and diff noise as well.

PR #1090, opened 2023-03-25
Mask API token by default when printing configuration

QuLogic:
This is already done for when requests are printed, but not configuration. This is unexpected and may lead to inadvertent leaks of tokens.


QuLogic:
The blog post is about Track maintenance, not the CLI tool.

PR #1092, opened 2023-06-25
Add `test` command to run any unit test

xavdid:
Per discussion in [this forum post](https://forum.exercism.org/t/introducing-the-universal-test-runner/6228), I've added a `test` command to the exercism CLI. Much like the [universal test runner](https://github.com/xavdid/universal-test-runner) it's inspired by, its goal is to be able to run the basic unit test suite for any exercism exercise. 

It does this by:

1. reading `.exercism/metadata.json` to get the track name
2. doing a lookup in a table which maps track name => a test command (and some configuration; see below)
3. maybe reading `.exercism/config.json` to get test file name(s) and including them
4. including any args from `exercism test`
5. Using go's `exec.Command` to run and return info about the exit code

## Types of Tests

Based on my (very brief) sampling of exercism tracks, there are 2 main ways tests are run:

1. Using a language-wide common command (e.g `cargo test`, `go test`, `lien test`, etc)
2. A single test file invoked like any program of that language (e.g. `ruby lasagna_test.rb`)

So, the `TestConfiguration` struct covers both of those cases (by including an option to append the test file(s) to the command).

The other configuration option aids in arg passing. Some language tools (namely `cargo`) expect arguments to the test command to be after a double dash (`--`). Unfortunately, so does `Cobra,` the `go` arg parser. So, I was running into the following trying to run rust tests with `--include-ignored`:

1. `exercism test --include-ignored`: `--include-ignored` is not a valid flag for the test command
2. `exercism test -- --include-ignored`: `--include-ignored` is not a valid flag for `cargo`, put it behind a `--` 
3. `exercism test -- -- --include-ignored`: works as expected

This isn't great UX though, so I added an option for languages to be able to ask that args to their test runner be put behind an implied `--`. This way, the rust command works when run as option 2 above (which is in line with most other languages).

## TODO

- [x] add tests
- [x] add more languages (I don't know that we need to ship with 100% coverage, but the top 20 most popular languages would be great)
- [x] add docs / changelog

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,3 +49,5 @@ On Windows:
 
 ### Releasing a new CLI version
 Consult the [release documentation](RELEASE.md).
+
+[fork]: https://docs.github.com/en/get-started/quickstart/fork-a-repo
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,5 +49,3 @@ On Windows:
 
 ### Releasing a new CLI version
 Consult the [release documentation](RELEASE.md).
-
-[fork]: https://docs.github.com/en/get-started/quickstart/fork-a-repo
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -26,12 +26,14 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
 // test, call the command by calling Execute on the App.
 //
 // Example:
-// cmdTest := &CommandTest{
-// 	Cmd:    myCmd,
-// 	InitFn: initMyCmd,
-// 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
-// 	MockInteractiveResponse: "first-input\nsecond\n",
-// }
+//
+//	cmdTest := &CommandTest{
+//		Cmd:    myCmd,
+//		InitFn: initMyCmd,
+//		Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
+//		MockInteractiveResponse: "first-input\nsecond\n",
+//	}
+//
 // cmdTest.Setup(t)
 // defer cmdTest.Teardown(t)
 // ...
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -47,6 +47,7 @@ func runTest(args []string) error {
 		cmdParts = append(cmdParts, args...)
 	}
 
+	fmt.Printf("--> %s\n", strings.Join(cmdParts, " "))
 	exerciseTestCmd := exec.Command(cmdParts[0], cmdParts[1:]...)
 
 	// pipe output directly out, preserving any color
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -5,31 +5,66 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/exercism/cli/workspace"
 	"github.com/spf13/cobra"
 )
 
+type PlatformSpecificCommands struct {
+	Windows string
+	Darwin  string
+	Linux   string
+}
+
 type TestConfiguration struct {
-	// The static portion of the test command, which will be run for every test on this track. Examples include `cargo test` or `go test`
-	command string
-	// Some tracks test by running a specific file, such as `ruby lasagna_test.rb`. Set this to `true` to look up and include the name of the default test file.
+	// The static portion of the test Command, which will be run for every test on this track. Examples include `cargo test` or `go test`.
+	// Might be empty if there are platform-specific versions
+	Command string
+
+	// Platform-specific test commands. Mostly relevant for tests wrapped by shell invocations. Falls back to `Command` if platform isn't available.
+	PlatformSpecificCommands
+
+	// Some tracks test by running a specific file, such as `ruby lasagna_test.rb`. Set this to `true` to look up and include the name of the default test file(s).
 	AppendTestFiles bool
 	// All args after `--` aren't parsed and are passed to the test command. Some languages (especially `rust`) expect an additional `--` between _their_ args. So instead of requiring a user to call `exercism test -- -- --include-ingored` to run all `rust` tests, set this to `true` to separate the args passed to the test runner by a `--` automatically.
-	autoSeparateArgs bool
+	AutoSeparateArgs bool
+}
+
+func (c *TestConfiguration) getTestCommand() string {
+	if c.PlatformSpecificCommands == (PlatformSpecificCommands{}) {
+		return c.Command
+	}
+
+	switch runtime.GOOS {
+	case "windows":
+		if c.PlatformSpecificCommands.Windows != "" {
+			return c.PlatformSpecificCommands.Windows
+		}
+	case "darwin":
+		if c.PlatformSpecificCommands.Darwin != "" {
+			return c.PlatformSpecificCommands.Darwin
+		}
+	case "linux":
+		if c.PlatformSpecificCommands.Linux != "" {
+			return c.PlatformSpecificCommands.Linux
+		}
+	}
+
+	return c.Command
 }
 
 var testConfigurations = map[string]TestConfiguration{
 	"go": {
-		command: "go test",
+		Command: "go test",
 	},
 	"rust": {
-		command:          "cargo test",
-		autoSeparateArgs: true,
+		Command:          "cargo test",
+		AutoSeparateArgs: true,
 	},
 	"ruby": {
-		command:         "ruby",
+		Command:         "ruby",
 		AppendTestFiles: true,
 	},
 }
@@ -59,7 +94,7 @@ func runTest(args []string) error {
 		return fmt.Errorf("test handler for the `%s` track not yet implemented. Please see HELP.md for testing instructions", track)
 	}
 
-	cmdParts := strings.Split(testConf.command, " ")
+	cmdParts := strings.Split(testConf.getTestCommand(), " ")
 
 	if testConf.AppendTestFiles {
 		testFiles, err := getTestFiles()
@@ -71,7 +106,7 @@ func runTest(args []string) error {
 
 	// pass args/flags to this command down to the test handler
 	if len(args) > 0 {
-		if testConf.autoSeparateArgs {
+		if testConf.AutoSeparateArgs {
 			cmdParts = append(cmdParts, "--")
 		}
 		cmdParts = append(cmdParts, args...)
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -1,0 +1,106 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/exercism/cli/workspace"
+	"github.com/spf13/cobra"
+)
+
+type TestConfiguration struct {
+	// The static portion of the test command, which will be run for every test on this track. Examples include `cargo test` or `go test`
+	command string
+	// Some tracks test by running a specific file, such as `ruby lasagna_test.rb`. Set this to `true` to look up and include the name of the default test file.
+	AppendTestFiles bool
+	// All args after `--` aren't parsed and are passed to the test command. Some languages (especially `rust`) expect an additional `--` between _their_ args. So instead of requiring a user to call `exercism test -- -- --include-ingored` to run all `rust` tests, set this to `true` to separate the args passed to the test runner by a `--` automatically.
+	autoSeparateArgs bool
+}
+
+var testConfigurations = map[string]TestConfiguration{
+	"go": {
+		command: "go test",
+	},
+	"rust": {
+		command:          "cargo test",
+		autoSeparateArgs: true,
+	},
+	"ruby": {
+		command:         "ruby",
+		AppendTestFiles: true,
+	},
+}
+
+var testCmd = &cobra.Command{
+	Use:     "test",
+	Aliases: []string{"t"},
+	Short:   "Infer and run the test command for an exercise.",
+	Long: `Infer and run the test command for an exercise.
+
+	Run this command in an exercise's root directory.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		track, err := getTrack()
+
+		if err != nil {
+			return err
+		}
+
+		testConf, ok := testConfigurations[track]
+
+		if !ok {
+			return fmt.Errorf("test handler for track `%s` not yet implemented. Please see HELP.md for testing instructions", track)
+		}
+
+		cmdParts := strings.Split(testConf.command, " ")
+
+		if testConf.AppendTestFiles {
+			testFiles, err := workspace.NewExerciseConfig(".")
+			if err != nil {
+				return err
+			}
+			cmdParts = append(cmdParts, testFiles.Files.Test...)
+		}
+
+		if len(args) > 0 {
+			if testConf.autoSeparateArgs {
+				cmdParts = append(cmdParts, "--")
+			}
+			cmdParts = append(cmdParts, args...)
+		}
+
+		exerciseTestCmd := exec.Command(cmdParts[0], cmdParts[1:]...)
+
+		// pipe output directly out, preserving any color
+		exerciseTestCmd.Stdout = os.Stdout
+		exerciseTestCmd.Stderr = os.Stderr
+
+		err = exerciseTestCmd.Run()
+		if err != nil {
+			// unclear what other errors would pop up here, but it pays to be defensive
+			if exitErr, ok := err.(*exec.ExitError); ok {
+				exitCode := exitErr.ExitCode()
+				// if subcommand returned a non-zero exit code, exit with the same
+				os.Exit(exitCode)
+			} else {
+				log.Fatalf("Failed to get error from failed subcommand: %v", err)
+			}
+		}
+		return nil
+	},
+}
+
+func getTrack() (string, error) {
+	metadata, err := workspace.NewExerciseMetadata(".")
+	if err != nil {
+		return "", err
+	}
+
+	return metadata.Track, nil
+}
+
+func init() {
+	RootCmd.AddCommand(testCmd)
+}
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -36,15 +36,11 @@ func runTest(args []string) error {
 		return fmt.Errorf("test handler for the `%s` track not yet implemented. Please see HELP.md for testing instructions", track)
 	}
 
-	cmdParts := strings.Split(testConf.GetTestCommand(), " ")
-
-	if testConf.AppendTestFiles {
-		testFiles, err := getTestFiles()
-		if err != nil {
-			return err
-		}
-		cmdParts = append(cmdParts, testFiles...)
+	command, err := testConf.GetTestCommand()
+	if err != nil {
+		return err
 	}
+	cmdParts := strings.Split(command, " ")
 
 	// pass args/flags to this command down to the test handler
 	if len(args) > 0 {
@@ -79,14 +75,6 @@ func getTrack() (string, error) {
 	}
 
 	return metadata.Track, nil
-}
-
-func getTestFiles() ([]string, error) {
-	testFiles, err := workspace.NewExerciseConfig(".")
-	if err != nil {
-		return []string{}, err
-	}
-	return testFiles.Files.Test, nil
 }
 
 func init() {
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -14,8 +14,8 @@ import (
 var testCmd = &cobra.Command{
 	Use:     "test",
 	Aliases: []string{"t"},
-	Short:   "Infer and run the test command for an exercise.",
-	Long: `Infer and run the test command for an exercise.
+	Short:   "Run the exercise's tests.",
+	Long: `Run the exercise's tests.
 
 	Run this command in an exercise's root directory.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -48,9 +48,7 @@ func runTest(args []string) error {
 
 	// pass args/flags to this command down to the test handler
 	if len(args) > 0 {
-		if testConf.AutoSeparateArgs {
-			cmdParts = append(cmdParts, "--")
-		}
+
 		cmdParts = append(cmdParts, args...)
 	}
 
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -44,7 +44,6 @@ func runTest(args []string) error {
 
 	// pass args/flags to this command down to the test handler
 	if len(args) > 0 {
-
 		cmdParts = append(cmdParts, args...)
 	}
 
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -42,55 +42,59 @@ var testCmd = &cobra.Command{
 
 	Run this command in an exercise's root directory.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		track, err := getTrack()
+		return runTest(args)
+	},
+}
 
+func runTest(args []string) error {
+	track, err := getTrack()
+
+	if err != nil {
+		return err
+	}
+
+	testConf, ok := testConfigurations[track]
+
+	if !ok {
+		return fmt.Errorf("test handler for the `%s` track not yet implemented. Please see HELP.md for testing instructions", track)
+	}
+
+	cmdParts := strings.Split(testConf.command, " ")
+
+	if testConf.AppendTestFiles {
+		testFiles, err := getTestFiles()
 		if err != nil {
 			return err
 		}
+		cmdParts = append(cmdParts, testFiles...)
+	}
 
-		testConf, ok := testConfigurations[track]
-
-		if !ok {
-			return fmt.Errorf("test handler for the `%s` track not yet implemented. Please see HELP.md for testing instructions", track)
+	// pass args/flags to this command down to the test handler
+	if len(args) > 0 {
+		if testConf.autoSeparateArgs {
+			cmdParts = append(cmdParts, "--")
 		}
+		cmdParts = append(cmdParts, args...)
+	}
 
-		cmdParts := strings.Split(testConf.command, " ")
+	exerciseTestCmd := exec.Command(cmdParts[0], cmdParts[1:]...)
 
-		if testConf.AppendTestFiles {
-			testFiles, err := getTestFiles()
-			if err != nil {
-				return err
-			}
-			cmdParts = append(cmdParts, testFiles...)
+	// pipe output directly out, preserving any color
+	exerciseTestCmd.Stdout = os.Stdout
+	exerciseTestCmd.Stderr = os.Stderr
+
+	err = exerciseTestCmd.Run()
+	if err != nil {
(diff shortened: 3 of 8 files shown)


cmd/test.go
```diff
@@ -0,0 +1,115 @@
...
+type TestConfiguration struct {
+	// The static portion of the test command, which will be run for every test on this track. Examples include `cargo test` or `go test`
+	command string
+	// Some tracks test by running a specific file, such as `ruby lasagna_test.rb`. Set this to `true` to look up and include the name of the default test file.
+	AppendTestFiles bool
+	// All args after `--` aren't parsed and are passed to the test command. Some languages (especially `rust`) expect an additional `--` between _their_ args. So instead of requiring a user to call `exercism test -- -- --include-ingored` to run all `rust` tests, set this to `true` to separate the args passed to the test runner by a `--` automatically.
+	autoSeparateArgs bool
+}
+
+var testConfigurations = map[string]TestConfiguration{
```
MatthijsBlom, line 23: It might make more sense to read this data from a user-configurable place: a user might want to modify the standard command, or add a missing one.

xavdid: Sure, I can look into that.

What configuration do you expect users to do? Replace the command entirely, always add specific arguments, etc. Off the top of my head, I'm thinking we add a field to `config.json` that covers cases we want to start with (`testCommand`, `args`, etc).

I think it would make sense for it to be user-overridable, but to include these defaults in the CLI. That way the command will work out of the box for most people without having to modify the config file, but the option is there if people want it.

glennj: Here's a thought: use ~/.config/exercism/user.json
Extend `exercism configure`
```sh
exercism configure --test --track go --script "go test"

exercism configure --test --track ruby --script "(cat <<'END'
  sh -c '
    for t in *_test.rb; do
      gawk -i inplace '1; /< Minitest::Test/ {print "  def skip; end"}' "$t"
      ruby "$t"
    done
  '
END
)"
```
And then ~/.config/exercism/user.json contains
```json
{
  "apibaseurl": "https://api.exercism.io/v1",
  "token": "my-uuid",
  "workspace": "/Users/glennj/src/exercism/exercism.io"
  "testing": {
    "go": "go test",
    "ruby": "  sh -c '\n    for t in *_test.rb; do\n      gawk -i inplace '1; /< Minitest::Test/ {print \"  def skip; end\"}' \"$t\"\n      ruby \"$t\"\n    done\n  '"
  }
}
```

xavdid: I'm not opposed to making it so configurable, but I wonder if we shouldn't release a simpler version initially and then add configuration options later?

With the universal test runner, one of my main design goals was "it just works". I think being able to run every test suite out of the box without configuration is a great starting point and will cover a lot of use cases and we can iterate from there.

workspace/test_configurations.go
```diff
@@ -0,0 +1,74 @@
...
+type TestConfiguration struct {
+	// The static portion of the test Command, which will be run for every test on this track. Examples include `cargo test` or `go test`.
+	// Might be empty if there are platform-specific versions
+	Command string
+
+	// Windows-specific test command. Mostly relevant for tests wrapped by shell invocations. Falls back to `Command` if we're not running windows or this is empty.
+	WindowsCommand string
+
+	// Some tracks test by running a specific file, such as `ruby lasagna_test.rb`. Set this to `true` to look up and include the name of the default test file(s).
+	AppendTestFiles bool
```
ErikSchierboom, line 14: I wonder if instead of this we could have a placeholder in the command itself, Something like:

```
jasmine-node --coffee %{test_files}
```

That way you're a little bit more flexible, and you could also support:

```
jasmine-node %{test_files} --coffee 
```

workspace/test_configurations.go
```diff
@@ -0,0 +1,74 @@
...
+	// All args after `--` aren't parsed and are passed to the test command. Some languages (especially `rust`) expect an additional `--` between _their_ args. So instead of requiring a user to call `exercism test -- -- --include-ingored` to run all `rust` tests, set this to `true` to separate the args passed to the test runner by a `--` automatically.
+	AutoSeparateArgs bool
```
ErikSchierboom, line 16: Couldn't this just be baked into the test command? `cargo test --` seems to work fine.

```
"rust": {
  Command:          "cargo test --"
}
```

CONTRIBUTING.md
```diff
@@ -49,3 +49,5 @@ On Windows:
 
 ### Releasing a new CLI version
 Consult the [release documentation](RELEASE.md).
+
+[fork]: https://docs.github.com/en/get-started/quickstart/fork-a-repo
```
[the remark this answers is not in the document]
xavdid: It is, I was just fixing a broken markdown link I found while reading the `CONTRIBUTORS.md`:

https://github.com/exercism/cli/blob/edee2079639ae4e5f2e628eb689bf489ed10b389/CONTRIBUTING.md#L17

Happy to take it out if you want to keep this PR more focused, I just figured I'd fix it while I was in there.

cmd/test.go
```diff
@@ -0,0 +1,96 @@
...
+	track, err := getTrack()
+
+	if err != nil {
+		return err
+	}
+
+	testConf, ok := workspace.TestConfigurations[track]
+
+	if !ok {
+		return fmt.Errorf("test handler for the `%s` track not yet implemented. Please see HELP.md for testing instructions", track)
```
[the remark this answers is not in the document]
xavdid: I also noticed that every track's testing instructions are at `https://exercism.org/docs/tracks/{TRACK}/tests`, so I could link out to that too (or instead).

workspace/exercise_config.go
```diff
@@ -0,0 +1,33 @@
...
+// ExerciseConfig contains metadata about the problem.
+// It's got a bunch more fields, but we don't need to read them
+type ExerciseConfig struct {
+	Files struct {
+		Test []string `json:"test"`
+	} `json:"files"`
+}
+
+// NewExerciseMetadata reads exercise metadata from a file in the given directory.
+func NewExerciseConfig(dir string) (*ExerciseConfig, error) {
```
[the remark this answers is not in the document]
xavdid: yeah! that would make sense.

workspace/test_configurations.go
xavdid: yeah, that's a great point. I think `rust` was the only track that cared about this, so that's probably simpler than having a struct field for it

workspace/test_configurations.go
xavdid: I would be fine with a placeholder, but I don't think any of the tracks included test file(s) anywhere besides the last position. I mostly went for simplicity here, but am happy to explore the template as well.

CONTRIBUTING.md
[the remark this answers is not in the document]
ErikSchierboom: > Happy to take it out if you want to keep this PR more focused, I just figured I'd fix it while I was in there.

I'd prefer that

cmd/test.go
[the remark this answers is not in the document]
ErikSchierboom: You could, but I think `HELP.md` is even better in this case

workspace/test_configurations.go
ErikSchierboom: Yeah I also don't know of any that don't do that, but I sort of like the extra flexibility. I also feel that the template version is more explicit, where one better sees the command being used.

xavdid: Yeah fair! That should be a straightforward change.

cmd/cmd_test.go
```diff
@@ -26,12 +26,14 @@ const cfgHomeKey = "EXERCISM_CONFIG_HOME"
...
 //
 // Example:
-// cmdTest := &CommandTest{
-// 	Cmd:    myCmd,
-// 	InitFn: initMyCmd,
-// 	Args:   []string{"fakeapp", "mycommand", "arg1", "--flag", "value"},
-// 	MockInteractiveResponse: "first-input\nsecond\n",
-// }
+//
+//	cmdTest := &CommandTest{
```
xavdid, line 30: CI was complaining about this block, so I ran `go fmt ./...` and this was the change. I realize you want to keep the PR focused, but I also assume you want to keep the CI green

workspace/test_configurations.go
```diff
@@ -0,0 +1,89 @@
...
+	},
+	"cfml": {
+		Command: "box task run TestRunner",
+	},
+	"cobol": {
+		Command:        "test.sh",
+		WindowsCommand: "test.ps1",
+	},
+	"coffeescript": {
+		Command: "jasmine-node --coffee {{test_files}}",
```
xavdid, line 55: I experimented with putting `{{test_files}}` in a const (which would improve the replacement code), but it made writing the commands less clear (one of `"ruby " + TEST_FiLES` or `fmt.Sprintf("ruby %s", TEST_FILES)` or `injectTestFilesPlaceholder("ruby %s")`, all of which seemed to defeat the point). 

As written, there's a higher chance that someone adding a new command will typo the placeholder and the replacement won't happen, but it provides the most clarity when reading the actual commands.

xavdid (commented):
Ok, I think that covers the feedback we had, and testing seems promising. 

The thing I'm not confident about is the best way to test it. In my [python version](https://github.com/xavdid/universal-test-runner), I focused mostly on two cases:

- given a test configuration, does it match a specific directory setup
- given a specific set of files, what's the resulting command when the top-level command is run

This is a little different, since we don't have a matching system. As long as reading the config works across tracks, this will just work. I think we'd want to:

- verify that test file replacement is happening
- verify that windows commands are being prioritized if available
- verify we can read configs (and error handling happens correctly)

I haven't done much unit testing in go, so I'd love to defer to someone else on these if it's something you (or another contributor) has a lot of experience with. I was playing with mocking the syscalls and directory lookups, but wasn't having much luck; there may be a better way to go about it.

workspace/test_configurations.go
ErikSchierboom: I think this is nice!

ErikSchierboom:
I'd like someone with Go experience to look at this (@ekingery or @junedev perhaps?).

I'm doing some functional testing:

- Simple command (tested with C#): works
- Command with {{test_files}} placeholder (tested with CoffeeScript and Ruby): work
- Windows/Linux commands: I don't have Go on my Windows machine, but the Linux bit works. Someone else should test this on Windows

workspace/test_configurations.go
```diff
@@ -0,0 +1,89 @@
...
+	},
+	"c": {
+		Command: "make",
+	},
+	"cfml": {
+		Command: "box task run TestRunner",
+	},
+	"cobol": {
+		Command:        "test.sh",
+		WindowsCommand: "test.ps1",
```
ErikSchierboom, line 52:
```suggestion
		Command:        "./test.sh",
		WindowsCommand: "pwsh test.ps1",
```

xavdid: I've updated, but was just copying from the docs. they may need to be updated: https://exercism.org/docs/tracks/cobol/tests#h-windows

junedev:
@ErikSchierboom The Go code itself looks fine to me. Can't comment on general code organization in the codebase or anything like that though.

Sidenote: On a conceptual level, I am not sure whether I am a big fan of this feature as we take away learning about how to actually run tests in the language. Even if it means just learning that in this language usually a make file is used or "there is a built-in test runner" like in the case of Go. I am ok with this being merged but I wanted to mention it.

junedev (commented):
@ErikSchierboom The Go code itself looks fine to me. Can't comment on general code organization in the codebase or anything like that though.

Sidenote: On a conceptual level, I am not sure whether I am a big fan of this feature as we take away learning about how to actually run tests in the language. Even if it means just learning that in this language usually a make file is used or "there is a built-in test runner" like in the case of Go. I am ok with this being merged but I wanted to mention it.

xavdid:
Ok, I've added little unit tests for the exercise config reading. It doesn't look like it's easy to test the windows behavior in a unit test (based on [this SO answer](https://stackoverflow.com/questions/59230895/how-should-i-handle-testing-of-a-go-function-that-uses-runtime-goos)), but I have a windows computer I can probably install `go` on and try it out.

> I am not sure whether I am a big fan of this feature as we take away learning about how to actually run tests in the language.

That's a great point! I got similar feedback on the standalone project. I've added a logline that shows exactly what's being run, so at least it's more clear what's happening:

```
% ~/projects/exercism-cli/testercism test
--> ruby lasagna_test.rb
Run options: --seed 33076

# Running:
....
```

If we're comfortable (enough to move forward) with the level of unit tests written, I think the last big tasks is go through the rest of the "easy" and "medium" language tracks and add them into the test configurations. I've been saving that for last so the PR doesn't get too unwieldy, but can grab it tomorrow.

xavdid:
Ok, I've added more tests than I thought I'd be able to. I saw that CI actually has a windows runner, so I added a test that pulls the windows command. I also got some file-based tests working by actually writing and removing files. It's in the test directory, which I don't love, but it should be fine.

The last thing is the actual `test` command itself, but that would require more refactoring to mock out the `exec.Command` call. The rest of it has enough tests that I'm personally comfortable shipping this PR as stands (+ the rest of the easy commands, as mentioned; I'll add those after the rest is ✅)

Thanks for bearing with me y'all!

ErikSchierboom:
@ekingery Are you up for a code review? I'm especially interested in the test coverage.

ekingery (approved):
Thanks for the contribution, @xavdid! I'm rusty now in both Go and this CLI codebase, however everything looks in line with what I'd expect and the test coverage is solid. Might be nice to add coveralls to this project for an empirical assessment on test coverage. All that said, 👍 from me.

ErikSchierboom:
Thanks for the reviews everybody!

ErikSchierboom:
Really happy about the state of this PR!

There are two things left to do:

1. Fix Windows CI
2. Add the test configurations for all tracks

xavdid:
Ah, pesky platform-specific error message. I _think_ I fixed CI in 540bef5. I'll get the rest of the tests in tomorrow.

xavdid:
Ok, I've added the remaining tracks (and more than I thought I'd get)!

Work is split into a few last commits:

- I added all the simple tests in https://github.com/exercism/cli/pull/1092/commits/fc6b22cc2a99db53d6a9e9d09f9f00340c21713f
- added `bash`-based tests and some sensible defaults for ambiguous tracks (`d` and `clojure`) in https://github.com/exercism/cli/pull/1092/commits/dcdf1681767fb079c053d0c9ab0aa3ef066665c7
- added `(c *ExerciseConfig) GetSolutionFiles()` and its tests (which closely mirrors `GetTestFiles()`) to power the `prolog` track (which needs both solution and test files in the command) in https://github.com/exercism/cli/pull/1092/commits/dcdf1681767fb079c053d0c9ab0aa3ef066665c7
- updated the changelog in https://github.com/exercism/cli/pull/1092/commits/8d18c398e3d2f51cb86d90945fd7febc03a8ad2b

@ErikSchierboom unless anything from those jumps out at you, I think we're all set!

The only remaining tracks are `scheme` (two tools mentioned in docs, unclear which a user will likely have installed) and `cpp` + `fortran` (complex, multi-stage setup process). My full notes are [in Airtable](https://airtable.com/appN6G2B1ylN5OQJa/shrIkepZbMogNSFpW) if you need to refer back to them.

ErikSchierboom:
> @ErikSchierboom unless anything from those jumps out at you, I think we're all set!

I'm quite busy today, so it will be next week (wednesday at the earliest). Sorry!

xavdid:
Ok, I debugged the tests on my windows machine and got them working in https://github.com/exercism/cli/pull/1092/commits/238b5729850c09a2ebc2ad0a7e1bef9895f6815d. Turns out windows won't delete a folder if it's got an open descriptor somewhere and I had neglected to close after write. Mac had no issue with this, so it was news to me. Also, I was using `path.join` instead of `filepath.join` in a couple of tests.

That CI _should_ be good to go. 🤞

cmd/test.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	cmdParts := strings.Split(command, " ")
+
+	// pass args/flags to this command down to the test handler
+	if len(args) > 0 {
+		cmdParts = append(cmdParts, args...)
+	}
+
+	fmt.Printf("--> %s\n", strings.Join(cmdParts, " "))
```
ErikSchierboom, line 49: I really like the logging! My only concern is that it might be _too_ subtle for the student to notice. Maybe we could make it more explicit? To instead of `--> <command>`, something like `Running tests via: <command>` or `Running test command: <command>`

ErikSchierboom (commented):
Works like a charm! Just a couple of minor comments (and CI is failing)

cmd/test.go
xavdid: Sure! updated to:

```
Running tests via `<command>`\n\n
```

which makes it clear exactly what's running and gives it some space to breathe

cmd/test.go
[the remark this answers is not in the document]
xavdid: yeah, great call. I went with `the \"%s\" track does not yet support running tests using the Exercism CLI. Please see HELP.md for testing instructions`, to be clear tests can be run via _a_ CLI, just not the exercism one. Also I kept the track name in, since that reminds them where they are (to reduce confusion vs "this track")

xavdid:
Ok, addressed that feedback. CI should be good to go as well!

workspace/test_configurations.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return "", err
+		}
+		cmd = strings.ReplaceAll(cmd, "{{test_files}}", strings.Join(testFiles, " "))
+	}
+
+	return cmd, nil
+}
+
+var TestConfigurations = map[string]TestConfiguration{
```
ErikSchierboom, line 62: I've just done a quick double check to see which active tracks are missing.
This is the list of missing tracks:

- abap
- common-lisp
- cpp
- delphi
- fortran
- objective-c
- pharo-smalltalk
- plsql
- powershell
- r
- scheme
- unison
- vimscript

Could you double check these tracks? If they're not supported, it might be useful to add a comment somewhere noting that those tracks are not supported (and why)

xavdid: Sure, added in ad33864eff918818939e183f8d7151582638c43a! I've confirmed that most of the remaining tracks are N/A (tests are run in IDE, etc).

glennj: For common-lisp, I used 
```sh
sbcl --load "${slug}-test" \
     --eval "(exit :code (if (${slug}-test:run-tests) 0 5))"
```
That assumes a particular distribution: there are several suggested on the track's intstallation page.

For fortran, I have this fish snippet
```fish
    if not test -d ./.exercism
        echo run me from the exercise root >&2
        return 1
    end
    if not test -d ./build
        mkdir -p build; and cd build
        test -d ./testlib; or cmake -G "Unix Makefiles" ..
        cd ..
    end
    cd build
    make; and ctest -V
    cd ..
```

workspace/test_configurations.go
```diff
@@ -209,9 +220,11 @@ var TestConfigurations = map[string]TestConfiguration{
 	"typescript": {
 		Command: "yarn test",
 	},
+	// unison: tests are run from an active UCM session
 	"vbnet": {
 		Command: "dotnet test",
 	},
+	// vimscript: tests are run from inside a vim session
```
ErikSchierboom, line 230: Looking at how [the test runner does it](https://github.com/exercism/vimscript-test-runner/blob/main/bin/run.sh#L39), it looks like this might be possible. Could you double-check?

workspace/test_configurations.go
ErikSchierboom: > R is just
> Rscript test_*.R

Could you update the configuration to use this command? I can confirm that this is also [how the test runner does it](https://github.com/exercism/r-test-runner/blob/main/bin/run.sh#L39).

workspace/test_configurations.go
```diff
@@ -90,6 +92,8 @@ var TestConfigurations = map[string]TestConfiguration{
 	"coffeescript": {
 		Command: "jasmine-node --coffee {{test_files}}",
 	},
+	// common-lisp: tests are loaded into a "running Lisp implementation", not the CLI directly
+	// cpp: tests are mostly run via IDE; only linux seems to support `make`. There's also a more complex test creation process (using `CMake`)
```
ErikSchierboom, line 96: Could we do cmake and then make via `&&`? See https://github.com/exercism/cpp-test-runner/blob/main/bin/run.sh#L43 We'd have to test if this works in Windows of course.

workspace/test_configurations.go
```diff
@@ -115,6 +120,7 @@ var TestConfigurations = map[string]TestConfiguration{
 	"erlang": {
 		Command: "rebar3 eunit",
 	},
+	// fortran: tests are mostly run via IDE; macOS/linux seem to support `make`, but there's also a more complex test creation process (using `CMake`)
```
ErikSchierboom, line 123: Similar comment to C++

testercism
sanderploegsma, line 1: I was just appreciating this new feature and noticed this file was added. I’m assuming it wasn’t meant to be committed so I thought I’d point it out 🤓

xavdid: good call, removed 😅

workspace/test_configurations.go
xavdid: Ah, good call! I didn't even think about the test runners - I've been going off the public docs

workspace/test_configurations.go
xavdid: It won't work as written. It works if I run `cmake . && make` manually, but because we're passing the command into a single `exec.Command`, it's being run as `cmake . make`, and `make` is ignored as an unknown path:

```
CMake Warning:
  Ignoring extra path from command line:

   "make"
```

To make that work, we'd need to make every command a slice of strings (most of which will only have 1 item) and then run each in order (which will work cross-platform). 

I think that's a reasonable improvement to make, but I'd recommend putting it in its own PR; this one has gotten big. I also have no idea if/how this build script works on windows- I left off the `-G "Unix Makefiles"` part of the command and I have no idea how important that is. Verifying this is probably best left to someone who knows more about those tracks.

workspace/test_configurations.go
xavdid: While this does seem to work, the output is super long and ugly. There doesn't seem to be a `--quiet` flag and I'm disinclined to ship it as is:

```
% vim -XN -i NONE -c ":so hello_world.vim" -c "Vader! hello_world.vader"
Vader note: cannot print to stderr reliably/directly.  Please consider using Vim's -es/-Es option (mode=n).
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 15 2023 04:26:46)
macOS version - arm64
Included patches: 1-1424
Compiled by root@apple.com
Normal version without GUI.  Features included (+) or not (-):
+acl               -dnd               +listcmds          +postscript        +termresponse
-arabic            -ebcdic            +localmap          +printer           +textobjects
+autocmd           -emacs_tags        -lua               -profile           +textprop
+autochdir         +eval              +menu              -python            +timers
-autoservername    +ex_extra          +mksession         -python3           +title
-balloon_eval      +extra_search      +modify_fname      +quickfix          -toolbar
-balloon_eval_term -farsi             +mouse             +reltime           +user_commands
-browse            +file_in_path      -mouseshape        -rightleft         -vartabs
++builtin_terms    +find_in_path      +mouse_dec         -ruby              +vertsplit
+byte_offset       +float             -mouse_gpm         +scrollbind        +vim9script
+channel           +folding           -mouse_jsbterm     +signs             +viminfo
+cindent           -footer            +mouse_netterm     +smartindent       +virtualedit
-clientserver      +fork()            +mouse_sgr         -sodium            +visual
+clipboard         -gettext           -mouse_sysmouse    -sound             +visualextra
+cmdline_compl     -hangul_input      +mouse_urxvt       +spell             +vreplace
+cmdline_hist      +iconv             +mouse_xterm       +startuptime       +wildignore
+cmdline_info      +insert_expand     +multi_byte        +statusline        +wildmenu
+comments          +ipv6              +multi_lang        -sun_workshop      +windows
+conceal           +job               -mzscheme          +syntax            +writebackup
+cryptv            +jumplist          +netbeans_intg     +tag_binary        -X11
+cscope            -keymap            +num64             -tag_old_static    -xfontset
+cursorbind        +lambda            +packages          -tag_any_white     -xim
+cursorshape       -langmap           +path_extra        -tcl               -xpm
+dialog_con        +libcall           -perl              +termguicolors     -xsmp
+diff              +linebreak         +persistent_undo   +terminal          -xterm_clipboard
+digraphs          +lispindent        +popupwin          +terminfo          -xterm_save
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
      user exrc file: "$HOME/.exrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
  fall-back for $VIM: "/usr/share/vim"
Compilation:
gcc -c -I. -Iproto -DHAVE_CONFIG_H   -DMACOS_X_UNIX  -g -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1
Linking: gcc   -L/usr/local/lib -o vim        -lm -lncurses  -liconv -framework Cocoa

Starting Vader: 1 suite(s), 1 case(s)
  Starting Vader: /Users/david/projects/exercism/vimscript/hello-world/hello_world.vader
    (1/1) [EXECUTE] Say Hi!
  Success/Total: 1/1
Success/Total: 1/1 (assertions: 1/1)
Elapsed time: 0.01 sec.
```

workspace/test_configurations.go
xavdid: Actually, it looks like the `cmake` portion is 1-time setup. Once it's run successfully, the only thing you need to run tests is `make`. I think it's reasonable to treat the `cmake` portion as user-required setup/installation and the actual test run as `make`. I'll update accordingly.

ErikSchierboom (commented):
It all looks really good! There is one final thing to do, and that is to remove the committed `testercism` binary from the history. It was added in a9c1fb1b65c9b487d94b129b1c5d95c06cc2ca41 and removed in 2991cfe7c2604802c4ffee6d676d7fefb6a139a5, but that still means it is in the Git repo. 

What you'll need to do is to an interactive rebase and edit the commit in which it was added to make sure that commit doesn't add the file and remove the commit in which the file was removed (which is unneeded).

Are you familiar with interactive rebasing? I tried doing it for you, but I can't push to your branch (it's a setting that you need to enable somewhere).

ErikSchierboom:
```shell
erik@schierheim:~/exercism/cli$ git log --all --full-history -- testercism
erik@schierheim:~/exercism/cli$ 
```

Nice! That looks good. Once CI is green, I'll merge. Thank you for this great contribution!

xavdid:
All right! Looks like we're all set. 

You're very welcome, glad we got it over the line. Thank you for your patience with the whole process - this one was a journey 😅

ErikSchierboom:
@xavdid Sorry that it took a bit long, but this is a big feature and one that many students will benefit from! 🎉

ErikSchierboom:
@kytrinyx Could we maybe do the release together, to make sure I have all the right permissions?

kytrinyx:
@ErikSchierboom For sure! Let's see what we can do this afternoon.

Issue #1103, opened 2023-08-04
Automate releases


ErikSchierboom:
We should automatically publish a release when a new tag is pushed to the repo.
The official goreleaser github action can be very useful: https://goreleaser.com/ci/actions/?h=github+ac action

nywilken:
Hi @ErikSchierboom is the cli project still open to contributions? If you're looking for help with this particular issue I would like to open a PR to configure goreleaser to run when a new version tag (vX.Y.Z) is pushed to the repo. 

I took a look at the release document to get a sense of how goreleaser is used now and to understand what you would like to add in the future. I have a few questions but there is enough to put something together. Let me know.

ErikSchierboom:
Yes such a PR would be much appreciated.

nywilken:
Hi @ErikSchierboom I opened a PR for adding a  release workflow for supporting automated releases.

I tested the action on my local fork to validate the configuration and give an example of the release working.  I included details in the PR description about the changes made, and left a few questions for you and other maintainers.

ErikSchierboom:
@nywilken Where is the PR? I don't see it in the list: https://github.com/exercism/cli/pulls

nywilken:
> @nywilken Where is the PR? I don't see it in the list: https://github.com/exercism/cli/pulls

@ErikSchierboom apologies for not linking it in my last comment https://github.com/exercism/cli/pull/1106 

It was auto closed by the paused community action.

ErikSchierboom:
Thanks. It'll take me a little while before I can properly review.

PR #1106, opened 2023-08-15
.github/workflows: Add workflow for supported automated releases

nywilken:
This change builds on the documented release process to add a workflow for automated releases using GoReleaser.
Upon merging this change, you will be able to trigger a new CLI release by pushing a VERSION tag to the repository.
The supported tag formats are `vX.Y.Z` and `vX.Y.Z-*` (pre-releases). By default goreleaser will use the tag
to determine if the triggered release is an official release or a pre-release.

Closes: #1103

### Additional changes for supporting automated releases:

 - The version command was updated to support variable injection via ldflags. This was done to
 allow GoReleaser to set the version of the CLI to the tagged version; goreleaser drops the `v` prefix by default setting
 the version to X.Y.Z

 - Added GPG signing action to support GPG signing the goreleaser generated checksum file. The workflow requires the following secrets be created in the upstream repository `secrets.GPG_PRIVATE_KEY` and `secrets.GPG_PASSPHRASE`. Instructions on the expected values for these secrets can be found [here](https://github.com/crazy-max/ghaction-import-gpg#prerequisites)

 - Updated the RELEASE.md document with details for triggering an automated release on GitHub. The manual instructions were left mostly unchanged for describing the process, which is a nice to have in case GitHub actions are experiencing issues.

### Open questions

 1.  How should the release notes be grouped, filtered, and sorted to ensure descriptive release notes?
 > The changelog functionality for goreleaser is currently set to use `git log` for generating release notes to attach to the GitHub release. A test release can be found
 at [nywilken/exercism-cli](https://github.com/nywilken/exercism-cli/releases/tag/v3.2.1-nywilkenrc1). The git log delta is due to this being the first release.
 1. Is there an option for using `github-native` as the changelog generation algorithm?
 > This option will generate a changelog entry for each merged PR. The PRs can be grouped by labels (enhancement, bug, breaking changing, etc...) and controlled by GitHub's native release note generation alogrithm. An example of the generated changelog is [here](https://github.com/hashicorp/packer-plugin-googlecompute/releases) and the controlling configuration is [here](https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/.github/release.yml). Refer to [GitHub docs](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configurations) for full config options. 
 1.  Is it okay for the automated release to complete before the installer is created?
 > In the current release process there is step for uploading copies of the Windows files for use by the Exercism Windows installer. But it is not clear why this is done before uploading the remaining release assets. 
 1.  Is there a reason why there are no labels being used for issues and pull-requests?


ErikSchierboom:
> How should the release notes be grouped, filtered, and sorted to ensure descriptive release notes?

I would be perfectly happy with just a regular list, not grouping or filtering or anything.

> The changelog functionality for goreleaser is currently set to use git log for generating release notes to attach to the GitHub release. A test release can be found
at [nywilken/exercism-cli](https://github.com/nywilken/exercism-cli/releases/tag/v3.2.1-nywilkenrc1). The git log delta is due to this being the first release.

Are releases created as drafts? If so, I'm happy for something like what you have to be present, as we can then tweak the text before publishing. That said, every release note starts with:

```
To install, follow the interactive installation instructions at https://exercism.io/cli-walkthrough
---
```


> Is there an option for using github-native as the changelog generation algorithm?
This option will generate a changelog entry for each merged PR. The PRs can be grouped by labels (enhancement, bug, breaking changing, etc...) and controlled by GitHub's native release note generation alogrithm. An example of the generated changelog is [here](https://github.com/hashicorp/packer-plugin-googlecompute/releases) and the controlling configuration is [here](https://github.com/hashicorp/packer-plugin-googlecompute/blob/main/.github/release.yml). Refer to [GitHub docs](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configurations) for full config options.

We could do that, we'd just need to be diligent about it. Did you check the existing labels we use https://github.com/exercism/cli/labels ?

> Is it okay for the automated release to complete before the installer is created?

The installer isn't updated, but retrieves the latest release automatically.

> In the current release process there is step for uploading copies of the Windows files for use by the Exercism Windows installer. But it is not clear why this is done before uploading the remaining release assets.

What do you mean that is it done before uploading the remaining release assets? Aren't they uploaded in `goreleaser --clean`?

> Is there a reason why there are no labels being used for issues and pull-requests?

It requires more effort, that's basically it. I'm not opposed to them being used, but it is a slight additional maintenance burden and we don't really have a maintainer.

nywilken:
Quick update - I re-rolled the proposed changes to account four your responses. I have a new example release available for you to reference https://github.com/nywilken/exercism-cli/releases/tag/v3.2.1-nywilkenrc3

The new release has the following changes:
1. A header file in a hidden directory `.release/header.md` that is used by goreleaser to automatically add the install instructions to the top of the generated release notes. 
2. Release archives for `exercism_windows_64bit.zip` and `exercism_windows_32bit.zip` are now generated and published at release time. The release archives are also included in the checksum file.
3. The generated release archives now contain the `shell/**` files. They were being skipped in previous releases.
4.  I added back the draft directive to create draft releases. I published the release so that you can preview it.

ErikSchierboom:
Sorry for the slow reviewing process, I'm slightly flooded with jobs at the moment. I _will_ get back to it.

nywilken:
> Sorry for the slow reviewing process, I'm slightly flooded with jobs at the moment. I _will_ get back to it.

@ErikSchierboom no worries. I understand. There's no rush. Please take the time you need.  I used to help with this project and  would like to help maintain it again if possible. So hoping that the release PR can help simplify some of the work for future releases.

.goreleaser.yml
```diff
@@ -1,33 +1,61 @@
...
-    - 5
-    - 6
-  ignore:
-    - goos: openbsd
-      goarch: arm
-    - goos: freebsd
-      goarch: arm
+  - id: release-build
+    main: ./exercism/main.go
+    mod_timestamp: '{{ .CommitTimestamp  }}'
```
ErikSchierboom, line 9: What does this do?

.goreleaser.yml
```diff
@@ -1,33 +1,61 @@
...
+    flags:
+      - -trimpath # removes file system paths from compiled executable
+      - -buildvcs=false # omit vcs information from binaries
```
ErikSchierboom, line 12: These seem to be new changes. Could you remove them from this PR and do them in a follow-up PR?

.goreleaser.yml
```diff
@@ -1,33 +1,61 @@
...
+    ldflags:
+      - '-s -w -X {{ .ModulePath }}/cmd.Version={{.Version}}' # sets binary version to match tag without 'v' prefix
```
ErikSchierboom, line 14: This is to replace the version with the tag, right? I'm not opposed to this change, but I also think there is some value in having the version directly defined in the code. With the current setup, there could be a disconnect between the tag version and the version in the file, right?

.goreleaser.yml
nywilken: The mod_timestamp configuration modifies the creation time of the built binary to match the value assigned to it. In this case, the creation time will be set to the time of the last commit within the tagged release.  If left unset it would just be the compilation time. 

This is done to ensure running the same job over results in reproducible (go binaries) and checksums. https://goreleaser.com/customization/builds/#reproducible-builds

.goreleaser.yml
nywilken: > This is to replace the version with the tag, right? I'm not opposed to this change, but I also think there is some value in having the version directly defined in the code. With the current setup, there could be a disconnect between the tag version and the version in the file, right?

Yes, the version is injected at build time. I agree that the version should be properly set within `cmd/version.go` before triggering the release. Seeing as bumping the version and CHANGELOG.md is part of the release process I figured the person releasing would bump the version before triggering the release. 

That said, this does add a little disconnect between the tag version and the version file if the release workflow is not followed directly. Which can also be a crutch if the tag version is correct but the version.go file is not.

If you this adds possible confusion or room for error we can drop the change. I thought locally (manual release) would be an issue but it looks like the process is to create the tag  before running goreleaser, which will result in the correct version being injected locally as well.

```
# Create a new tag on the main branch and push it
git tag -a "${TAG_NAME}" -m "Trying out GoReleaser"
git push origin "${TAG_NAME}"

# Build and release
goreleaser --clean
```

.goreleaser.yml
nywilken: Sounds good. I'll do just that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Go
```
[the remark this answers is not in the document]
nywilken: I believe the suggestion was to add a line between the steps, correct?

I made the change locally but it looks like I might need an extra line.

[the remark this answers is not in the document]
ErikSchierboom: Yes an empty line between steps would be helpful

.goreleaser.yml
ErikSchierboom: > If you this adds possible confusion or room for error we can drop the change.

I think I would prefer that. We can always add it later. Ideally we'd be able to automate as much of the version bumping too at some point.

cmd/version.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // Version is the version of the current build.
 // It follows semantic versioning.
-const Version = "3.2.0"
+var Version = "3.2.0"
```
ErikSchierboom, line 12: I guess this change can be reverted too

nywilken: Yup!

.goreleaser.yml
nywilken: Yeah good call. The change has been removed. I rebased and squashes commits a bit but you should be able to see the last commit where I revert

.goreleaser.yml
```diff
@@ -1,33 +1,48 @@
...
+        goarch: arm
+      - goos: freebsd
+        goarch: arm
+  - id: installer-build
+    main: ./exercism/main.go
+    mod_timestamp: '{{ .CommitTimestamp  }}'
+    flags:
+      - -trimpath # removes file system paths from compiled executable
+    ldflags:
+      - '-s -w -X {{ .ModulePath }}/cmd.Version={{.Version}}' # sets binary version to match tag without 'v' prefix
```
nywilken, line 40: Oops missed this one. 

```suggestion
      - '-s -w' # strip debug symbols and DWARF debugging info
```

nywilken:
@ErikSchierboom thanks for the time and help with the PR. I re-rolled to account for the feedback. Feel free to review at your convenience. 

I’m heading out for the long weekend so my responses will be delayed. Cheers!

RELEASE.md
```diff
@@ -46,12 +46,6 @@ git push origin "${TAG_NAME}"
 # Build and release
 goreleaser --clean
```
ErikSchierboom, line 47: I think this line should no longer be a part of the instructions, correct? You could instead refer to the workflow that automatically runs.

ErikSchierboom (commented):
One last tiny comment and one question: the workflow creates a draft release, correct?

nywilken:
> One last tiny comment and one question: the workflow creates a draft release, correct?

You are correct. My original commit was set to auto-publish the release but I updated [.goreleaser.yml#L68](https://github.com/exercism/cli/pull/1106/files#diff-42e26dc67aed8aa3edb2472b4403288c1699fb6dc47419b9a475f0f224fe4689L68) to create a draft after your initial feedback.

RELEASE.md
nywilken: Good catch. I removed the steps. 

The current release documentation reads well if I simply remove this step so I don't think any additional changes are needed.

nywilken:
@ErikSchierboom this is ready for review. To summarize the change, upon pushing a semver formatted tag to the repository GitHub will trigger the release workflow to create a draft release matching the pushed version tag. 

The draft release will contain all the previously released binaries along with two Windows binaries required by the installer. The generated checksum file will also containing SHA256SUM entries for the installer binaries, as they are now created by goreleaser. 

The draft release will also contain generated release-notes that include the standard install walk-through header along with a bulleted list of git log entries for all changes that went into the release. 

Lastly, before using this action the following secrets need to be added to the  repository: secrets.GPG_PRIVATE_KEY and secrets.GPG_PASSPHRASE. Instructions on the expected values for these secrets can be found [here](https://github.com/crazy-max/ghaction-import-gpg#prerequisites)

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+        with:
+          go-version: '1.19'
+
+      - name: Import GPG Key
+        id: import_gpg
+        uses: crazy-max/ghaction-import-gpg@72b6676b71ab476b77e676928516f6982eef7a41 # v5.3.0
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
```
ErikSchierboom, line 32: We already have a secret for the passphrase, which is called:

```suggestion
          passphrase: ${{ secrets. PASSPHRASE }}
```

I could regenerate, but I don't know the existing data that was used so I'd rather just re-use it.

ErikSchierboom (commented):
One tiny comment

.github/workflows/release.yml
nywilken: Ah okay that’s easy enough to change. Thanks for calling that out. 

Is it okay for me to squash all the changes into a single commit after this change? Or would you prefer I keep them separate?

ErikSchierboom: Squashing is great

nywilken: Great! All should be good to go now. Thanks again for working through these changes with me.

ErikSchierboom:
@nywilken I've just done the first release and it worked really well (I had to handle some GPG key shenanigans, but that's not on you): https://github.com/exercism/cli/actions/runs/7912951304/job/21600762816

The only thing that is not yet working is the homebrew part, which we discuss here: https://github.com/exercism/cli/pull/1121#issuecomment-1945648960

And the draft release name was `exercism-v3.3.0 runner`. Where does the " runner" suffix come from?

nywilken:
> @nywilken I've just done the first release and it worked really well (I had to handle some GPG key shenanigans, but that's not on you): https://github.com/exercism/cli/actions/runs/7912951304/job/21600762816
> 

Sweet, thanks for that info. I’m glad you found to work well for you. 

> And the draft release name was `exercism-v3.3.0 runner`. Where does the " runner" suffix come from?

Ah I didn’t notice this in the .goreleaser file but in the release name there looks to be a env.user suffix, which might where runner is coming from. 

Trying removing that from the line below 👇 

https://github.com/exercism/cli/blob/8f1e8f296f7a08158a2f8138c810ddc95e39f279/.goreleaser.yml#L130

PR #1124, opened 2024-01-04
Update FAQ link in Upgrade output

tomasnorre:
The existing link gives a 404, so after posting to the forum: 
https://forum.exercism.org/t/https-exercism-io-faqs-gives-404/8965

I have updated the link

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -32,7 +32,7 @@ We were not able to upgrade the cli because we encountered an error:
 
 Please check the FAQ for solutions to common upgrading issues.
 
-https://exercism.org/docs/using/faqs`, err)
+https://exercism.org/faqs`, err)
 		}
 		return nil
 	},
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -32,7 +32,7 @@ We were not able to upgrade the cli because we encountered an error:
 
 Please check the FAQ for solutions to common upgrading issues.
 
-https://exercism.io/faqs`, err)
+https://exercism.org/docs/using/faqs`, err)
 		}
 		return nil
 	},


cmd/upgrade.go
```diff
@@ -32,7 +32,7 @@ We were not able to upgrade the cli because we encountered an error:
 
 Please check the FAQ for solutions to common upgrading issues.
 
-https://exercism.io/faqs`, err)
+https://exercism.org/docs/using/faqs`, err)
```
iHiD, line 35:
```suggestion
https://exercism.org/faqs`, err)
```

I'd rather `faqs` is the permalink and we do the redirect on the website. But the domain should still be updated :)

tomasnorre: Ok, but me. But as the site https://exercism.org/faqs still gave 404 I didn't thought I should like to that.

cmd/upgrade.go
```diff
@@ -32,7 +32,7 @@ We were not able to upgrade the cli because we encountered an error:
 
 Please check the FAQ for solutions to common upgrading issues.
 
-https://exercism.io/faqs`, err)
+https://exercism.org/faqs`, err)
```
ErikSchierboom, line 35: As per the forum post, this should be:

```suggestion
https://exercism.org/docs/using/faqs`, err)
```

ErikSchierboom (changes requested):

cmd/upgrade.go
tomasnorre: I have adjusted this according to comment :) 
https://github.com/exercism/cli/pull/1124#pullrequestreview-1804909560

ErikSchierboom: @tomasnorre I don't see the updated version?

tomasnorre: See closed resolved conversation between @iHiD and me.

github-exercism

That's why I haven't changed it back to your suggestion.

ErikSchierboom: Oh! I didn't see that because it was resolved. Makes a ton of sense then, sorry!

ErikSchierboom: Resolved in https://github.com/exercism/website/pull/6624

ErikSchierboom (approved):

Issue #903, opened 2019-11-19
Setup chmod for certain files in downloaded exercise.


dector:
According to [sources](https://github.com/exercism/cli/blob/d95de6b2cb5d41f5d557ffcef35ed7b16008e2d9/cmd/download.go#L118), exercism tool is just downloading files. But sometimes it's really nice to make certain files executable (see [exercism/4991](https://github.com/exercism/exercism/issues/4991)).

kytrinyx:
This has been discussed previously, and the conclusion at the time was to create files as non-executable: https://github.com/exercism/cli/issues/276

uzilan:
I can't say I understand why, as running the gradlew scripts should be the way mentees run the tests, isn't it? Why make it harder?

In any case, I think we need to include instructions which explain how to chmod the files if we decide to make them non-executable.

dector:
I agree with @uzilan here. There are some traideoffs for running executable scripts fetched from remote source, indeed. But I think people will not check executable files anyway.

sshine:
## Summary and evaluation of argument made in #276

#276 was (presumably) created because some tracks commit some files as executable because track maintainers assume file permissions are retained upon download. But they're not. To resolve this discrepancy we should either retain file permissions upon download, or at least have a principle to not do so.

Making script files executable via a `#!` shebang is one of convenience: We want students to type less to run tests. This is (presumably) the motivation for some track maintainers to commit certain files as executable, and for the present issue to be raised. If there are other reasons they should be mentioned.

The only argument against this, mentioned in #276, is one of *perceived* security:

> I don't like to encourage people to simply run an executable they've fetched from a remote.

... which they would be doing anyway, even when prefixed with an interpreter. It'd be code from a remote location executed via an interpreter, whether through `./file` or `interp file`. So while there's nothing more or less secure about it, executing a script that is downloaded is *perceived* to be more dangerous if run directly.

I think the argument against executable scripts can be reduced to an aesthetical preference.

## Windows compatibility and consistency of documentation

I present another argument against making scripts directly executable.

While `./file` is an option on Linux/MacOSX/BSD, it isn't in Windows unless you [set `PATHEXT=...`](https://stackoverflow.com/questions/6818031/use-shebang-hashbang-in-windows-command-prompt) and the executable has a file ending that is listed in that environment variable.

[...]

PR #1126, opened 2024-01-21
test: use Gradle wrapper for Java track

sanderploegsma:
This updates the `exercism test` command to use the Gradle wrapper for exercises on the Java track.

Dependencies:

- [ ] https://github.com/exercism/java/pull/2657

### DIFF
--- a/workspace/test_configurations.go
+++ b/workspace/test_configurations.go
@@ -141,7 +141,8 @@ var TestConfigurations = map[string]TestConfiguration{
 		Command: "stack test",
 	},
 	"java": {
-		Command: "gradle test",
+		Command:        "./gradlew test",
+		WindowsCommand: "gradlew.bat test",
 	},
 	"javascript": {
 		Command: "npm run test",


arlm (approved):
Looks good to me

ErikSchierboom:
Note that on Linux, this requires student to set the executable bit before it can work. @sanderploegsma would you still want this to be merged?

sanderploegsma:
That's right, thanks for reminding me! To be honest I'm not sure what the right approach is here. 

Right now the installation instructions on the Java track do not include the installation of Gradle and the working locally instructions point to using the Gradle wrapper. However, this already requires students on unix-based systems to set the executable bit before they can run tests locally, unless they happen to use a local installation of Gradle instead.

The ideal solution here is still to retain file permissions when downloading exercises using the CLI. I know there is an open discussion about this in #903, perhaps it's a good idea to pick that back up? It's already mentioned there, but in the Java/Kotlin world it's considered a best-practice to use the Gradle/Maven wrappers so I really think this approach should work out-of-the-box on Exercism as well.

sanderploegsma:
Also I'd like to note that these changes at least align the Java and Kotlin tracks again. Plus, when a student is starting out with working locally and following the instructions they will:

1. Install the Exercism CLI
2. Install Java
3. Download an exercise using the CLI
4. Invoke `exercism test`
5. Get an error about the `gradle` command not being found.

With these changes step 5 will change to the `./gradlew` command not being executable, which is IMO easier to address. Plus, Windows users shouldn't run into issues at all.

ErikSchierboom:
> I know there is an open discussion about this in https://github.com/exercism/cli/issues/903, perhaps it's a good idea to pick that back up?

Yeah let's do that. Maybe for now we'll merge this?

sanderploegsma:
Yeah let's do that 👍

uzilan:
I'd like to suggest that you add instructions in _Testing on the Java track_ that explain how to chmod the .gradlew file in case people are unsure why they get the error message when running the tests

endoarato:
One issue that I found for Windows specifically is when I use Powershell 7 or nushell, 'gradlew.bat test' does not work. In both, I need to use "./gradlew.bat test'. I think it would be good to make exercism be aware of the shell that it is currently run in, since the Windows command seems to assume 'cmd'.

PR #1128, opened 2024-02-04
test: update 8th and emacs-lisp test commands

glennj:
8th as documented: https://exercism.org/docs/tracks/8th/tests

emacs-lisp to use `{{test_files}}` placeholder, initially reported in http://forum.exercism.org/t/emacs-lisp-tests-cannot-be-run/9639


ErikSchierboom (approved):
LGTM!

PR #1130, opened 2024-02-15
Fix homebrew repo name

ErikSchierboom:


baduker:
@ErikSchierboom never mind. I just saw it's fixed already.

PR #1139, opened 2024-05-02
Add pyret support to `exercism test`

ErikSchierboom:
CC @bnandras


BNAndras:
Thanks! I was thinking about doing this yesterday. :)

PR #1157, opened 2024-08-20
Updated test_configurations.go for new batch track

GroophyLifefor:
I think I have to leave it like this at this stage, because we only have the verify process. can you verify me @ErikSchierboom ?

LoC tool supports batch language,
and we don't generally use snippets in batch language.


ErikSchierboom:
> and we don't generally use snippets in batch language.

I'm not sure what you mean by this. The snippets are the bits of code (like summaries) that are shown on the website. They're not opt-into, every iteration has one created for them.

workspace/test_configurations.go
```diff
@@ -74,6 +74,7 @@ var TestConfigurations = map[string]TestConfiguration{
 	"ballerina": {
 		Command: "bal test",
 	},
+	// batch: batch test-runner disabled and focus solely on Windows.
```
ErikSchierboom, line 77: This doesn't really have anything to do with the test runner, it is all about running things locally.
And we have a `WindowsCommand` option for Windows-specific commands: https://github.com/exercism/cli/blob/27119adcb28b27d91c7e86a3160b4df9ad38093a/workspace/test_configurations.go#L157

GroophyLifefor:
> > and we don't generally use snippets in batch language.
> 
> I'm not sure what you mean by this. The snippets are the bits of code (like summaries) that are shown on the website. They're not opt-into, every iteration has one created for them.

Ah, I thinked snipped which used in website code editor, okay.

workspace/test_configurations.go
GroophyLifefor: I'm a bit confused about how this works.

In windows terminal I start batch files with call built-in command like
```cmd
call main.bat
```

but in this case if we call test files, how can I validate path?

there a example but may not correct one.
```go
"batch": {
  WindowsCommand: "call {{test_files}}",
},
```

ErikSchierboom: The command you listed will replace `{{test_files}}` with a space-separated list of all the test files, as defined in the `.files.test` key in the exercise's `.meta/config.json` file. E.g. if the student is working on the `darts` exercise, it would expand too: `call DartsTest.bat` (see https://github.com/exercism/batch/blob/main/exercises/practice/darts/.meta/config.json#L7-L9).

GroophyLifefor: I got it, wow, it's a huge ecosystem and everything can work in parallel 😄

GroophyLifefor: Can you now reReview?

ErikSchierboom (approved):

PR #1159, opened 2024-08-20
Fix dup alias "t", keep it for "test" only

muzimuzhi:
Command alias "t" was registered by both "troubleshoot" and "test" commands. This PR drops it from "troubleshoot" aliases.


ErikSchierboom:
@muzimuzhi Having slept on it a bit, I think my previous suggestion was wrong. We should keep the existing `t` shortcut for troubleshoot as to not cause any breaking changes. Maybe we can use another letter for the test command, e.g. `e`? (if that's unused)

muzimuzhi:
> We should keep the existing `t` shortcut for troubleshoot as to not cause any breaking changes.

I believe it's already been a regression among versions 3.2.0 (2023-08-01) and the latest 3.4.2, that `exercism t` stopped resolving to `troubleshoot` command, but to `test`. The culprit commit is https://github.com/exercism/cli/commit/a8ffc316a462ea07a5e4cfbfb3df24eb0f9342eb.

I'll adjust this PR to drop `t` alias from `test` command instead.

If assigning single-character aliases for each command is an important convention, compared to `e`, uppercase `T` might be better. `e` can be reserved for future command which starts with  `e`, for example `explore`.

ErikSchierboom:
> I believe it's already been a regression starting from versions 3.2.0 (2023-08-01), that exercism t stopped resolving to troubleshoot command, but to test. The culprit commit is https://github.com/exercism/cli/commit/a8ffc316a462ea07a5e4cfbfb3df24eb0f9342eb.

Right, then I'm fine with keep it as is.

ErikSchierboom (approved):

PR #1172, opened 2024-09-17
Fix `batch` track test command

BNAndras:
See http://forum.exercism.org/t/track-specific-testing-documentation/12829/6 for context. `exercism test` currently doesn't work for Batch Script exercises in Command Prompt or PowerShell. I locally built `testercism` with the proposed `cmd /c {{test_files}}` command, and it worked in my limited testing within both shells.


glennj (approved):

ErikSchierboom:
Note: this isn't automatically deployed. That's a manual task.

PR #1174, opened 2024-09-27
Fix BinaryName initialization from args

petrem:
Fixes an issue with `completion bash` where the command name is not present in the completion output.


iHiD:
(cc @exercism/cli)

glennj (approved):
Don't know go all that well. This looks reasonable.

vaeng (approved):

iHiD:
@ErikSchierboom Are you ok to merge/release this when you're back around pls?

PR #1192, opened 2025-05-22
idris uses slug

keiravillekode:


keiravillekode:
When testing in a shell, 

```
    pack test `basename *.ipkg .ipkg`
```

achieves what we need, as we have `hello-world.ipkg`

keiravillekode:
cc @ErikSchierboom from idris track-maintainers

keiravillekode:
> @keiravillekode Thanks! The gist of it looks good, except for the part where it infers the slug from the directory name. Technically, the better way to do that would be to read the config file (I think `exercism.json`) in the directory.

Updated. It is `.exercism/metadata.json` , which `ExerciseMetadata` reads (and writes in tests).

ErikSchierboom:
@keiravillekode Would you mind also updating the documentation here: https://github.com/exercism/docs/blob/main/building/tooling/cli.md#test-command-placeholders?

keiravillekode:
If you are doing a release, perhaps https://github.com/exercism/cli/pull/1193 can be included.

PR #1196, opened 2025-06-13
Check HTTP response content type before trying to decode it as JSON

IsaacG:


SleeplessByte:
You should check for both this or ends with +json so that Jeremy can switch in the future to a +json type.

E.g.

`application/problem+json`

IsaacG:
> You should check for both this or ends with +json so that Jeremy can switch in the future to a +json type.

Done. I'm not familiar with that new format. Is the intend that you can put any arbitrary string there? `application/<anything>+json`?

SleeplessByte:
Yes, common is making a media type in the `vnd.` (vendor) tree, e.g.

```
application/vnd.exercism.submission.v1+json
```

That could be specced to always give the same *format* for the JSON, instead of merely the JSON syntax.

Similar:

```
# XLS
application/vnd.ms-excel

# XLSX (open format)
application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
``` 

It's vital for HATEOAS (a way of designing HTTP APIs as described in [Roy Fieldings thesis](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm)), used by [some of the big ones](https://docs.delftsolutions.nl/wiki/HATEOAS_API). Resource dump with surface level information: [Don't be smart and use web standards](https://blog.delftsolutions.nl/software/2024/05/24/dont-be-smart-and-use-web-standards.html). [Example of implementation of type-checked JSON requests and responses](https://github.com/XPBytes/media_types-serialization).

SleeplessByte (approved):

ErikSchierboom (approved):

PR #1206, opened 2025-07-15
Drop team-specific logic from the CLI

### LINKED ISSUE #956: Remove teams for v3
The CLI currently has functionality in it for submitting to teams. With the launch of v3 that won't need to be in the CLI any more as the submitting to teams is a separate step in the website. So that code can be removed. It's not critical it happens quickly, but the API won't support the teams option any more, so probably best to launch the new CLI with the website launch.

IsaacG:
Closes #956

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -135,7 +135,7 @@ type download struct {
 	token, apibaseurl, workspace string
 
 	// optional
-	track, team    string
+	track          string
 	forceoverwrite bool
 
 	payload *downloadPayload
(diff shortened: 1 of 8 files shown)


glennj (approved):

kotp:
The history, the commit message, does not include the "why" of this commit.  Would you consider adding the "why" from issue 956 to the commit message so that this is self contained?

iHiD (approved):
Good idea. Thanks!

PR #1210, opened 2025-08-16
Use mode 0700 for the config dir, not 0755; other users should not have access

IsaacG:


keiravillekode (approved):

BNAndras (approved):

SleeplessByte (approved):
Yah I think this is correct but please try it out first 😬

PR #1215, opened 2025-09-23
Update the token URL to point to the API settings page

IsaacG:


IsaacG:
See also #1217 ; this PR actually generates a broken URL.

PR #1227, opened 2026-03-09
Add `rescript` support

BNAndras:


glennj (approved):
Approved, but it seems a bit premature for this step.

BNAndras:
Is it? Apologies. I saw the Lean PR was submitted shortly after their track and test runner repos were spun up. I'm actively working on the ReScript test runner so it should be ready by the end of this weekend.

glennj:
There are only 3 practice exercises at the moment -- that's the source my "premature" opinion.

But reconsidering, there's no harm in adding this to the CLI.

BNAndras:
Gotcha. That's fair. We have another seventeen or so drafted but not PRed. :)

IsaacG:
Still a WIP track

BNAndras:
Yup, we’re largely waiting on the Terraform stuff Jeremy is supposed to do so the track is available for testing.

PR #1228, opened 2026-05-03
Add factor support to `exercism test`

keiravillekode:


keiravillekode:
We use the `{{slug}}` functionality implemented and tested for Idris.

https://github.com/exercism/cli/pull/1192

We could add a test, but it would be a near-duplicate of the Idris test.

glennj (approved):

keiravillekode:
I don't have merge permissions

kotp (approved):

IsaacG:
Future note, we tend to only add test commands when the track is nearing launch.

PR #1246, opened 2026-08-21
Handle receiving HTTP error codes (non-200) when downloading files.

IsaacG:
Related to [this forum thread](https://forum.exercism.org/t/cli-silently-creates-0-byte-files-when-file-downloads-hit-http-429-rate-limit/75208/).

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -80,6 +80,7 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		return err
 	}
 
+	var successes int
 	for _, sf := range download.payload.files() {
 		url, err := sf.url()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/download.go
```diff
@@ -98,8 +99,10 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
 		defer res.Body.Close()
 
 		if res.StatusCode != http.StatusOK {
-			// TODO: deal with it
-			continue
+			if successes > 0 {
```
glennj, line 102: I see `successes` being declared on line 83, but it's not initialized, and I don't see where it's ever *set*.

IsaacG: Go strongly encourages the use of the default zero-values of variables. An `int` is initialized to 0. 

The lack of an explicit initializer in intentional and idiomatic. See https://github.com/exercism/go/blob/main/concepts/zero-values/about.md

glennj: Got it. But where is it incremented? When will it be greater than zero?

IsaacG: Oh. That's _supposed_ to happen at the end of the loop :smile: Added!

BNAndras (approved):

glennj (approved):

cmd/download.go
```diff
@@ -98,8 +99,10 @@ func runDownload(cfg config.Config, flags *pflag.FlagSet, args []string) error {
...
+				fmt.Fprintf(Err, "Downloaded %d/%d files", successes, len(download.payload.files()))
```
keiravillekode: We should add a trailing `\n`

to avoid

```
Downloaded 1/9 filesError: received HTTP/...
```

IsaacG: Good catch! Added.

keiravillekode (approved):

